PR #4, opened 2019-04-22
F11 for Fullscreen

kayabe:


27136773:
Please note that this change will not be available until the release of the next Desktop Client (probably in August, depending on electron 5 stable releases)

PR #67, opened 2019-04-26
Update F11

kayabe:
electron.globalShortcut listens to the registered key even if the window is not focused, so we are supposed to unregister the key when it loses focus, and register it back when its focused


27136773:
Not tested, but looks good.

PR #356, opened 2019-08-15
Update holymanga.html with new url

samlh:
Fixes chapter loading


27136773 (approved):
Confirmed changed domain

PR #368, opened 2019-08-21
Update theme.html

KennethLR:
Update on the dark theme. Changed button colors so user knows which can and can't be clicked (as best as possible). Scroll wheels now fit with theme and are stylish. Reduced number of colors that make up theme. Text is now easier to read.

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/theme.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/theme.html
@@ -38,9 +38,9 @@
                 /*
                  * application control group
                  */
-                --app-loading-screen-color: #c0c0c0;
-                --app-loading-screen-background-color: #f8f8f8;
-                --app-loading-screen-quote-border: 8px solid lightgrey;
+                --app-loading-screen-color: #aaabad;
+                --app-loading-screen-background-color: #202225;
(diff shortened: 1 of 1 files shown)


src/web/lib/hakuneko/frontend@classic-dark/theme.html
```diff
@@ -146,19 +146,32 @@
                 /*
                  * style of page view
                  */
-                --page-background-image: url('/img/background-classic-dark.png');
-                --page-control-background-color: #1d1f22; /* When nothing is selected */
+                --page-background-image: url('/img/background.png');
```
27136773, line 149: use the existing background image for darkmode

src/web/lib/hakuneko/frontend@classic-dark/theme.html
```diff
@@ -186,7 +199,7 @@
             select {
                 width: calc(100% - 0em - 0.5em); /* adjust with padding and margin */
                 padding-right: 1.5em !important;
-                background-image: url('/img/dropdown-classic-dark.png');
+                background-image: url('/img/dropdown.png');
```
27136773, line 202: use the existing dropdown image for darkmode

PR #411, opened 2019-09-10
HotChocolateScans is dead

Efreak:
Website was shut down 2+ months ago. See https://mangadex.org/thread/15603/1/#post_841580


27136773 (approved):
Please also remove the related icon in `/src/web/img/connectors/hotchocolatescans`

PR #425, opened 2019-09-15
Basic new content notification function

Sheepux:
Will notify changes and overline mangas with updates.

Only works in dark theme (waiting for review).
Trigger the process by clicking the refresh button on the bottom left corner.

Updates are detected if the bookmark is not the first in the chapter list (possible bug: the list is not ordered).
Warning: create bulk checks (all mangas at the same time). This might be considered for future fixing (bulk/batch limiter)

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
@@ -127,13 +127,13 @@
             </tr>
         </table>
         <ul class="list" >
-                <template is="dom-if" if="[[ existMangasForValidConnector(selectedConnector, mangaList.length) ]]">                        
-                    <li class="notification">
-                        Manga list is loading or empty<br>
-                        Click &nbsp;<i class$="fas fa-sync [[ getRefreshClass(selectedConnector.isUpdating) ]] refresh" on-click="onUpdateMangaListClick" title$="Synchronize local manga list with online list from &lt;[[ selectedConnector.label ]]&gt;"></i>&nbsp;
-                        button to update list
-                    </li>
-                </template>
+            <template is="dom-if" if="[[ existMangasForValidConnector(selectedConnector, mangaList.length) ]]">                        
+                <li class="notification">
+                    Manga list is loading or empty<br>
+                    Click &nbsp;<i class$="fas fa-sync [[ getRefreshClass(selectedConnector.isUpdating) ]] refresh" on-click="onUpdateMangaListClick" title$="Synchronize local manga list with online list from &lt;[[ selectedConnector.label ]]&gt;"></i>&nbsp;
+                    button to update list
+                </li>
+            </template>
             <template is="dom-repeat" items="[[ mangaList ]]" filter="[[ filterMangas(mangaPattern) ]]" rendered-item-count="{{ mangaFilteredCount }}">
                 <li class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item, item.hasUpdate) ]] " title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</li>
             </template>
--- a/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
@@ -86,7 +86,6 @@
             .footer {
                 flex: 0;
                 display: flex;
-                flex-direction: row;
             }
         </style>
         <div class="header separator">
@@ -135,14 +134,14 @@
                 </li>
             </template>
             <template is="dom-repeat" items="[[ mangaList ]]" filter="[[ filterMangas(mangaPattern) ]]" rendered-item-count="{{ mangaFilteredCount }}">
-                <li class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item, item.hasUpdate) ]] " title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</li>
+                <li class$="manga [[ item.status ]]  [[ getMangaClass(selectedManga, item, item.hasUpdate) ]] " title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</li>
             </template>
         </ul>
         <!-- https://github.com/PolymerElements/iron-list/issues/536 -->
         <!--
         <iron-list class="list" items="[[ filteredMangas(mangaList,mangaPattern) ]]">
             <template>
-                <div class$="manga [[ item.status ]] [[ item.hasUpdate]] [[ getMangaClass(selectedManga, item.id) ]]" title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</div>
+                <div class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item, item.hasUpdate) ]]" title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</div>
             </template>
         </iron-list>
         -->
@@ -198,7 +197,6 @@
                 this.mangaList = this['mangaList'];
                 // register callbacks for published events
                 document.addEventListener( EventListener.onMangaStatusChanged, this.onMangaStatusChanged.bind( this ) );
-                document.addEventListener( 'onCheckNewContent', this.onCheckNewContent.bind( this ) );
                 Engine.BookmarkManager.addEventListener( 'changed', this.onBookmarksChanged.bind( this ) );
                 this.set( 'updatingContent', false );
             }
@@ -251,7 +249,6 @@
                 this.notifyPath( 'selectedConnector.isUpdating' );
             }
             
-
             /**
              *
              */
@@ -275,8 +272,8 @@
             }
             
             /**
-            * Event lister attached to the refresh button
-            * Checks for new content to read by refresh the current manga list, updating the web source and checking the bookmarks
+            * Event listener attached to the refresh button
+            * Checks for new content to read by refreshing the current manga list, updating the web source and checking the bookmarks
             */
             onCheckNewContent() {
                 this.set( 'updatingContent', true );
--- a/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
@@ -134,7 +134,7 @@
                         button to update list
                     </li>
                 </template>
-            <template id='mangatable' is="dom-repeat" items="[[ mangaList ]]" filter="[[ filterMangas(mangaPattern) ]]" rendered-item-count="{{ mangaFilteredCount }}">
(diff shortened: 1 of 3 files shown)


src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -189,7 +198,9 @@
                 this.mangaList = this['mangaList'];
                 // register callbacks for published events
                 document.addEventListener( EventListener.onMangaStatusChanged, this.onMangaStatusChanged.bind( this ) );
+                document.addEventListener( 'onCheckNewContent', this.onCheckNewContent.bind( this ) );
```
27136773: Remove event registration if the event is never going to be dispatched

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -261,12 +273,92 @@
...
+                })
+                .catch(e => console.log(`Error checking new chapters  ${e}`)); // Catch the error
+            }
             /**
              *
              */
-            getMangaClass( selectedManga, id ) {
-                return ( !selectedManga || selectedManga.id !== id ? '' : 'focus' );
+            getMangaClass( selectedManga, manga, hasUpdate ) {
+                if(manga.hasUpdate)  { return 'newcontent'}
```
27136773: This will lead to highlighted mangas can never be focused (background color for selected manga).
Is this behavior intended or a bug?
If a bug, return the correct classes for all 4 permutations:
e.g.

- ''
- 'focused'
- 'newContent'
- 'focused newContent'

Focus works for non-highlighted manga (Zan)
image
Focus works not for highlighted mangas (Air Gear)
image

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -130,19 +135,23 @@
...
-                <li class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item.id) ]]" title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</li>
+                <li class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item, item.hasUpdate) ]] " title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</li>
             </template>
         </ul>
         <!-- https://github.com/PolymerElements/iron-list/issues/536 -->
         <!--
         <iron-list class="list" items="[[ filteredMangas(mangaList,mangaPattern) ]]">
             <template>
-                <div class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item.id) ]]" title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</div>
+                <div class$="manga [[ item.status ]] [[ item.hasUpdate]] [[ getMangaClass(selectedManga, item.id) ]]" title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</div>
```
27136773: Signature mismatch, second parameter must be of type Manga, third must be of type boolean

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -82,6 +85,8 @@
             }
             .footer {
                 flex: 0;
+                display: flex;
```
27136773, line 88: Declared footer as flexbox, but the present elements seems to be used as inline text

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -239,6 +250,7 @@
                 // trigger update for refresh button style (update started)
                 this.notifyPath( 'selectedConnector.isUpdating' );
             }
+
```
27136773, line 251: Linting error, no consecutive empty lines

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -261,12 +273,92 @@
...
+                        await connector.initialize();
+                        return Promise.resolve(); 
+                };
+                //Function to refresh chapters and check if new content
+                const checkManga = async function checkManga(manga){
+                    context.set('nbContentUpdating',++nbContentUpdating);
+                    await manga._getUpdatedChaptersFromWebsite();
+                    //console.debug(`Check manga: ${manga.title} with connector ${manga.connector.label} initialized: ${manga.connector.initialized} - ${manga.connector.isUpdating}`);           
+                    await manga.getChapters( ( error, chapters ) => {
+                        //Is the first chapter on the top of the list marked ? If not, we have new content !
```
27136773, line 303: There might be some flaws.

- What if an missing older chapter has been added later to the website?
- What if multiple languages or scanlation groups for the same chapter exists, so the marked one is not the first?

----

image

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -261,12 +273,92 @@
             onSelectedMangaChanged( manga ) {
                 //
             }
+            
+            /**
+            * Event lister attached to the refresh button
+            * Checks for new content to read by refresh the current manga list, updating the web source and checking the bookmarks
+            */
+            onCheckNewContent() {
```
27136773, line 278: Remove the business logic from UI and put it into the HakuNeko engine.
In this case, maybe introduce a new function in Manga.mjs such as `async getAllChaptersNewerThenMarked()` which does all the workload and e.g. returns an array or the number of all chapters that are newer then the one with the marked chapter.

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -261,12 +273,92 @@
...
+                this.set( 'updatingContent', true );
+                console.debug('Checking new mangas');
+                let newContent=[];
+                let connectors=[];
+                let nbContentUpdating=0;
+                let context=this;
```
27136773, line 284: Use `this` directly in the scope instead of assigning to local variable which may reduce the expressiveness.
Also be consistent with the usage, at the end of the function this is used instead of context.

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -261,12 +273,92 @@
...
+                    if (! manga.connector.initialized && ! manga.connector.isUpdating ) {
+                        manga.connector.isUpdating=true;
+                        connectors.push(manga.connector);
+                    }
+                });
 
+                //Function to initialize connector
+                const initConnector = async function initConnector(connector){
+                        await connector.initialize();
+                        return Promise.resolve();
```
27136773: No need to return promise in async function

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -261,12 +273,92 @@
...
+                            manga.hasUpdate=true;
+                            newContent.push(manga);
+                        }       
+                    });
+                    context.set('nbContentUpdating',--nbContentUpdating);
+                    return Promise.resolve(manga); 
+                };
+                context.set('nbContentUpdating','init');
+                //Init all connectors
+                Promise.all(connectors.map(initConnector))
```
27136773, line 316: MangaDex has 30000+ mangas, opening 30000 requests to the server concurrently and performing the UI update will freeze the application.
Some websites such as KissManga will ban your IP for to many requests per second.

May perform update check sequentially instead of concurrently.
May perform update check only for bookmarks (disable button for all others).

src/web/mjs/engine/Manga.mjs
```diff
@@ -22,7 +22,7 @@ export default class Manga {
         this.status = status;
         this.chapterCache = [];
         this.existingChapters = [];
-
+        this.hasUpdate=false;
```
27136773: Maybe use more expressive name as already done in the UI, e.g. `hasNewContent`, `hasNewChapters`, ...

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -261,12 +273,92 @@
...
+                       //Send Notification
+                        console.log(`${newContent.length} New content: ${contentShort}`); 
+                        let myNotification = new Notification('Hakuneko', {
+                            body: 'New content : '+contentShort.join(',').substring(0,185)
+                        })
+                    }
+                    this.set('updatingContent', false );
+                    this.set('nbContentUpdating',0);
+                })
+                .catch(e => console.log(`Error checking new chapters  ${e}`)); // Catch the error
```
27136773: Use `console.warn` or `console.error`, should there be a message for the user as well?

27136773 (commented):
**General:**

- Try to separate UI and business logic
- Try to split large method into smaller expressive functions

**Bugs:**

- A manga with 0 chapters is highlighted red
- A manga keeps the red highlighting, even after setting the chaptermark at the latest chapter and re-check for new content

**Notes:**

Due to the refactoring of the event architecture, each Manga will extend the EventTarget class, which may become very convenient for the update check.
Instead of waiting for the returned updated chapters, the UI can listen on the event.

Something like this:
```javascript
mangaList.forEach(manga => manga.addEventListener('updated', this.updateStuff.bind(this)));

// ...

updateStuff(event) {
    let updatedManga = event.target; // or maybe event.details
    // check manga properties and update UI if necessary
}

// ...

onCheckUpdateClick() {
    // just call the update check for each manga, it will dispatch the 'updated' event when ompleted
    mangaList.forEach(manga => manga.checkChaptersNewerThenMarked());
}
```

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: for this i plan to use something like https://www.npmjs.com/package/bottleneck

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: Indeed, it was only there when the refresh button was in the jobs pannel.

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: Indeed, forgot to backport in the comments the changes on line 138

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: fixed

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: That was the only way i found out to get it working as inline. Any suggestion ?

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: This was the only way i was able to reuse "this" into the checkManga function as i struggled to pass parameters to the map functions.
I think i should stop using the map function.

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: Also, this function is supposed to be used only on mangas in the bookmarks. I guess we should disable the button if the bookmarks connector is not active.

Sheepux:
Closing PR, not merged, reworking with events. Also using a new branch.

Issue #446, opened 2019-09-22
Recently read update


hypoflex:
Update recently read according to what has last been read with the build in reader

hypoflex:
Feature has been implemented

Issue #447, opened 2019-09-25
Cross platform bookmarks / chaptermarkers


hypoflex:
would love to sync bookmarks and chapters between my laptop and desktop

one way of achieving this relative easily (since hakuneko cant provide a cloud save) being able to set the location of bookmarks and chaptermarkers file.

so this can be set to a external drive / cloud folder (dropbox/Gdrive) / NAS / SMB server

27136773:
Path of the [User Data](https://github.com/manga-download/hakuneko/wiki/User-Manual#22-user-data) directory can be changed via command line arguments (e.g. by creating a shortcut to HakuNeko's executable containing this argument)

`hakuneko.exe --user-directory='D:\...'`

hypoflex:
but that would covert the hole directory, 
meaning all files would be synced. now from windows to windows that wouldn't be a problem
but windows to OSX (which i use) the settings file would not like that.

of course this could be manipulated by only syncing the bookmarks and chapter file.
but a dedicated check button + directory location in the settings would be more user friendly.

hypoflex:
Feature was added, Bookmark file and Chaptermarker can be transfered to different folder for sync with a cloud option

PR #448, opened 2019-09-25
Auto chaptermarker & custom bookmark location

hypoflex:
#446
Once a chapter is previewed it will be marked as "last read"

Using the build in viewer/reader will keep track of where you left off. once you close the build in viewer/reader it will automatically put that last chapter as "last read".


src/web/lib/hakuneko/frontend@classic-dark/chapters.html
```diff
@@ -94,12 +94,12 @@
...
                     <td>
-                        Chapter List 
+                        Chapter List
                     </td>
-                    <td width="1"> 
+                    <td width="1">
                         <i class$="fas fa-fw button [[ getSortClass(chapterSort) ]]" on-click="onToggleSortClick" title="Click to toggle chapter sorting"></i>
                     </td>
-                    <td width="1"> 
+                    <td width="1">
```
hypoflex, line 102: Not sure why its doing this...?
tried reforking the hole project and removing it from my machine.
but it keeps editing these, bug?!

src/web/lib/hakuneko/frontend@classic-dark/chapters.html
```diff
@@ -574,16 +581,16 @@
...
             onChapterStatusChanged( e ) {
                 let chapter = e.detail;
                 if( !this.chapterList
                     || !this.selectedManga
                     || !this.selectedConnector
                     || this.selectedManga.id !== chapter.manga.id
-                    || (this.selectedConnector.id !== this.bookmarkConnectorID 
-                        && this.selectedConnector.id !== this.clipboardConnectorID 
+                    || (this.selectedConnector.id !== this.bookmarkConnectorID
+                        && this.selectedConnector.id !== this.clipboardConnectorID
```
hypoflex, line 587: same story here, nothing changed nothing added...
there are more places it "fucks" with the code like this

src/web/lib/hakuneko/frontend@classic-dark/chapters.html
```diff
@@ -217,6 +217,7 @@
                 document.addEventListener( EventListener.onChapterStatusChanged, this.onChapterStatusChanged.bind( this ) );
                 window.addEventListener( 'chapterUp', this.onRequestChapterUp.bind( this ) );
                 window.addEventListener( 'chapterDown', this.onRequestChapterDown.bind( this ) );
+                window.addEventListener( 'hideViewer', this.onRequestHideViewer.bind( this ) );
```
hypoflex: added listener to find if viewer is closed

hypoflex (commented):
-

hypoflex:
https://github.com/manga-download/hakuneko/pull/448/commits/fe98c0cb517e55683e700b8ac98039cbd755b8e9 Needs to be reviewed for approval as well for #447 this way the use can choose where to save the 2 files.
default location on default platform place so no further conflict with existing installations.

src/web/lib/hakuneko/frontend@classic-dark/pages.html
```diff
@@ -327,6 +327,7 @@
              */
             hideViewer( event ) {
                 this.set( 'selectedMedia', -1 );
+                window.dispatchEvent( new CustomEvent( 'hideViewer', { detail: this.selectedChapter } ) );
```
27136773: May just set the chapter marker directly:
```javascript
Engine.ChaptermarkManager.addChaptermark( this.selectedChapter );
```
Then no changes in `chapters.html` are required at all

src/web/lib/hakuneko/frontend@classic-light/pages.html
```diff
@@ -327,6 +327,7 @@
              */
             hideViewer( event ) {
                 this.set( 'selectedMedia', -1 );
+                window.dispatchEvent( new CustomEvent( 'hideViewer', { detail: this.selectedChapter } ) );
```
27136773: May just set the chapter marker directly:
```javascript
Engine.ChaptermarkManager.addChaptermark( this.selectedChapter );
```
Then no changes in `chapters.html` are required at all

src/web/lib/hakuneko/frontend@classic-dark/chapters.html
```diff
@@ -94,12 +94,12 @@
             <table border="0" cellpadding="0" cellspacing="0" width="100%">
```
27136773, line 94: No need to change this file at all when doing
https://github.com/manga-download/hakuneko/pull/448/files#r329122054

src/web/lib/hakuneko/frontend@classic-light/chapters.html
```diff
@@ -94,12 +94,12 @@
             <table border="0" cellpadding="0" cellspacing="0" width="100%">
```
27136773, line 94: No need to change this file at all when doing
https://github.com/manga-download/hakuneko/pull/448/files#r329122332

src/web/mjs/engine/Settings.mjs
```diff
@@ -315,4 +336,40 @@ export default class Settings {
 
         }
     }
+
+    /**
+     * HYPOFLEX
+     * Save the given value for the given key in the bookmark storage
+     */
+    saveBookmarks( key, value, indentation ) {
```
27136773: Move method either to `Storage` or to the following both: `BookmarkManager` and `ChaptermarkManager`

src/web/mjs/engine/Settings.mjs
```diff
@@ -315,4 +336,40 @@ export default class Settings {
...
+                }
+            } );
+        } );
+    }
+
+    /**
+     * HYPOFLEX
+     * Load the value for the given key from the bookmark storage
+     */
+    async loadBookmarks( key ) {
```
27136773: Move method either to `Storage` or to the following both: `BookmarkManager` and `ChaptermarkManager`

src/web/mjs/engine/Settings.mjs
```diff
@@ -315,4 +336,40 @@ export default class Settings {
...
         }
     }
+
+    /**
+     * HYPOFLEX
+     * Save the given value for the given key in the bookmark storage
+     */
+    saveBookmarks( key, value, indentation ) {
+        return new Promise( ( resolve, reject ) => {
+            this.fs.writeFile( this.bookmarkDirectory.value + '\\hakuneko.' + key, JSON.stringify( value, undefined, indentation ), function( error ) {
```
27136773: Will `\\` path separator work on linux and macOS?

src/web/mjs/engine/Settings.mjs
```diff
@@ -315,4 +336,40 @@ export default class Settings {
...
+    }
+
+    /**
+     * HYPOFLEX
+     * Load the value for the given key from the bookmark storage
+     */
+    async loadBookmarks( key ) {
+        //return fetch( this.config + key ).then( response => response.json() );
+        return new Promise( ( resolve, reject ) => {
+            this.fs.readFile( this.bookmarkDirectory.value + '\\hakuneko.' + key, 'utf8', ( error, data ) => {
```
27136773: Will `\\` path separator work on linux and macOS?

src/web/mjs/engine/Settings.mjs
```diff
@@ -39,6 +39,9 @@ export default class Settings {
         } catch ( e ) {
             docs = '.';
         }
+        //HYPOFLEX
+        this.fs = require( 'fs' );
```
27136773: Use `fs-extra` instead, it provides everything from `fs` but returns more convenient promises instead of callback functions

src/web/mjs/engine/Settings.mjs
hypoflex: Moved to `Storage`

Reason for moving  to storage instead of keeping it in  settings is to minimize app performance impact?

src/web/mjs/engine/Settings.mjs
hypoflex: i knew this would  maybe  cause an issue
have not tested on linux/osx yet, didnt have my laptop in reach, will test later

src/web/mjs/engine/Settings.mjs
hypoflex: Reason for moving  to storage instead of keeping it in  settings is to minimize app performance impact?

src/web/mjs/engine/Settings.mjs
```diff
@@ -65,6 +68,15 @@ export default class Settings {
...
             value: path.join( docs, 'Mangas' )
         };
+        //HYPOFLEX
+        let electron = require( 'electron' );
+        this.bookmarkDirectory = {
+            label: 'Bookmarks Directory',
+            description: 'The directory where bookmarks file will be stored',
+            input: types.directory,
+            value: path.join( electron.remote.app.getPath( 'userData' ) )
+            // value: path.join( docs )
```
hypoflex, line 79: This cant be tested well running  in npm. when clicking the browser icon both on this new directory browser as the already existing one for manga's it returns a {Promise Object} in the field, how can i  get around this.

right  now ive been just  switching the value within the code to test.

src/web/mjs/engine/Settings.mjs
hypoflex: the `fs` function i grabbed from the `storage` file, where `fs` is still used, maybe good to change it there to `fs-extra` as well?

src/web/mjs/engine/Settings.mjs
hypoflex: i knew this would maybe cause an issue
have not tested on linux/osx yet, didnt have my laptop in reach, will test later

src/web/lib/hakuneko/frontend@classic-dark/chapters.html
```diff
@@ -436,10 +437,16 @@
                     return;
                 }
                 this.set( 'selectedChapter', e.model.item );
+                Engine.ChaptermarkManager.addChaptermark( this.selectedChapter );
```
hypoflex, line 439: this line still "needed" though. its used to when preview is clicked  the chapter mark is updated.
So for example:
 when opening Tokyo ghoul and having no  chaptermark, and starting at chapter 100
the marker will  be  set there.

on the other hand i could see confusion happening with this if a user desides to go back a chapter the marker also gets updated.

hypoflex: Does not seem to work on OSX.

will first convert other changes before testing on other platform again

src/web/lib/hakuneko/frontend@classic-dark/pages.html
hypoflex: Changes made

src/web/mjs/engine/Settings.mjs
hypoflex: Moved to `Storage`

hypoflex:
Going to test changes on OSX now

src/web/lib/hakuneko/frontend@classic-dark/chapters.html
27136773: Removing the trailing space seems correct to me, this is in compliance with the current lint configuration.
Unfortunately linting is not yet supported for HTML files, so thats the reason the trailing whitespace was not detected as error.
Your IDE may see the lint configuration and adjust the errors automatically according to the configuration, which is correct.

src/web/lib/hakuneko/frontend@classic-dark/chapters.html
27136773: Your change is correct, trailing white-spaces must be removed according to the lint configuration.

src/web/lib/hakuneko/frontend@classic-light/pages.html
```diff
@@ -327,6 +327,7 @@
              */
             hideViewer( event ) {
                 this.set( 'selectedMedia', -1 );
+                Engine.ChaptermarkManager.addChaptermark( this.selectedChapter );
```
27136773, line 330: It is up to you to decide which actions should update the chapter marker, the following candiates may be possible:

- `showViewer`
- `hideViewer`
- `requestChapterUp`
- `requestChapterDown`

I think your current solution is fine, chapter mark will be updated when the reader is closed (which probably always is the case, except the program is closed while reading or crashes for some reason)

src/web/mjs/engine/Storage.mjs
```diff
@@ -869,4 +878,40 @@ export default class Storage {
...
         return undefined;
     }
+
+    /**
+     * HYPOFLEX
+     * Save the given value for the given key in the bookmark storage
+     */
+    saveBookmarks( key, value, indentation ) {
+        return new Promise( ( resolve, reject ) => {
+            this.fs.writeFile( Engine.Settings.bookmarkDirectory.value + '\\hakuneko.' + key, JSON.stringify( value, undefined, indentation ), function( error ) {
```
27136773: I think storage provides `path` or `this.path` so your separator issue can now be solved with `path.join(Engine.Settings.bookmarkDirectory.value, 'hakuneko.' + key)`

hypoflex: `\\hakuneko` is indeed windows only, since windows needs a backslash and linux/osx needs a forward slash.

i was working on a switch case based on variable

```javascript
let platform = undefined;
switch(this.platform.indexOf){
case 'win':
 platform = '\\hakuneko'
 break;
case 'darwin':
 platform = '/hakuneko'
break;
}
```
(above code does not work as intended yet)

i will try what you posted, see what that results in

src/web/mjs/engine/Settings.mjs
27136773: I forgot to fix the bug in dark theme.
Done now.
https://github.com/manga-download/hakuneko/commit/618ecde29a464e6a838fe03fb6c04738317f11d7

src/web/mjs/engine/Storage.mjs
```diff
@@ -869,4 +878,40 @@ export default class Storage {
...
+    }
+
+    /**
+     * HYPOFLEX
+     * Load the value for the given key from the bookmark storage
+     */
+    async loadBookmarks( key ) {
+        //return fetch( this.config + key ).then( response => response.json() );
+        return new Promise( ( resolve, reject ) => {
+            this.fs.readFile( Engine.Settings.bookmarkDirectory.value + '\\hakuneko.' + key, 'utf8', ( error, data ) => {
```
27136773: I think storage provides `path` or `this.path` so your separator issue can now be solved with `path.join(Engine.Settings.bookmarkDirectory.value, 'hakuneko.' + key)`

src/web/mjs/engine/Settings.mjs
27136773: Yes, good idea

src/web/mjs/engine/Settings.mjs
27136773: No its more to honor clean code principle (class responsibility), settings shall not handle bookmark and chapter-mark I/O.

hypoflex:
- edited `pages.html` to nolonger require `chapter.html`
- changed `fs` to `fs-extra`
- moved `loadbookmark` and `savebookmark` to `Storage` instead of `Settings`
- changed the `Writefile` function to use `path.join` so separator works on all systems
-  folder async fixed by ronny https://github.com/manga-download/hakuneko/commit/618ecde29a464e6a838fe03fb6c04738317f11d7

src/web/mjs/engine/BookmarkManager.mjs
```diff
@@ -145,7 +146,7 @@ export default class BookmarkManager extends EventTarget {
      * Callback will be provided with an error (or null if no error).
      */
     saveProfile( profile, callback ) {
```
27136773, line 149: Will bookmarks be saved after changing the directory in settings?
If not they may be lost when the directory is changed and the application is closed right after.

If not it might be a good idea to save the bookmarks whenever the settings changed event is fired
Example: https://github.com/manga-download/hakuneko/blob/618ecde29a464e6a838fe03fb6c04738317f11d7/src/web/mjs/connectors/NineAnime.mjs#L46

src/web/mjs/engine/ChaptermarkManager.mjs
```diff
@@ -35,7 +35,8 @@ export default class ChaptermarkManager extends EventTarget {
      * Will reset chaptermarks when saving fails.
      */
     _syncChaptermarks( callback ) {
-        Engine.Storage.saveConfig( 'chaptermarks', this.chaptermarks, 2 )
+        //HYPOFLEX
+        Engine.Storage.saveBookmarks( 'chaptermarks', this.chaptermarks, 2 )
```
27136773, line 40: Will chaptermarks be saved after changing the directory in settings?
If not they may be lost when the directory is changed and the application is closed right after.

If not it might be a good idea to save the chaptermarks whenever the settings changed event is fired
Example: https://github.com/manga-download/hakuneko/blob/618ecde29a464e6a838fe03fb6c04738317f11d7/src/web/mjs/connectors/NineAnime.mjs#L46

src/web/mjs/engine/BookmarkManager.mjs
hypoflex: good point!
looking into it

src/web/mjs/engine/BookmarkManager.mjs
```diff
@@ -113,7 +115,6 @@ export default class BookmarkManager extends EventTarget {
      * Callback will be provided with an error (or null if no error).
      */
     loadProfile( profile, callback ) {
-        //HYPOFLEX
```
hypoflex: removed the name tags, these were mainly used to quick search without knowning the exact function name.

src/web/mjs/engine/Settings.mjs
```diff
@@ -286,6 +283,7 @@ export default class Settings {
         Engine.Storage.saveConfig( 'settings', data, 2 )
             .then( () => {
                 document.dispatchEvent( new CustomEvent( EventListener.onSettingsChanged, { detail: this } ) );
+                document.dispatchEvent( new CustomEvent( EventListener.onSettingsSaved, { detail: this } ) );
```
hypoflex, line 285: I had to add a new event listener here. 
the `onSettingsChanged` is being created twice, once here `line 285` and above `line 255`.

`Line 255` is being called once hakuneko is started.
`(Line 233 in current master)`
https://github.com/manga-download/hakuneko/blob/9a52af9cf2605bdb922a61b323f3b06ddb968220/src/web/mjs/engine/Settings.mjs#L233
problem with that is, if I bind the `savebookmark` function to `onSettingsChanged`
it will save a empty bookmark, since its not loaded yet.

src/web/mjs/engine/ChaptermarkManager.mjs
```diff
@@ -8,6 +8,8 @@ export default class ChaptermarkManager extends EventTarget {
     constructor() {
         super();
         this.chaptermarks = [];
+
+        document.addEventListener( EventListener.onSettingsSaved, this._syncChaptermarks.bind( this ) );
```
hypoflex, line 12: Created new listener `onSettingsSaved` instead of `onSettingsChanged`, read comment below

src/web/mjs/engine/BookmarkManager.mjs
```diff
@@ -13,6 +13,8 @@ export default class BookmarkManager extends EventTarget {
         super();
         this.bookmarks = [];
         this._bookmarkImporter = bookmarkImporter;
+
+        document.addEventListener( EventListener.onSettingsSaved, this.saveProfile.bind( this ) );
```
hypoflex, line 17: Created new listener onSettingsSaved instead of onSettingsChanged, read comment below

src/web/mjs/engine/Settings.mjs
```diff
@@ -254,7 +254,8 @@ export default class Settings {
             } );
             document.dispatchEvent( new CustomEvent( EventListener.onSettingsChanged, { detail: this } ) );
         } catch( error ) {
-            console.error('Failed to load HakuNeko settings!', error);
+            console.error('Failed to load HakuNeko settings! New settings file created.', error);
+            this.saveProfile();
```
hypoflex: `saveProfile` if settings file is not found? resolves #451

src/web/mjs/engine/ChaptermarkManager.mjs
27136773: Unfortunately i hadn't time yet to refactor `Settings.mjs` to the new event target system, so it seems you need to use the deprecated document based system to use your new event.
In that case you also need to declare a string for this event in [Enums.mjs](https://github.com/manga-download/hakuneko/blob/9a52af9cf2605bdb922a61b323f3b06ddb968220/src/web/mjs/engine/Enums.mjs#L8), where you have do define a string for `EventListener.onSettingsSaved`, otherwise this is `undefined` which may lead to problems.

src/web/mjs/engine/BookmarkManager.mjs
27136773: Unfortunately i hadn't time yet to refactor `Settings.mjs` to the new event target system, so it seems you need to use the deprecated document based system to use your new event.
In that case you also need to declare a string for this event in [Enums.mjs](https://github.com/manga-download/hakuneko/blob/9a52af9cf2605bdb922a61b323f3b06ddb968220/src/web/mjs/engine/Enums.mjs#L8), where you have do define a string for `EventListener.onSettingsSaved`, otherwise this is `undefined` which may lead to problems.

src/web/mjs/engine/Settings.mjs
```diff
@@ -65,6 +67,18 @@ export default class Settings {
             input: types.directory,
             value: path.join( docs, 'Mangas' )
         };
+        let electron = require( 'electron' );
```
27136773: You may place this at the top and then change the `app` assignment to use this variable instead to avoid double `require` statements
```javascript
let electron = require( 'electron' );
let app = electron.remote.app;
```

src/web/mjs/engine/Settings.mjs
```diff
@@ -65,6 +67,18 @@ export default class Settings {
...
+        this.bookmarkDirectory = {
+            label: 'Bookmarks Directory ⁽¹⁾',
+            description: [
+                'The directory where bookmarks file will be stored',
+                '',
+                '⁽¹⁾ Restart required to take affect',
+            ].join('\n'),
+            input: types.directory,
+            value: path.join( electron.remote.app.getPath( 'userData' ) )
```
27136773: use `this.app` instead of `electron.remote.app`

27136773: The statement `path.join` seems obsolete if there is only a single parameter assigned => nothing to join

src/web/mjs/engine/Settings.mjs
hypoflex: ahh yes, app was already declared there, electron was probably automatically declared by the IDE
since you recommended that for the `getPath`

src/web/mjs/engine/Settings.mjs
hypoflex: you mean `app` cause `this.app` is seen as not defined (surprisingly) by the IDE

src/web/mjs/engine/Settings.mjs
```diff
@@ -232,7 +254,8 @@ export default class Settings {
             } );
             document.dispatchEvent( new CustomEvent( EventListener.onSettingsChanged, { detail: this } ) );
         } catch( error ) {
-            console.error('Failed to load HakuNeko settings!', error);
+            console.error('Failed to load HakuNeko settings! New settings file created.', error);
+            this.saveProfile();
```
27136773: This should be done for the expected condition (`File does not Exist`) somewhere in the try block.
This catch block catches any exception, including errors like `Out of Memory`, `Permission Denied`, ...
calling `saveProfile` may lead to further re-throw errors.

src/web/mjs/engine/Storage.mjs
```diff
@@ -869,4 +877,37 @@ export default class Storage {
         }
         return undefined;
     }
+
+    /**
+     * Save the given value for the given key in the bookmark storage
+     */
+    saveBookmarks( key, value, indentation ) {
+        return new Promise( ( resolve, reject ) => {
+            this.fs.writeFile( this.path.join(Engine.Settings.bookmarkDirectory.value, 'hakuneko.' + key), JSON.stringify( value, undefined, indentation ), function( error ) {
```
27136773: For consistency: Either use `this._bookmarkOutpotPath()` instead of `Engine.Settings.bookmarkDirectory.value` or keep as is and remove `this._bookmarkOutpotPath()`.

src/web/mjs/engine/Storage.mjs
```diff
@@ -869,4 +877,37 @@ export default class Storage {
...
+            } );
+        } );
+    }
+
+    /**
+     * Load the value for the given key from the bookmark storage
+     */
+    async loadBookmarks( key ) {
+        return new Promise( ( resolve, reject ) => {
+            this.fs.readFile( this.path.join(Engine.Settings.bookmarkDirectory.value, 'hakuneko.' + key), 'utf8', ( error, data ) => {
```
27136773: For consistency: Either use `this._bookmarkOutpotPath()` instead of `Engine.Settings.bookmarkDirectory.value` or keep as is and remove `this._bookmarkOutpotPath()`.

src/web/mjs/engine/Storage.mjs
```diff
@@ -720,6 +720,14 @@ export default class Storage {
         } );
     }
 
+    /**
+     * Helper function to generate the path where the bookmarks and markers are stored.
+     */
+    _bookmarkOutputPath( ) {
```
27136773: May make this a `getter`, so it can be used like a member and not like a function

27136773 (approved):
Some room left for improvements, but generally it should work like expected.

src/web/mjs/engine/Settings.mjs
hypoflex: ahh alright, i figured it was for loading the settings only

27136773:
I will try to get the new deployment pipeline running before merging this change

src/web/mjs/engine/Settings.mjs
27136773: Yes you IDE is correct, its local variable

src/web/mjs/engine/Storage.mjs
```diff
@@ -723,9 +723,8 @@ export default class Storage {
     /**
      * Helper function to generate the path where the bookmarks and markers are stored.
      */
-    _bookmarkOutputPath( ) {
-        let output = Engine.Settings.bookmarkDirectory.value;
-        return output;
+     get _bookmarkOutputPath() {
+        return Engine.Settings.bookmarkDirectory.value;
     }
```
hypoflex, line 728: get and set in JS is new one for me, this should be correct to what you meant

src/web/mjs/engine/Storage.mjs
```diff
@@ -883,7 +882,7 @@ export default class Storage {
      */
     saveBookmarks( key, value, indentation ) {
         return new Promise( ( resolve, reject ) => {
-            this.fs.writeFile( this.path.join(Engine.Settings.bookmarkDirectory.value, 'hakuneko.' + key), JSON.stringify( value, undefined, indentation ), function( error ) {
+            this.fs.writeFile( this.path.join(Storage._bookmarkOutputPath.value, 'hakuneko.' + key), JSON.stringify( value, undefined, indentation ), function(error ) {
```
hypoflex: that should be `this._bookmarkOutputPath`

src/web/mjs/engine/Storage.mjs
```diff
@@ -898,7 +897,7 @@ export default class Storage {
      */
     async loadBookmarks( key ) {
         return new Promise( ( resolve, reject ) => {
-            this.fs.readFile( this.path.join(Engine.Settings.bookmarkDirectory.value, 'hakuneko.' + key), 'utf8', ( error, data ) => {
+            this.fs.readFile( this.path.join(Storage._bookmarkOutputPath.value, 'hakuneko.' + key), 'utf8', (error, data ) => {
```
hypoflex: that should be `this._bookmarkOutputPath`

src/web/mjs/engine/Storage.mjs
```diff
@@ -869,4 +876,37 @@ export default class Storage {
         }
         return undefined;
     }
+
+    /**
+     * Save the given value for the given key in the bookmark storage
+     */
+    saveBookmarks( key, value, indentation ) {
+        return new Promise( ( resolve, reject ) => {
+            this.fs.writeFile( this.path.join(Storage._bookmarkOutputPath.value, 'hakuneko.' + key), JSON.stringify( value, undefined, indentation ), function(error ) {
```
27136773: The getter is not static, so it needs to be accessed via instance (`this`), also the getter already returns the `value` property, simply `this._bookmarkOutputPath` will do the trick

src/web/mjs/engine/Storage.mjs
```diff
@@ -869,4 +876,37 @@ export default class Storage {
...
+            } );
+        } );
+    }
+
+    /**
+     * Load the value for the given key from the bookmark storage
+     */
+    async loadBookmarks( key ) {
+        return new Promise( ( resolve, reject ) => {
+            this.fs.readFile( this.path.join(Storage._bookmarkOutputPath.value, 'hakuneko.' + key), 'utf8', (error, data ) => {
```
27136773: The getter is not static, so it needs to be accessed via instance (`this`), also the getter already returns the `value` property, simply `this._bookmarkOutputPath` will do the trick

hypoflex: wouldn't be better to have the getter return the object `Engine.Settings.bookmarkDirectory` instead of only the value? that way it has more potential use

27136773: Its up to you to decide how this should be used

Issue #470, opened 2019-10-01
Settings undefined


hypoflex:
running latest source version, no console log about it 
https://github.com/manga-download/hakuneko/commit/794446a3b3137d3bf73a31cd65ab36acbaff9a2c
![image](https://user-images.githubusercontent.com/36196198/66002292-ab860500-e4a3-11e9-9a5d-5819452ca55c.png)

hypoflex:
seems to be coming from my commit, checked last 3 commits https://github.com/manga-download/hakuneko/commit/c2f479cca0a9191d99afad1f781c6bdf3163dea0 does not seem to have it

https://github.com/manga-download/hakuneko/commit/83882b8aaa5eccee99b1fab2118585840f89c7c7 https://github.com/manga-download/hakuneko/commit/794446a3b3137d3bf73a31cd65ab36acbaff9a2c both have it

hypoflex:
`this.fs = require( 'fs-extra' );` seems to be the one causing it.
https://github.com/manga-download/hakuneko/commit/83882b8aaa5eccee99b1fab2118585840f89c7c7#diff-05f50b790108df767965ed64aba734a2R42

this is still a left over of https://github.com/manga-download/hakuneko/pull/448#pullrequestreview-294410758 which eventually got scraped with https://github.com/manga-download/hakuneko/pull/448#pullrequestreview-294406624

Issue #469, opened 2019-10-01
Preview should not set chapter marker


27136773:
When clicking the preview button for a certain chapter to show its images, the chapter mark will be assigned to this chapter.
It would be expected that the chapter marker would only be assigned through the reader (open, close or change chapter up/down).

Occurs in development version 83882b8a

hypoflex:
Correct, I've done this with intention.
Reason for doing it like this out of personal prefrence, since i read on multiple platforms (Windows, OSX, IOS (mangarock))

so getting back on a different platform i occasionally press preview to see if that's the correct chapter i left off on my other device.

Guess with bookmark file being able to locate somewhere else now this option can be scraped
and like you said yourself, it also put it once the viewer is closed
https://github.com/manga-download/hakuneko/commit/83882b8aaa5eccee99b1fab2118585840f89c7c7#diff-6dc48bd82fa99890cd7410d38f3f0ea6R439

https://github.com/manga-download/hakuneko/commit/83882b8aaa5eccee99b1fab2118585840f89c7c7#diff-f5eaf4cdf511723135b827d6c5326dd1R439

PR #471, opened 2019-10-01
Small hotfix 

hypoflex:
updated some files based on #470 and #469


hypoflex:
Question, its pointing out all commits from before.
now I did reset my branch/fork to your last commit before my last commit. (so no issue's should occur)

would it be better to refork every time or is this fine as well?

27136773 (approved):
Looks good

PR #505, opened 2019-10-09
Introduction of settings events instead of global document events

27136773:


hypoflex (approved):
did quick look through it, kind busy lately, would love to help out again.
looks neat. lot more efficient

PR #561, opened 2019-10-21
[DynastyScans] Use JSON API

medmig:
Use JSON API instead of scraping HTML.


src/web/mjs/connectors/DynastyScans.mjs
```diff
@@ -26,14 +26,17 @@ export default class DynastyScans extends Connector {
...
         let promises = ['/series', '/anthologies', '/issues', '/doujins'].map( page => {
-            return this.fetchDOM( this.url + page, '.tag-list dd a' )
-                .then( data => {
-                    let mangaList = data.map( element => {
-                        return {
-                            id: this.getRelativeLink( element ),
-                            title: element.text.trim()
-                        };
-                    } );
+            return this.fetchJSON( this.url + `${page}.json` )
```
27136773: Tip: In the current HakuNeko versions you can also use async/await pattern, which makes the promises more readable.
Example: https://github.com/manga-download/hakuneko/wiki/Developer-Website-Tutorial#47-from-callbacks-over-promises-to-asyncawait

src/web/mjs/connectors/DynastyScans.mjs
```diff
@@ -52,12 +55,13 @@ export default class DynastyScans extends Connector {
...
     _getChapterList( manga, callback ) {
-        this.fetchDOM( this.url + manga.id, '.chapter-list dd a[class="name"]' )
-            .then( data => {
-                let chapterList = data.map( element => {
-                    let title = element.text.replace( manga.title, '' ).trim();
+        this.fetchJSON( this.url + manga.id + '.json' )
+            .then( json => {
+                // filter removes the "header" elements from the list
+                let chapterList = json['taggings'].filter( chapter => !chapter.header ).map( chapter => {
+                    let title = chapter.title.replace( manga.title, '' ).trim();
```
27136773: Does the json data still has the manga title within the chapter title and needs to be replaced?

src/web/mjs/connectors/DynastyScans.mjs
```diff
@@ -82,16 +86,9 @@ export default class DynastyScans extends Connector {
...
-                    throw new Error( `Failed to receive page list (status: ${response.status}) - ${response.statusText}` );
-                }
-                return response.text();
-            } )
-            .then( data => {
-                let pages = data.match( /var\s+pages\s*=\s*(\[.*?\])\s*;/ )[1];
-                let pageList = JSON.parse( pages ).map( p => this.url + p.image );
+        this.fetchJSON( this.url + chapter.id + '.json' )
+            .then( json => {
+                let pageList = json['pages'].map( page => this.url + page.url );
```
27136773: May use `new URL(page.url, this.url).href`, this will cover the case that page.url already is an absolute link

27136773 (approved):
Not tested, but source looks good

src/web/mjs/connectors/DynastyScans.mjs
medmig: Maybe not, it's the same as in HTML pages. Actually I don't really remember why it was like that in the first place.

src/web/mjs/connectors/DynastyScans.mjs
medmig: Noted, thanks.
I won't have time to make the modifications tonight, I'll do it ASAP and push again !

src/web/mjs/connectors/DynastyScans.mjs
medmig: Good idea, thanks. I'll update this ASAP too.

src/web/mjs/connectors/DynastyScans.mjs
medmig: Quickly searched but didn't find any example where this would apply.
However I found [an example](https://dynasty-scans.com/anthologies/an_absurd_relationship_collection) of why it might have been done this way. The same series is also under another format [here](https://dynasty-scans.com/series/an_absurd_relationship).
I'll simplify it since it can easily be restored later if needed anyway.

EDIT: Found [an example](https://dynasty-scans.com/anthologies/mermaid_line) where it was applicable. Still, not sure it is really important or bothering anyway...

PR #730, opened 2019-11-30
[Kuman5] added website support

dodieboy:
Everything work until the picture retrieve part.
When downloading, it will show "TypeError: Failed to fetch".

put it can get information about number of page in that chapter


27136773:
There are some eslint errors, please use `npm run lint`to spot and fix them

27136773 (approved):
Looks good

PR #785, opened 2019-12-20
New Server

marrokin:
Add new server (vcp)


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
```
27136773: This is a user specific file, add it to `.gitignore` and do not check in

src/web/mjs/connectors/AnimeYT.mjs
```diff
@@ -0,0 +1,24 @@
+import Connector from '../engine/Connetors.mjs';
```
27136773: Implementation of AnimeYT is incomplete, do not check in

src/web/mjs/connectors/VerComicsPorno.mjs
```diff
@@ -0,0 +1,107 @@
...
+export default class VerComicsPorno extends Connector {
+
+    /**
+     *
+     */
+    constructor() {
+        super();
+        super.id = 'vercomicsporno';
+        super.label = 'VerComicsPorno';
+        this.tags = ['manga', 'comics' ,'spanish', 'english'];
```
27136773: From the name i would imply that the tag should be `hentai` instead of manga/comic

27136773:
May only check in relevant changes (e.g. gitignore, VerComicsPorno).
May use the new API as described in the tutorial instead of the old API methods.

src/web/mjs/connectors/VerComicsPorno.mjs
```diff
@@ -0,0 +1,59 @@
...
+            };
+        }).filter(manga => manga.id.startsWith('/'));
+    }
+
+    async _getMangas() {
+        let mangaList = [];
+        let request = new Request(this.url + this.path + '1' , this.requestOptions);
+        let data = await this.fetchDOM(request, this.pager);
+        let pageCount = parseInt(data[0].href.match(/\d+$/));
+        console.log(pageCount);
```
27136773: Remove debug output before check-in

src/web/mjs/connectors/VerComicsPorno.mjs
```diff
@@ -0,0 +1,59 @@
...
+        for(let page = 1; page <= pageCount; page++) {
+            let mangas = await this._getMangasFromPage(page);
+            mangaList.push(...mangas);
+        }
+        return mangaList;
+    }
+
+    async _getChapters(manga) {
+        console.log(manga);
```
27136773: Remove debug output before check-in

27136773:
I removed all files not meant to be checked-in right now from the pull request => ready for review

27136773 (approved):
Looks good after some adjustments

Issue #804, opened 2019-12-22
https://www.mangaae.com


satyogi:
hakuneko-desktop_6.1.5_windows-portable_amd64
all chapter is not working its the same for azoramanga

PR #808, opened 2019-12-24
[mangaArab] fixed website support (#804)

dodieboy:


27136773 (changes requested):
This will be fixed by PR #827 
Its safer to change the domain instead of dealing with FQDNs (especially if there are already existing bookmarks with relative paths)

27136773:
Handled by #827

PR #871, opened 2020-01-07
[bacamanga] fixed image link extraction

27136773:


27136773:
References fmd-project-team/FMD#719

Issue #922, opened 2020-01-12
Can't get manga from MangaKakalot sites


Iloveboobs84:
I tried again after the Nightly version download and Manga are still missing chapters from these sites 
[MangaDex] and [MangaKakkalot]

Missing Manga list 
1) Boku no Kanojo Sensei
2) Fairy Tail 100yr quest
3) Freezing
4) Goblin Slayer Year One
5) Go Touban (aka 5 Sister manga)
6) Kenja no Mango
7) Maou Gakuin no Futekigousha
8) Nanatsu no Taizai
9) Tate no Yuusha no Narigari

I'm sure there are more missing but these are the ones I was looking for.

If there not missing chapters than my Windows 10 is not letting it updates and How would I fix it to update.

ronny1982:
Have you tested if these are fully available on the website?
HakuNeko only shows what is available on the website.

dodieboy:
Tested, these are fully available on the website
![image](https://user-images.githubusercontent.com/9015764/72229276-46849400-35e8-11ea-9d0b-ed1d12ff60e0.png)
![image](https://user-images.githubusercontent.com/9015764/72229282-5308ec80-35e8-11ea-8dfa-e62eda821e8d.png)

Iloveboobs84:
I test all manga websites to see if the Chapters are AVAILABLE before trying to download them from HakunNeko. Currently On Websites but Unavailable in the download programs with ZERO Chapters.

Iloveboobs84:
As of right NOW, I checked for the 2nd time today and these are the Updates for mangadex from the list not shown in Hakuneko Desktop

Also, mangadex changed from .com to .cc might be the reason it's not working. Doman needs to be updated in the Downloaders Code. 

Boku no Kanojo Sensei - updated to Ch 09
Fairy Tail 100yr quest - updated to Ch 45
Freezing - updated to Ch 225
5Touban no Hanayome - updated to Ch 116
Kenja no Mango - updated to Ch 36
Maou Gakuin no Futekigousha - updated to Ch 13.2
Nanatsu no Taizai - updated to Ch 335
Tate no Yuusha no Nariagari {aka Shield Hero} - updated to Ch 61

Also No listing for Fairy Tail Happy's Great Adventure Side story to manga list

Goblin Slayer Year One is - updated to Ch 34 on Manganelo.com previously Mangakakalot

ronny1982:
For MangaDex See #885

Iloveboobs84:
[...]

PR #924, opened 2020-01-13
[manganel, mangakakalot, mangabat] view mangas/chapters from cross-references

27136773:
show a popup stating that a cross-reference can not be loaded and a different website should be used

References #922


Sheepux (approved):
Yup yup, good old alert doing the job.

Issue #1218, opened 2020-03-01
[Japscan] Error "Cannot read property 'src' of null"


nzoba:
Hello,

The problem [here](https://github.com/manga-download/hakuneko/commit/fc9aee2065284d738d1fe341f6ecea65ad78cd0e#commitcomment-37564700) that didn't display all the pages is solved thank you but I still have another problem : I have the error "Cannot read property 'src' of null" for some mangas like Psycho Bank chapter 22, Time Shadows 42 and Blizzard Axel chapter 43.

![Error HakuNeko](https://user-images.githubusercontent.com/55888232/75626328-fdec5e80-5bc6-11ea-9730-48fbbb2cbded.PNG)

Thanks again.

27136773:
Looks like new protection, needs to be analyzed

27136773:
Added provisional brute-forcing technique for new chapters which uses scrambled images.
This requires a lot of resources on HakuNeko and also on JapScan servers, so better use these advises:

- avoid using the preview for not downloaded chapters (download first, then read)
- if connection issues occur, try to increase the throttling for JapScan in settings
- be patient when browsing manga/chapters from JapScan while also downloading from JapScan (wait until stuff has been loaded)
- do not synchronize the manga list and download chapters at the same time

PR #1225, opened 2020-03-01
[japscan] brute-forcing scrambled images for new chapters

27136773:
Hey JapScan team... look i cracked your stuff again :3


27136773:
References #1218

PR #1293, opened 2020-03-10
FEAT-1292

### LINKED ISSUE #1292: Page viewer retry when error on loading image
When using the image viewer sometimes the img url fails to load because of the website itself (exemple random http 520).
Implement a retry system on error (max 3 times)

Sheepux:
Fixes #1292 by implementing a dom retry (img onError)
Cannot fix the preview (div background url) as this html element doesn't trigger error events


27136773 (approved):
Looks good

Issue #252, opened 2019-07-09
Website Request: mag-garden


hakuneko-user:
> **From:** AthK

https://comic.mag-garden.co.jp/
http://kansai.mag-garden.co.jp/

Pretty simple to download images because they are sequencially placed.

For example: 
For Chapter
https://comic.mag-garden.co.jp/assets/files/fumisan190630/HTML5/pc.html

Images are located as below

Page1: 
https://comic.mag-garden.co.jp/assets/files/fumisan190630/books/images/2/1.jpg
Page2 : 
https://comic.mag-garden.co.jp/assets/files/fumisan190630/books/images/2/2.jpg
Page3 : 
https://comic.mag-garden.co.jp/assets/files/fumisan190630/books/images/2/3.jpg

And so on...

27136773:
Both websites are using WordPress (though templates seems to be different).
Both websites are using ActiBook reader.

Image links can be extracted by injecting the following script into the [page](http://kansai.mag-garden.co.jp/assets/files/ee1/HTML5/pc.html#/page/1):
```javascript
[ ...( new Array( App.book.lastPage ) ).keys() ].map( page => new URL( `${App.book.hd.get( 'pieceDirectory' )}/${page + 1}.jpg`, window.location.href ).href );
```

newaccforhc:
This issue is from me only.. haha
Well... comic.mag-garden.co.jp is now magcomi.. And, already supported... So, need to add support for http://kansai.mag-garden.co.jp/

Issue #89, opened 2019-05-04
Add magkan


Ever4engel:
http://kansai.mag-garden.co.jp/

27136773:
See: #252

PR #1361, opened 2020-03-20
added ComicFire (HobbyJapan), Yawaraka Spirits, MagComi (Mag Garden)

ghost:
ComicFire:
-had to modify the SpeedBinb.mjs template due to the website's chapters being hosted on binb's own site -- thus, 'this.url' has no relevance and 'chapterID' must be used instead; just added simple checks for what site is being used and had to modify one method header to include the chapterID/chapter URL
-the first attempt to download a chapter from ComicFire (seemingly per session) will always fail, with subsequent requests to download chapters succeeding. this seems to be because the request to access the first chapter's API url fails, however I not sure as to why. my best guess is it may need to establish some sort of session before trying to access it.


src/web/mjs/connectors/ComicFire.mjs
```diff
@@ -0,0 +1,79 @@
...
+            language: chapter.language
+        };
+        if(chapter.id.includes("SWF")){ //legacy format; older chapters may use this
+            ch.id = "http://tachiyomi.yomeru-hj.net/comic/" + ch.id.substr(17).replace("_SWF_Window.html","");
+            let request = new XMLHttpRequest();
+            request.onreadystatechange = function() {
+                if(this.readyState == 4 && this.status == 200) {
+                    let xmlDOM = request.responseXML;
+                    let pageList = [];
+                    let n = parseInt(xmlDOM.getElementsByTagName("total")[0].textContent);
```
27136773: Maybe this complete request could be done using `this.fetchDOM(url, 'total:first-of-type')`

src/web/mjs/connectors/templates/SpeedBinb.mjs
```diff
@@ -20,7 +20,11 @@ export default class SpeedBinb extends Connector {
                 data = data[0];
                 if( data.dataset['ptbinb'] && data.dataset['ptbinbCid'] )
                 {
-                    return this._getPageList_v016113( chapter.id + '&cid=' + data.dataset['ptbinbCid'], data.dataset.ptbinb );
```
27136773, line 23: We should use URL object to handle URIs instead of dealing with manual string composition.
```javascript
let uri = new URL(chapter.id, this.url);
uri.searchParams.set('cid', data.dataset['ptbinbCid']);
return this._getPageList_v016113(uri.pathname + uri.search, data.dataset.ptbinb);
```

src/web/mjs/connectors/templates/SpeedBinb.mjs
```diff
@@ -160,22 +164,26 @@ export default class SpeedBinb extends Connector {
     _getPageList_v016130( chapterID, apiURL ) {
```
27136773, line 163: either make a new method which implements the hobbyjapan specific behavior, or overwrite this method in hobbyjapan connector itself

src/web/mjs/connectors/templates/SpeedBinb.mjs
```diff
@@ -160,22 +164,26 @@ export default class SpeedBinb extends Connector {
...
-                return this._getPageLinks_v016130( data.items[0], sharingKey );
+                return this._getPageLinks_v016130( data.items[0], sharingKey, chapterID );
             } );
     }
 
     /**
      *
      */
-    _getPageLinks_v016130( configuration, sharingKey ) {
+    _getPageLinks_v016130( configuration, sharingKey, chapterID ) {
```
27136773: Do not change this method, instead adjust the configuration parameter when this method is called

src/web/mjs/connectors/templates/SpeedBinb.mjs
```diff
@@ -160,22 +164,26 @@ export default class SpeedBinb extends Connector {
...
+        else 
+            uri = new URL( apiURL, this.url );
         uri.searchParams.set( 'cid', cid );
         uri.searchParams.set( 'dmytime', Date.now() );
         uri.searchParams.set( 'k', sharingKey );
         let request = new Request( uri.href, this.requestOptions );
         return fetch( request )
             .then( response => response.json() )
             .then( data => {
-                return this._getPageLinks_v016130( data.items[0], sharingKey );
```
27136773: Set the specific `ContentsServer` property of `data.items[0]` before calling `this._getPageLinks_v016130`

Ever4engel:
there is two duplitates (requests): 
[issue 89](https://github.com/manga-download/hakuneko/issues/89) and [issue 252](https://github.com/manga-download/hakuneko/issues/252)

src/web/mjs/connectors/ComicFire.mjs
```diff
@@ -58,21 +58,17 @@ export default class ComicFire extends SpeedBinb {
...
+                let total = parseInt(element[0].textContent);
+                let pageList = [];
+                for(let i = 1; i <= total; i++){
+                    pageList.push(ch.id + "books/images/3.5/" + i + ".jpg");
                 }
-            };
-            request.open("GET", ch.id + 'books/db/book.xml', true);
-            request.send();
+                callback(undefined, pageList);
+            });
```
27136773: Also chain a catch to log the error to console and perform a callback with the error
```javascript
.catch(error => {
    console.error(error, chapter);
    callback(error, undefined);
});
```

27136773 (changes requested):
Seems good so far, but the old HakuNeko API functions have no outer error handling, so i would recommend to add the catch chain as written in the comments

27136773 (approved):
Looks good

PR #1392, opened 2020-03-24
removes period from end of folder names for linux

### LINKED ISSUE #1385: Can't create folder ending in period
When attempting to download a folder witha name ending in a period it will fail.
Running manjaro linux up to date. an option to edit folder names for just a filter that strips out illegal characters would be nice

tested with 'gaikotsu shotenin hondaa san vol 1 ch 1' kissmanga

yobleck:
fixes error where folder can not be downloaded because name of folder ends with period "."

EDIT:
Resolves #1385


yobleck:
This is my first attempt at a pull request sorry if its trash

27136773:
Please fix the lint error (shown by `npm run lint`)

src/web/mjs/engine/Storage.mjs
```diff
@@ -802,6 +802,9 @@ export default class Storage {
             return path.replace( /\.+$/g, '' ).trim(); // remove trailing dots and whitespaces
         }
         if( this.platform.indexOf( 'linux' ) === 0 ) {
+            if(path.substring(path.length-1,path.length)==".") { //check if last character is a period
```
27136773: This seems to be not robust enough, what if there are 2 or more dots at the end
Maybe use a regex instead, something like this:
`path.replace(/[\.\s]+$/, '')`

src/web/mjs/engine/Storage.mjs
```diff
@@ -802,6 +802,9 @@ export default class Storage {
             return path.replace( /\.+$/g, '' ).trim(); // remove trailing dots and whitespaces
         }
         if( this.platform.indexOf( 'linux' ) === 0 ) {
```
27136773, line 804: Should this improvement be done for all operating systems, not just linux?

ronny1982:
Resolves #1385

yobleck:
having trouble fixing
##[error]  918:1  error  Newline not allowed at end of file  eol-last
there is no eol character but lint seems to think there is.

27136773:
> having trouble fixing
> ##[error] 918:1 error Newline not allowed at end of file eol-last
> there is no eol character but lint seems to think there is.

There is probably a blank newline at the end of the file.
Maybe your editor automatically appends one.

yobleck:
I fixed the issue with the eol character but integration fails and appears to be due to files I haven't touched so I'm not sure where to go from here

src/web/mjs/engine/Storage.mjs
```diff
@@ -795,17 +795,19 @@ export default class Storage {
      * LINUX: wxT("/\r\n\t");
      * WINDOWS: wxT("\\/:*?\"<>|\r\n\t");
      */
-    sanatizePath ( path ) {
+    sanitizePath ( path ) {
```
27136773: This is a public method, if renamed all its occurrences must be renamed as well, otherwise the application stops to work correctly.

See:
- BookmarkConnector.mjs
- Manga.mjs
- Chapter.mjs

yobleck:
At this point I'd say I'm in over my head and don't know how to proceed. It's probably better if I just close the pull request or let someone else fix it. Sorry for wasting your time @ronny1982

PR #1406, opened 2020-03-24
Rashid 

arsenbekov:
Hi


src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -262,11 +295,41 @@
...
+                    context.set('enableCheckNewContent',true);
+                },60000,this);
+            }
+
             /**
              *
              */
-            getMangaClass( selectedManga, id ) {
-                return ( !selectedManga || selectedManga.id !== id ? '' : 'focus' );
+            getMangaClass( selectedManga, manga ) {
```
27136773, line 321: May reduce nested conditions to make code more understandable, e.g. classes array:
```javascript
let classes = [];
if(conditionSelected) {
    classes.push('classSelected')
}
if(conditionNewContent) {
    classes.push('classNewContent')
}
return classes.join(' ');
```

src/web/lib/hakuneko/frontend@classic-light/mangas.html
```diff
@@ -262,11 +296,41 @@
...
+            * Checks for new content to read by refreshing the bookmarks, updating the web source and checking the chapters bookmarks
+            */
+            async onCheckNewContent() {
+                if(! this.get('enableCheckNewContent'))
+                {
+                    alert('Please wait 60s before checking again.');
+                    return;
+                }
+                this.set('nbContentUpdating',Engine.BookmarkManager.bookmarks.length); 
+                await Engine.BookmarkManager.checkNewContent();
```
ronny1982: I don't understand the 60 second timer.
The content check is awaitable, so after this is done the flag can just be set to false directly without using the timer?

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -143,6 +147,10 @@
         -->
         <div class="footer">
             <hakuneko-status id="status" message="Mangas: [[ mangaFilteredCount ]] / [[ mangaList.length ]]"></hakuneko-status>
+            &nbsp;<i class$="fas [[ getContentUpdateClass(nbContentUpdating,enableCheckNewContent) ]] button" title="Check new content" on-click="onCheckNewContent"></i>
+            <template is="dom-if" if="[[ isUpdatingContent(nbContentUpdating) ]]">                        
+                   ([[ nbContentUpdating ]])
```
ronny1982: This field seems to be a number, but is handled as bool => make it bool

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -262,11 +295,41 @@
...
+                {
+                    alert('Please wait 60s before checking again.');
+                    return;
+                }
+                this.set('nbContentUpdating',Engine.BookmarkManager.bookmarks.length); 
+                await Engine.BookmarkManager.checkNewContent();
+                this.set('nbContentUpdating',0);
+                this.set('enableCheckNewContent',false);
+                //Disable refresh for 1 minute to prevent hammering
+                setTimeout(function(context){
```
27136773, line 313: I don't understand the 60 second timer.
The content check is awaitable, so after this is done the flag can just be set to true directly without using the timer?

src/web/mjs/engine/BookmarkManager.mjs
```diff
@@ -211,4 +212,22 @@ export default class BookmarkManager extends EventTarget {
...
+            //Retrieve connector
+            let mangaconnector = Engine.Connectors.find( connector => {
+                return connector.id === bookmark.key.connector ;
+            } );
+            //Prepare manga
+            let manga = new Manga(mangaconnector,bookmark.key.manga,bookmark.title.manga);
+            await manga.checkNewContent();
+        };
+        //Check all bookmarks for new content
+        await Promise.all(this.bookmarks.map(checkManga));
```
27136773, line 231: Spawning concurrent updates for all bookmarks at the same time may strain resources and has a high risk to getting IP banned by some websites

src/web/mjs/engine/Manga.mjs
```diff
@@ -210,4 +211,30 @@ export default class Manga extends EventTarget {
         count = Math.min( digits, count );
         return '0'.repeat( digits - count ) + text;
     }
+
+    /**
+     * Check newer content within bookmarks
+     */
+    async checkNewContent()
+    {
+        await this._getUpdatedChaptersFromWebsite();
```
27136773, line 220: Every time a chapter is marked or unmarked, the chapter list is loaded from the website?
Not sure if this is a good idea, especially for websites that are loading the chapters very slow (e.g. 9anime, VRV).
If the chapter list is already cached, this my be used instead...

src/web/mjs/engine/Connector.mjs
```diff
@@ -64,7 +65,11 @@ export default class Connector {
     async initialize() {
         try {
             if(!this.initialized) {
-                await this._initializeConnector();
+                if(!this.initializing)
```
27136773, line 68: Avoid nested conditions if possible.
When initialization fails, the flag is not reset and therefore the connector can never be initialized.

suggestion, something like this
```javascript
if(!this.initialized && !this.initializing) {
    this.initializing = true;
    await this._initializeConnector();
    this.initialized = true;
}
// use a finally for the try block
} finally {
    this.initializing = false;
}
```

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
```diff
@@ -262,11 +295,41 @@
...
             }
 
+            /**
+            * Event listener attached to the refresh button
+            * Checks for new content to read by refreshing the bookmarks, updating the web source and checking the chapters bookmarks
+            */
+            async onCheckNewContent() {
+                if(! this.get('enableCheckNewContent'))
+                {
+                    alert('Please wait 60s before checking again.');
```
27136773, line 305: Clicking on a disabled button should not show any actions

Sheepux:
@ronny1982  Thanks for the review, however i'd reject the merge request as it's still work in progress. The branch was here to share the progress is someone wanted to help/review the current state.
Don't know how someone got to request to merge it

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: During my testing, i've found out that i'm tempted to click on the button to check again for new content just after checking it (yeah that's basically stupid). That would hammer the website when you have quite a long list of bookmarks (i have 200 bookmarks, and they get updated in less than 10s).
Disabling for 1min was the nicest option i've figured out (but it should also be disabled as long as the check content is running). The 1 min wait is not here to wait for the full refresh to end.

src/web/mjs/engine/BookmarkManager.mjs
Sheepux: Yes, but i've planned to use a throttling system within the connector. Will not request for merge as long as it's not done (got much work to do)

src/web/lib/hakuneko/frontend@classic-dark/mangas.html
Sheepux: I should indeed introduce a simple tooltip instead of an alert

Sheepux:
closing because not requested by branch "owner"

Issue #1114, opened 2020-02-14
Find manga from all already indexed/downloaded manga site/connector


kokizzu:
Add one more type other than "bookmarked"/"folder view"/"clipboard", that is "indexed" for all sites that already being indexed.
![image](https://user-images.githubusercontent.com/1061610/74509557-63d4b700-4f34-11ea-9a3e-a07ce9c0f961.png)

27136773:
Please provide more details about what shall be achieved.

Currently it seems the request is to add a new connector that will show all mangas from all websites that has been synced, therefore the connector would show a list of multiple thousands of mangas when selected?

Or shall there be a tag to hide all websites that are not yet synced from the website selection list?

kokizzu:
yes, hundreds of thousands of mangas at once, just like bookmark, but with [suffix] of the connector

kokizzu:
for the tag to hide all websites that are not yet synced will be good solution for https://github.com/manga-download/hakuneko/issues/1113

27136773:
Showing all mangas from all synchronized websites would kill the application with the current UI framework => **rejected**.
You may consider to use the [(2) Manga Search across websites](https://github.com/manga-download/hakuneko/wiki/Quick-Reference#source-selection) instead.

Tag based filtering can be implemented with less effort => **accepted**

hypoflex:
Talked about this type of feature before with a user in #535 
see my suggestion back then here: 
https://github.com/manga-download/hakuneko/issues/535#issuecomment-544678387

Issue #535, opened 2019-10-18
Connector favorite/bookmark feature request


Darkrift411:
There are a TON of connectors (which is great) but it gets kinda hard to remember which ones you have sync'd (and syncing can take a while with some). It would be great if you could either bookmark or favorite connectors you use frequently (have it so they show up at the top, or you have a favorites button you can click that filters out all non favorited connectors).

Another similar idea that might work instead would be to show the last-update status on each connector ("Last updated: 3 days ago" or "Last updated: never") and possibly a way to sort it so the "never" ones are at the end.

hypoflex:
I like the idea, personally i thought it would be best to display all the already downloaded lists as favorite next to the other connectors. what do you think of this example

this would also allow to work with the search option and tags
favorites would be better on the left, and if no favorites are known it wouldnt display
![image](https://user-images.githubusercontent.com/36196198/67238028-b69dd680-f44c-11e9-84d9-648fc07675ae.png)

Darkrift411:
That would work fine. Anything to help you see ones you have already downloaded or to mark favorites so they are easier to find.

Robonau:
it's a bummer that nothing came of this

hypoflex:
> it's a bummer that nothing came of this

due to covid my schedule is all over the place, but i've not forgotten this feature.
with Ronny his feedback ready for me to further implement/change its a matter of finding the moment to uninterrupted continue this.

alienxp03:
Hopefully this is still being considered. I have dory memory so this feature would tremendously help me 😁

MikeZeDev:
Planned in Haruneko, there is even a PR for that.

PR #1456, opened 2020-04-05
Added the ability to see connectors that have been downloaded

hypoflex:
With this addition you're able to see what connectors have already been downloaded,
since there's such great amount of connectors it can be daunting to know which one you used before. this feature is based on the reqeust from #535 and #1114 

this code is far from optimal, best would be to convert `Favorite.mjs`  into `Connectors.mjs`
but to proof it works i've done it this way. made it easier for me to play around without rewriting original code.

(07/04 13:00) Edit based on feedback from Ronny:
- [x] rewrite function to `Storage.mjs` and use already existing methods
- [x] Test on different platforms
- [x] Light theme version

(07/04 13:42) TODO:
- [ ] Resolve the `this.root` variable to local variable using `electron.remote.app.getPath('userData')`
- [ ] Combine `register(file)` and `registerfav(file)` to reduce duplicate code


src/app/ElectronBootstrap.js
```diff
@@ -30,7 +30,8 @@ module.exports = class ElectronBootstrap {
         ];
         this._directoryMap = {
             'cache': this._configuration.applicationCacheDirectory,
-            'plugins': this._configuration.applicationUserPluginsDirectory
+            'plugins': this._configuration.applicationUserPluginsDirectory,
+            'root': this._configuration.applicationUserDataDirectory
```
hypoflex: i was looking for a way to get into the root userData directory to scan for which files were already downloaded.
this seemed like the correct way to do it.
looking for feedback on this

src/web/mjs/engine/Favorites.mjs
```diff
@@ -0,0 +1,66 @@
...
+        try {
+            let response = await fetch(uri);
+            let data = await response.json();
+            let cc = data.filter(plugin => plugin.startsWith('hakuneko.mangas.'));
+            let connectorName = [];
+            cc.forEach(file => {
+                let name = file.substr(file.indexOf('.', 15)+1);
+                connectorName.push("hakuneko://cache/mjs/connectors/" + name + ".mjs");
+            });
+            return connectorName;
```
hypoflex: feel like this could've been done better, specially since it needs to be integrated into the original `Connectors.mjs`
requesting feedback

src/app/ElectronBootstrap.js
hypoflex: Something i forgot while doing this is that it eventually has to work with the `bookmark directory` function. added to the todo list

27136773: No need for this, the `Storage.mjs` has access to the path (`electron.remote.app.getPath( 'userData' )`)

src/web/mjs/engine/Favorites.mjs
27136773: Move this method into the `Storage.mjs` engine and use the appropriate methods available or `this.fs` directly (see `saveMangaList` and `loadMangaList`)
Doing so will save you a lot of code instead of dealing with the protocol interface for accessing files...

27136773: There is also a private method `_readDirectoryEntries(path)` which may come in handy

src/web/lib/hakuneko/frontend@classic-dark/connectors.html
```diff
@@ -279,7 +279,7 @@
                 this.favoriteList = [];
                 // load connectors
                 this.set( 'connectorList', Engine.Connectors );
-                this.set( 'favoriteList', Engine.Favorites );
+                this.set( 'favoriteList', Engine.Storage.downloadedConnectors );
```
hypoflex, line 282: this still needs some work, does not do what i want it yet.

src/web/mjs/engine/Connectors.mjs
```diff
@@ -3,6 +3,7 @@ export default class Connectors {
     constructor(ipc) {
         ipc.listen('on-connector-protocol-handler', this._onConnectorProtocolHandler.bind(this));
         this._list = [];
+        this._favList = [];
```
hypoflex, line 6: since im making a new window with a new list, will need a new list in the code as well.

src/web/mjs/engine/Connectors.mjs
```diff
@@ -24,16 +25,22 @@ export default class Connectors {
         ];
         let userPlugins = await this._loadPlugins('hakuneko://plugins/');
         let internalPlugins = await this._loadPlugins('hakuneko://cache/mjs/connectors/');
+        let favoritePlugins = Storage.downloadedConnectors;
 
         await this.register(systemPlugins);
         await this.register(userPlugins);
         await this.register(internalPlugins);
+        await this.register(favoritePlugins);
```
hypoflex, line 33: this failed. needs changing

src/web/mjs/engine/Storage.mjs
```diff
@@ -30,6 +30,7 @@ export default class Storage {
         this.fs = require( 'fs' );
         this.path = require( 'path' );
         this.config = this.path.join( electron.remote.app.getPath( 'userData' ), 'hakuneko.' );
+        this.root = electron.remote.app.getPath( 'userData');
```
hypoflex, line 33: couldn't use `this.config` since it contains hakuneko in the path. i don't like rewriting original paths or duplicating variable values, this needs to change

src/web/mjs/engine/Storage.mjs
```diff
@@ -911,4 +912,20 @@ export default class Storage {
...
+        //seems like performance drain
+        let rootDirectoryEntries = this._readDirectoryEntries(this.root);
+        let arrayFiltered = [];
+        rootDirectoryEntries.then(function(arrayRaw){
+            arrayRaw = arrayRaw.filter(plugin => plugin.startsWith('hakuneko.mangas.'));
+            arrayRaw.forEach(file => {
+                arrayFiltered.push(file.substr(file.indexOf('.', 15)+1));
+            });
+            return arrayFiltered;
+        })
```
hypoflex, line 927: `electron.remote.app` was not working so decided on the new variable `this.root` here.
retrieved the directory with `_readDirectoryEntries` as Ronny surgested

src/web/mjs/engine/Connectors.mjs
```diff
@@ -63,7 +63,30 @@ export default class Connectors {
...
+                }
+            }
+            this._favlist.sort( ( a, b ) => {
+                return ( a.label.toLowerCase() < b.label.toLowerCase() ? -1 : 1 );
+            } );
+        } catch(error) {
+            console.warn(`Failed to load connector`, error);
+        }
+    }
+
```
hypoflex, line 89: only part I'm not happy with, since original `register(files)` uses `this._list` (which can't be rewritten) I've had to create a new register function.

hypoflex:
(07/04 13:42) TODO:
- Resolve the `this.root` variable to local variable using `electron.remote.app.getPath('userData')`
- Combine `register(file)` and `registerfav(file)` to reduce duplicate code

src/web/lib/hakuneko/frontend@classic-dark/connectors.html
```diff
@@ -212,6 +212,34 @@
                         </template>
                     </div>
                 </div>
+                <div class="connectors">
```
27136773, line 215: All this duplicated code should not be added.
Just introduce a new checkbox or something to toggle between the full list of websites and the websites that have been downloaded. In the code behind, assign the corresponding list to the property which is bound to the dom-repeat template in the UI.

```javascript
onClick(evt) {
    this.connectorList  = evt.value ? /* downloaded website list */ this.favoriteList : /* full website list */;
}
```

src/web/lib/hakuneko/frontend@classic-light/connectors.html
```diff
@@ -212,6 +212,34 @@
                         </template>
                     </div>
                 </div>
+                <div class="connectors">
```
27136773, line 215: Same as for the dark theme

src/web/mjs/engine/Connectors.mjs
```diff
@@ -24,16 +25,22 @@ export default class Connectors {
         ];
         let userPlugins = await this._loadPlugins('hakuneko://plugins/');
         let internalPlugins = await this._loadPlugins('hakuneko://cache/mjs/connectors/');
+        let favoritePlugins = Engine.Storage.downloadedConnectors;
 
         await this.register(systemPlugins);
         await this.register(userPlugins);
         await this.register(internalPlugins);
+        await this.registerfav(favoritePlugins);
```
ronny1982, line 33: Is this required?
All websites should be registered by `await this.register(userPlugins);` and `await this.register(internalPlugins);`. If there is a website in favorites, that is not in `internalPlugins` or `userPlugins` that software is doing something wrong.

src/web/mjs/engine/Storage.mjs
```diff
@@ -911,4 +912,18 @@ export default class Storage {
             } );
         } );
     }
+
+    get downloadedConnectors() {
```
27136773, line 916: Just return a list of connector IDs (string), makes further processing much simpler.

src/web/mjs/engine/Connectors.mjs
```diff
@@ -24,16 +25,22 @@ export default class Connectors {
...
     }
 
     get list() {
         return this._list;
     }
 
+    get favlist() {
+        return this._favlist;
```
27136773, line 41: Just get the connector IDs of downloaded lists as suggested in `Storage.mjs` and then filter the full list based of the downloaded...
e.g.
```javascript
let downloaded = Storage.getDownloadedConnectorIDs();
return this._list.filter(connector => downloaded.includes(connector.id));
```

src/web/mjs/engine/Connectors.mjs
```diff
@@ -24,16 +25,22 @@ export default class Connectors {
         ];
         let userPlugins = await this._loadPlugins('hakuneko://plugins/');
         let internalPlugins = await this._loadPlugins('hakuneko://cache/mjs/connectors/');
+        let favoritePlugins = Engine.Storage.downloadedConnectors;
```
27136773, line 28: Not needed anymore if you accept the suggested change based on `Storage.mjs`

src/web/mjs/engine/Connectors.mjs
27136773, line 33: Not needed anymore if you accept the suggested change based on `Storage.mjs`

src/web/mjs/engine/Connectors.mjs
27136773: Method is not needed anymore if you accept the suggested change based on `Storage.mjs`

src/web/mjs/engine/Connectors.mjs
```diff
@@ -3,6 +3,7 @@ export default class Connectors {
     constructor(ipc) {
         ipc.listen('on-connector-protocol-handler', this._onConnectorProtocolHandler.bind(this));
         this._list = [];
+        this._favlist = [];
```
27136773, line 6: Not needed anymore if you accept the suggested change based on `Storage.mjs`

27136773 (commented):
Looks good, but i would strongly recommend to take a look at the suggestions especially in the UI and Storage as these will greatly simplify the implementation and reuse existing concepts.

hypoflex:
> Looks good, but i would strongly recommend to take a look at the suggestions especially in the UI and Storage as these will greatly simplify the implementation and reuse existing concepts.

again thanks for your feedback, going to pick this up asap

27136773:
Is this PR still active?

27136773:
Will close for now, maybe re-opened when work is continued

Issue #955, opened 2020-01-19
Pre-Flight Check for 6.1.7 Releases


27136773:
### Pre-Flight Check

Validate that the current HakuNeko release builds for 6.1.7 are running on the following systems by performing the appended (manual) test suite:

----
#### Test Environment (Production Systems)

![](https://img.shields.io/badge/-Windows%2010.1903%20(64--Bit)-blue?logo=windows)
![](https://img.shields.io/badge/-macOS%2010.15.2%20(64--Bit)-silver?logo=apple&logoColor=whitesmoke)
![](https://img.shields.io/badge/-Ubuntu%2018.04.3%20(64--Bit)-orange?logo=ubuntu)
![](https://img.shields.io/badge/-Ubuntu%2019.10%20(64--Bit)-orange?logo=ubuntu)
![](https://img.shields.io/badge/-Linux%20Mint%2019.3%20Cinnamon%20(64--Bit)-green?logo=linux-mint)
![](https://img.shields.io/badge/-Debian%2010.2.0%20(64--Bit)-red?logo=debian)
![](https://img.shields.io/badge/-Manjaro%2018.1.5%20KDE%20(64--Bit)-grey?logo=manjaro)
![](https://img.shields.io/badge/-MX%20Linux%2019%20(64--Bit)-teal?logo=mx-linux)
![](https://img.shields.io/badge/-deepin%2015.11%20(64--Bit)-blueviolet?logo=linux)
![](https://img.shields.io/badge/-elementary%20OS%205.1%20(64--Bit)-lightgrey?logo=elementary)
![](https://img.shields.io/badge/-Fedora%2031--1.9%20(64--Bit)-blue?logo=fedora)
![](https://img.shields.io/badge/-Solus%204.0%20Budgie%20(64--Bit)-midnightblue?logo=solus)
![](https://img.shields.io/badge/-centOS%208.1.1911%20(64--Bit)-mediumaquamarine?logo=centos)
![](https://img.shields.io/badge/-openSUSE%20Leap%2015.1%20(64--Bit)-limegreen?logo=opensuse)

----
#### Prerequisites

- On macOS install `ImageMagick`
- On linux install `ffmpeg` and `imagemagick`

----
#### Test Suite

<details><summary>Show Instructions</summary>

1. Download and install/extract the corresponding HakuNeko package for the corresponding testing platform
2. Start HakuNeko via command line (observe output for possible errors)
3. Press F12 to open the developer console (observe output for possible errors)
[...]

PR #1457, opened 2020-04-05
Main Window Frameless

Sheepux:
Changes the main window to frameless.
Top left menu is now draggable and includes minimize,maximize and close buttons.
Connectors's browser window is still with frame enabled.

Todo :
- [X] Check pre-flight-check steps that can be related to this change and detail the following tests
- [x] Test on Windows
- - [X] The default menubar should appear during loading then disapears after theme load
- - [X] Button Minimize sends hakunko to the minimized state
- - [X] Button Maximize sends hakunko to the full size of the screen (but not in "full screen" mode)
- - [X] Clicking on Maximize again restores the window to it's original size
- - [X] The windows is only draggable on the top left part (#Menu), move it around while in non maximized state.
- - [X] Double clicking on the draggable part is triggering the Maximize function (do the same tests as before)
- - [X] The configuration menu (on top left) is still clickable.
- - [X] The unmaximized window can be resized in 3 corners (top right, bottom right, bottom left). The resize icon appears when on the 4 edge (top bottom, left, right)
- - [X] Open a connector's website from the connector pannel and asses that the window topbar is here as well as system controls (min, max, close buttons)
- - [X] Button Close closes hakunko (and no electron process is still runing afterwards)
- - [X] Run pre-flight steps 40 & 41
- [x] Test on Linux
- - [x] The default menubar should appear during loading then disapears after theme load
- - [x] Button Minimize sends hakunko to the minimized state
- - [x] Button Maximize sends hakunko to the full size of the screen (but not in "full screen" mode)
- - [x] Clicking on Maximize again restores the window to it's original size
- - [x] The windows is only draggable on the top left part (#Menu), move it around while in non maximized state.
- - [x] Double clicking on the draggable part is triggering the Maximize function (do the same tests as before)
- - [x] The configuration menu (on top left) is still clickable.
- - [x] The unmaximized window can be resized in 3 corners (top right, bottom right, bottom left). The resize icon appears when on the 4 edge (top bottom, left, right)
- - [x] Open a connector's website from the connector pannel and asses that the window topbar is here as well as system controls (min, max, close buttons)
- - [x] Button Close closes hakunko (and no electron process is still runing afterwards)
- - [x] Run pre-flight steps 40 & 41
- [x] Test on MacOS
- ⚠️ The default menubar should appear during loading then disapears after theme load
Menu bar appears, but there are 3 squares instead of min/max/close symbols
- - [x] Button Minimize sends hakunko to the minimized state
- - [x] Button Maximize sends hakunko to the full size of the screen (but not in "full screen" mode)
- - [x] Clicking on Maximize again restores the window to it's original size
- - [x] The windows is only draggable on the top left part (#Menu), move it around while in non maximized state.
- ❌ Double clicking on the draggable part is triggering the Maximize function (do the same tests as before)
- - [x] The configuration menu (on top left) is still clickable.
- - [x] The unmaximized window can be resized in 3 corners (top right, bottom right, bottom left). The resize icon appears when on the 4 edge (top bottom, left, right)
- - [x] Open a connector's website from the connector pannel and asses that the window topbar is here as well as system controls (min, max, close buttons)
- - [x] Button Close closes hakunko (and no electron process is still runing afterwards)
- - [x] Run pre-flight steps 40 & 41

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/menu.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/menu.html
@@ -130,9 +130,9 @@
         <i class="fas fa-fw fa-bars fa-2x item button" title="Toggle menu" on-click="togglePopup"></i>
         <span class="title">HakuNeko</span>
         <span class="right">
-        <i class="far fa-fw fa-window-minimize item button" title="Minize HakuNeko" on-click="minimizeWindow"></i>
-        <i class="far fa-fw fa-window-maximize item button" title="Maximize HakuNeko" on-click="maximizeWindow"></i>
-        <i class="far fa-fw fa-2x fa-window-close item button" title="Close HakuNeko" on-click="closeWindow"></i>
+        <i class="far fa-fw fa-window-minimize item button" title="Minize HakuNeko" onclick="minimizeWindow()"></i>
+        <i class="far fa-fw fa-window-maximize item button" title="Maximize HakuNeko" onclick="maximizeWindow()"></i>
+        <i class="far fa-fw fa-2x fa-window-close item button" title="Close HakuNeko" onclick="closeWindow()"></i>
         </span>
         <div class$="popup [[ getPopupClass(popupVisibility) ]]">
             <div class="separator">
@@ -237,7 +237,6 @@
                 this.version = Engine.Version;
                 this.popupVisibility = false;
                 this.settingCategories = Engine.Settings.getCategorizedSettings();
-                this.win=require('electron').remote.getCurrentWindow();
             }
 
             getPopupClass(visibility) {
@@ -311,21 +310,6 @@
                 }
             }
 
-            minimizeWindow(event){
-                this.win.minimize();
-            }
-
-            maximizeWindow(event){
-                if (!this.win.isMaximized()) {
-                    this.win.maximize();
-                } else {
-                    this.win.unmaximize();
-                }
-            }
-
-            closeWindow (event){
-                this.win.close();
-            }
         }
         window.customElements.define(HakunekoMenu.is, HakunekoMenu);
     </script>
--- a/src/web/lib/hakuneko/frontend@classic-dark/menu.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/menu.html
@@ -106,6 +106,10 @@
             }
             .button {
                 cursor: pointer;
+                -webkit-app-region: no-drag;
+            }
+            .right {
+                float: right;
             }
             .buttons {
                 display: flex;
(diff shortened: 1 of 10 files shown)


src/web/lib/hakuneko/frontend@classic-dark/menu.html
```diff
@@ -300,6 +310,22 @@
...
+            maximizeWindow(event){
+                if (!this.win.isMaximized()) {
+                    this.win.maximize();
+                } else {
+                    this.win.unmaximize();
+                }
+            }
+
+            closeWindow (event){
+                this.win.close();
```
27136773: Does the confirmation dialog still popup if downloads are running?
See: [Pre-Flight Check](https://github.com/manga-download/hakuneko/issues/955) Step 40

src/web/lib/hakuneko/frontend@classic-light/menu.html
```diff
@@ -300,6 +310,22 @@
...
+            maximizeWindow(event){
+                if (!this.win.isMaximized()) {
+                    this.win.maximize();
+                } else {
+                    this.win.unmaximize();
+                }
+            }
+
+            closeWindow (event){
+                this.win.close();
```
27136773: Same as for dark theme

27136773 (commented):
Looks good.
Needs to be tested on macOS and Linux (Ubuntu, debian w. GNOME)

The pro of this may also be the con:
Each front-end can have a customized title bar, but each front-end must always implement a title bar.

How about evaluating a solution (if this is even possible), where the title bar is part of the `index.html` file and a front-end may or may not replace it?

Sheepux:
> How about evaluating a solution (if this is even possible), where the title bar is part of the `index.html` file and a front-end may or may not replace it?

I see what you mean, i'll take a look at it

Sheepux:
Added default menubar in index.html that can be hidden within the theme css ( it works!:)
Tests coverage updated.
Ran successfully on windows.

src/web/lib/hakuneko/frontend@classic-dark/menu.html
Sheepux: Yup ! (added in test coverage)

src/web/lib/hakuneko/frontend@classic-light/menu.html
Sheepux: Yup ! (added in test coverage)

Sheepux:
Regarding the icons during loading time maybe macOS dooesn't recognise those UTF8 chars:
🗕🗖🗙
https://www.fileformat.info/info/unicode/char/1f5d5/index.htm
https://www.fileformat.info/info/unicode/char/1f5d6/index.htm
https://www.fileformat.info/info/unicode/char/1f5d6/index.htm

I should probably prefix it with pure text :
Min🗕 Max🗖 Close🗙
Any idea? :)
Edit: images ?

Sheepux:
Regarding the double click... could it be a windows's behavior ? Do we consider this as an anomaly (don't know much about MacOS) ?

27136773:
I couldn't find a font that shows these unicode characters, so i suggested either to use embedded gifs as base64 data URLs or using SVG sprites, which seems to be the best practice right now (maybe also switching fontawesome from font system to SVG sprites ...)

macOS usually maximizes the window as well when double clicking the title bar. I have no clue why this is not working and if it is essential to be working on macOS or not.

Sheepux:
Regarding the double click maximize I've just found this :
https://github.com/electron/electron/issues/16385
I'm unsure if we should add the extra code just for MacOS.

27136773:
Maybe its possible to just capture the javascript double click event on the element and call the same methods (toggle) as when clicking the maximize/minimize buttons

27136773 (approved):
Looks good

PR #1542, opened 2020-04-18
[ReadMangaToday] Remove Ads images - Fixes #1541

### LINKED ISSUE #1541: [ReadMangaToday] Picture of Ads displayed instead of manga pictures
Problem : see title

https://i.imgur.com/pDpXlmB.png

How to reproduce : Try to view any chapter from any manga on ReadMangaToday.

Of course i am using latest version atm.

Apparently is not "replacing manga pages" but just an added ad, meaning the appropriate solution would be some kind of filtering

Sheepux:
Fixes #1541 by filtering images that have the class "img-responsive" (ads don't)

Checked a few random mangas and the images have the same count as on the site and display correctly.

Checked chapter download, works fine.

### DIFF
--- a/src/web/mjs/connectors/ReadMNG.mjs
+++ b/src/web/mjs/connectors/ReadMNG.mjs
@@ -71,21 +71,13 @@ export default class ReadMNG extends Connector {
             } );
     }
 
-    /*
-    * Used to check if the img is not an ad
-    */
-    checkMangaImage(img) {
-        return img.classList.contains('img-responsive');
-    }
-
     /**
      *
      */
     _getPageList( manga, chapter, callback ) {
         let request = new Request( this.url + chapter.id + '/all-pages', this.requestOptions );
-        this.fetchDOM( request, 'div.page_chapter source' )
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/ReadMNG.mjs
```diff
@@ -71,13 +71,21 @@ export default class ReadMNG extends Connector {
...
+    checkMangaImage(img) {
+        return img.classList.contains('img-responsive');
+    }
+
     /**
      *
      */
     _getPageList( manga, chapter, callback ) {
         let request = new Request( this.url + chapter.id + '/all-pages', this.requestOptions );
         this.fetchDOM( request, 'div.page_chapter source' )
```
27136773: Could this filtering be done in the CSS query, with `:not(..)` or match operator?
something like `div.page_chapter source[class*="img-responsive"]`

Sheepux: Indeed, your filter worked nicely. I have not done such filtering for years, thanks for the help.

Tested again (preview/download): ok

27136773 (approved):
Looks good

PR #1545, opened 2020-04-19
Updated keyboard shortcuts in reader

guydht:
- Changed key recognition to be by code instead of by key
- Swiitched page up/down and arrow up/down
- Made the arrow up/down (previously page up/down) event be smooth

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/pages.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/pages.html
@@ -202,18 +202,10 @@
             onKeyDown( event ) {
                 switch( true ) {
                     case event.code === 'ArrowUp' && !event.ctrlKey:
-                        this.$.container.scrollBy({ 
-                            top: -64,
-                            left: 0, 
-                            behavior: 'smooth' 
-                        });
+                        scrollSmoothly(this.$.container, -64);
                         break;
                     case event.code === 'ArrowDown' && !event.ctrlKey:
-                        this.$.container.scrollBy({ 
-                            top: 64,
-                            left: 0, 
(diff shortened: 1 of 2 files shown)


27136773:
I think changes should be applied to both, light theme and dark theme.
Is the key code reliable for different keyboard locales?
e.g. US keyboard uses the same key for `+` and `=`, so the key code for `=` will work, but on other locales the `=` sign may not be on the same key as the `+`

Sheepux (changes requested):
You should not use smooth scrolling on "simple move down" as it locks repeating scrolling (everytime you press the key the velocity restarts and during my testing it felt weird/wrong).
You should let it to the default behavior even though it's not smooth :  this.$.container.scrollTop -= 64;

Sheepux:
Regarding the swap of up/down & page/up i'm ok to get it back to what we expect from a default browser behavior. 
I'll have to work on being able to change keys settings in the config menu (we expected a key behavior war).

Sheepux (changes requested):
As Ronny mentioned, changes should also be done in the light theme (currently 2 themes to maintain, that will maybe change "soon")

guydht:
I'm not 100% happy with the "scrollsmoothly" function but it gets the job done (doesn't feel weird when spamming the arrow keys).
I changed the minus/plus handling because I really don't know if there are locales that behave differently and better safe than sorry.
Aaaaand forgot that light theme exists :/

Sheepux:
The "tiny scroll" function seems to be working nicely (i've tested your branch).
I'm glad someone is also working on improving the reader 👍

Sheepux (approved):
Looks Good To Me

PR #1592, opened 2020-04-27
[niconico seiga] add support

### LINKED ISSUE #25: Add Nico Nico Shizuka
https://seiga.nicovideo.jp/manga/

ghost:
-manga list is 27000+ series long; currently does not send list requests concurrently and requires ~2000 API requests just to fetch it all. with this website, this takes approximately 20 minutes.

Resolves #25


src/web/mjs/connectors/NicoNicoSeiga.mjs
```diff
@@ -0,0 +1,75 @@
...
+        return data.map(series => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink('/comic/' + series.id, request.url),
+                title: series.title
+            };
+        });
+    }
+
+    async _getMangas() {
+        let request = new Request(this.url + this.mangaListPage, this.requestOptions);
```
27136773: May use URL object instead of UVString
`new Request(new URL(this.mangaListPage, this.url), this.requestOptions)`
For all other `Requests` in this file as well

27136773 (approved):
Looks good

27136773:
There is nothing wrong with making 20k requests as there is no way around it. Other websites do this as well.

PR #1594, opened 2020-04-27
Add Comic Polaris

chocolatier:
Adds support for comic polaris. Their system is identical to Comic Meteor, so ComicPolaris.mjs is basically ComicMeteor.mjs with a few strings changed.


27136773 (approved):
Looks good

PR #1602, opened 2020-04-29
[mangajar] add website support

### LINKED ISSUE #1068: please add https://mangajar.com/
add the site please

spyoff:
Changelogs:
- add mangajar connector (https://mangajar.com)
- add mangajar icon (from favicon)
- close #1068


src/web/mjs/connectors/MangaJar.mjs
```diff
@@ -0,0 +1,65 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class MangaJar extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'mangajar';
+        super.label = 'MangaJar';
+        this.tags = [ 'manga', 'english', 'scanlation' ];
```
27136773: It seems the website also provides `webtoon` which may be added as tag

src/web/mjs/connectors/MangaJar.mjs
```diff
@@ -0,0 +1,65 @@
...
+        });
+    }
+
+    async _getChapters( manga ) {
+        let request = new Request(new URL(`/manga/${manga.id}/chaptersList`, this.url));
+        let data = await this.fetchDOM(request, 'li.chapter-item');
+
+        return data.map(element => {
+            return {
+                manga: manga,
```
27136773: The internal chapter constructor will ignore the property `manga`, so it is redundant and can be removed

27136773 (approved):
Looks good (optional minor improvements)

src/web/mjs/connectors/MangaJar.mjs
spyoff: added webtoon tag

src/web/mjs/connectors/MangaJar.mjs
spyoff: remove the manga object, previously I think that I need to keep it, but chapter-id actually stores the url `path` rather than `chapter-id` 🙈

Issue #1558, opened 2020-04-20
Site Request: readm.org


Ognisty321:
Hello, can you add this website? https://readm.org

Thanks!

27136773:
See #1344

PR #1636, opened 2020-05-02
[readm] Add website support + fix MangaGo test

### LINKED ISSUE #1344: Add website readm
https://readm.org

XenonR:
- Add website support for readm (#1558 #1344)
- fixed test for MangaGo (changed protocol to https)

----
Resolves #1344


src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+            let url = this.url + '/manga-list/' + letter;
+
+            let request = new Request(url, this.requestOptions);
+            let data = await this.fetchDOM(request, '.poster > a');
+
+            for (let manga of data) {
+                mangas.push(
+                    {
+                        id: this.url+manga.pathname,
+                        title: manga.querySelectorAll('h2')[0].innerText.trim()
```
27136773: If it is always the first element consider using `querySelector` instead of `querySelectorAll`

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+        for(let letter of index) {
+            let url = this.url + '/manga-list/' + letter;
+
+            let request = new Request(url, this.requestOptions);
+            let data = await this.fetchDOM(request, '.poster > a');
+
+            for (let manga of data) {
+                mangas.push(
+                    {
+                        id: this.url+manga.pathname,
```
27136773: I think it is more appropriate to just save the path as ID without the origin.
Also good find of property `pathname` didn't know this exist...

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+            }
+        }
+
+        return mangas;
+    }
+
+    async _getChapters(manga) {
+        let chapters = [];
+
+        let request = new Request(manga.id, this.requestOptions);
```
27136773: May use `new URL(manga.id, this.url)` in case the manga.id is a relative URL

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+        let data = await this.fetchDOM(request, '.episodes-list .item');
+
+        for (let chapter of data) {
+            chapters.push(
+                {
+                    id: this.url+chapter.querySelectorAll('#table-episodes-title > h6 > a')[0].pathname,
+                    title: chapter.querySelectorAll('#table-episodes-title')[0].innerText.trim(),
```
27136773: If it is always the first element consider using `querySelector` instead of `querySelectorAll`

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+    async _getChapters(manga) {
+        let chapters = [];
+
+        let request = new Request(manga.id, this.requestOptions);
+        let data = await this.fetchDOM(request, '.episodes-list .item');
+
+        for (let chapter of data) {
+            chapters.push(
+                {
+                    id: this.url+chapter.querySelectorAll('#table-episodes-title > h6 > a')[0].pathname,
```
27136773: I think it is more appropriate to just save the path as ID without the origin.
If it is always the first element consider using `querySelector` instead of `querySelectorAll`

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+                }
+            );
+        }
+
+        return chapters;
+    }
+
+    async _getPages(chapter) {
+        let images = [];
+        let request = new Request(chapter.id, this.requestOptions);
```
27136773: May use `new URL(manga.id, this.url)` in case the chapter.id is a relative URL

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+        let data = await this.fetchDOM(request, '.ch-image-container');
+
+        for ( let image of data[0].querySelectorAll('source') ) {
+            images.push(this.url+image.attributes.getNamedItem('src').value);
```
27136773: May use `this.getRootRelativeOrAbsoluteLink(image, request.url)` to create the link

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        return images;
+    }
+
+    async _getMangaFromURI(uri) {
+        let href;
+
+        if ( uri.href.endsWith('/all-pages') ) {
+            href = uri.href.split('/');
```
27136773: May use `uri.pathname` instead of a local variable.
May be optimized to `uri.pathname = uri.pathname.split('/').slice(0, -2).join('/')`

src/web/mjs/connectors/ReadM.mjs
```diff
@@ -0,0 +1,86 @@
...
+            href.pop();
+            href.pop();
+            href = href.join('/');
+        } else {
+            href = uri.href;
+        }
+
+        let request = new Request(href, this.requestOptions);
```
27136773: My use `uri` object directly instead of `href` when opting in for the modification approach posted above

src/__tests__/Connectors.e2e.js
```diff
@@ -174,7 +174,7 @@ describe("HakuNeko Engine", () => {
                     chapterID: '/read-manga/black_clover/bt/314637/Ch1/',
                     chapterTitle: 'Ch.1  : The Boy\'s Vow',
                     pageCount: 51,
-                    pageMatcher: /^http:\/\/iweb\d.mangapicgallery.com\/r\/newpiclink\/black_clover\/1\/[a-z0-9]{32}\.(?:png|jpg|jpeg)$/
+                    pageMatcher: /^(http|https):\/\/iweb\d.mangapicgallery.com\/r\/newpiclink\/black_clover\/1\/[a-z0-9]{32}\.(?:png|jpg|jpeg)$/
```
27136773: May be simplified to `^https?:`

27136773 (approved):
Looks good

src/web/mjs/connectors/ReadM.mjs
XenonR: Is there a benefit in using `new URL(...)` over `this.url+manga.id`?

src/web/mjs/connectors/ReadM.mjs
XenonR: Isn't `_getPages` supposed to return a list of absolute url's? `getRootRelativeOrAbsoluteLink(...)` always returns me an relative version.

src/web/mjs/connectors/ReadM.mjs
27136773: From users view: no
From developers view: The URL object is preferred over USVStrings in case of robustness and further processing (e.g. changing paramters)

src/web/mjs/connectors/ReadM.mjs
27136773: Yes correct, in this case may use `this.getAbsolutePath(image, request.url)`

src/web/mjs/connectors/ReadM.mjs
XenonR: ok, fair point. will incorporate `new URL()`

src/web/mjs/connectors/ReadM.mjs
XenonR: That did the trick.

PR #1640, opened 2020-05-02
[onepunchman] account for moving subdomain

XenonR:
The subdomain is constantly changing.
- id's always absolute
- removed paste URI support


src/web/mjs/connectors/OnePunchMan.mjs
```diff
@@ -23,22 +22,20 @@ export default class OnePunchMan extends Connector {
     async _getChapters(manga) {
         let request = new Request(manga.id, this.requestOptions);
         let data = await this.fetchDOM(request, '#Chapters_List a');
+
         return data.map(element => {
             return {
-                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
```
27136773, line 28: I would keep it unchanged, it will return a relative path if the the link is on the same origin(this.url), or the absolute path if the link is pointing to a different origin.

src/web/mjs/connectors/OnePunchMan.mjs
```diff
@@ -23,22 +22,20 @@ export default class OnePunchMan extends Connector {
...
+                id: this.getAbsolutePath(element, this.url),
                 title: element.text.split(', ')[1],
                 language: 'en'
             };
         });
     }
 
     async _getPages(chapter) {
-        let request = new Request(this.url + chapter.id, this.requestOptions);
```
27136773, line 36: May use `new URL(chapter.id, this.url)` instead to handle both cases:
1. The chapter id is a relative path
2. The chapter id is an absolute URL

27136773 (commented):
Looks good.
Minor optimizations possible.

src/web/mjs/connectors/OnePunchMan.mjs
XenonR: I see your point, but I'll actually never expect a same hostname. Also for simplicity I prefer it that way (eg. use only one function instead of two `getAbsolutePath` vs `getRootRelativeOrAbsoluteLink`)

src/web/mjs/connectors/OnePunchMan.mjs
XenonR: chapter id will always be absolute

27136773 (approved):
Looks good

PR #1645, opened 2020-05-03
FIX- Frameless issue with menu buttons

Sheepux:
The buttons (github doc, issues etc...)in the menu opens a browser page with an url, but then the close buttons aren't there because of the frameless mode.
Adding back the default frame menu in such cases.


27136773 (approved):
Looks good

PR #1657, opened 2020-05-05
[anyacg] fixed manga with missing language code

27136773:


src/web/mjs/connectors/templates/AnyACG.mjs
```diff
@@ -24,7 +24,8 @@ export default class AnyACG extends Connector {
 
     _getLanguage(element, queryFlag) {
         let language = queryFlag ? element.querySelector(queryFlag) : undefined;
-        return language ? ' (' + language.className.match(/flag_([_a-z]+)/)[1].replace(/_/g, '-') + ')' : '';
+        let code = language ? language.className.match(/flag_([_a-z]+)/) : undefined;
+        return code && code.length > 1 ? ' (' + code[1].replace(/_/g, '-') + ')' : '';
```
XenonR, line 28: Are we sure the language code will always be lower case?
And `code.length`can only be <1 if the class name is just `flag_` or ends in capital letters or numbers.
If both cases are intended it looks good to me.

XenonR (approved):
Ok to go.

src/web/mjs/connectors/templates/AnyACG.mjs
27136773: Unfortunately the [history](https://github.com/manga-download/hakuneko/commits/master/src/web/mjs/connectors/templates/AnyACG.mjs) is lost of AnyACG after re-structuring the project...

If my memory is correct i fixed the regex in the past because there were some edge cases on some websites regarding the `flag_*` classes.
For now i would suggest to leave it as-is and re-evaluate the behavior if a related bug occurs...

PR #1662, opened 2020-05-05
[hipercool] add website support

### LINKED ISSUE #1095: Add site HipercooL plis (pt-br)
If possible, please add the website:
https://hiper.cool/
because there is still no hentai content in portuguese ..

XenonR:
Closes #1095


src/web/mjs/connectors/HiperCool.mjs
```diff
@@ -0,0 +1,73 @@
...
+import Manga from '../engine/Manga.mjs';
+
+export default class HiperCool extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'hipercool';
+        super.label = 'Hiper Cool';
+        this.tags = [ 'manga', 'portugese', 'hentai' ];
+        this.url = 'https://hiper.cool/';
```
27136773: May remove trailing slash

src/web/mjs/connectors/HiperCool.mjs
```diff
@@ -0,0 +1,73 @@
...
+
+        while ( morePages ) {
+            try {
+                request = new Request(new URL('/home/' + page, this.url), this.requestOptions);
+                data = await this.fetchDOM(request, '.news-list > li');
+                page++;
+
+                for (let manga of data) {
+                    if ( !uniq.has(manga.querySelector('a.news-thumb').pathname) ) {
+                        uniq.add(manga.querySelector('a.news-thumb').pathname);
```
27136773: I think duplicates will already be removed in the base class, so this would be redundant

src/web/mjs/connectors/HiperCool.mjs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'ul.chapter-list li a.title');
+
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: "Chapter "+element.pathname.split('/').slice(-1)[0],
```
27136773: May be simplified to `.pop()` instead of `.slice(-1)[0]`

src/web/mjs/connectors/HiperCool.mjs
```diff
@@ -0,0 +1,73 @@
...
+                language: 'pt'
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'div.pages source');
+        return data.map(element => this.getRootRelativeOrAbsoluteLink(element, request.url));
```
27136773: Page link should be absolute path instead of relative path

src/web/mjs/connectors/HiperCool.mjs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    async _getMangaFromURI(uri) {
+        let request = new Request(new URL(uri.href), this.requestOptions);
+        let data = await this.fetchDOM(request, '.title-bar span');
+        return new Manga(this, this.getRootRelativeOrAbsoluteLink(uri, this.url), data[0].textContent);
```
27136773: In this method it is safe to assume that `uri.origin` is the same as `this.url` so `uri.pathname` can be used directly without applying the transformation function.

27136773 (commented):
The absolute path in `_getPages` should be addressed before merging

src/web/mjs/connectors/HiperCool.mjs
XenonR: Oh indeed. was thinking too complicated here.

src/web/mjs/connectors/HiperCool.mjs
XenonR: Looks you're right. Didn't know that.

src/web/mjs/connectors/HiperCool.mjs
27136773: It looks page links may still return the relative path, did you push all changes to the PR?

XenonR: right. don't know what happened. fixed now.

27136773 (approved):
Looks good

RenanMenck:
The language was placed as 'portugese' instead of 'portuguese'

XenonR:
> 
> 
> The language was placed as 'portugese' instead of 'portuguese'

Good catch. Put on my todo list.

PR #1665, opened 2020-05-06
[KissLove] add website support (#1658)

### LINKED ISSUE #1658: [Site Request] kisslove (Raw)
Site; KissLove
SiteUrl; https://kisslove.net
SiteMangaList; https://kisslove.net/manga-list.html
Site; RAW MANGA
Site language; Japannese

Their themes look like; loveheaven.net

dodieboy:


src/web/mjs/connectors/KissLove.mjs
```diff
@@ -0,0 +1,13 @@
+import FlatManga from './templates/FlatManga.mjs';
+
+export default class LHScan extends FlatManga {
```
27136773: Change classname from `LHScan` to `KissLove`, i think there is already a class with that name.

src/web/mjs/connectors/KissLove.mjs
```diff
@@ -0,0 +1,13 @@
...
+
+    constructor() {
+        super();
+        super.id = 'kisslove';
+        super.label = 'KissLove';
+        this.tags = [ 'manga', 'raw', 'japanese' ];
+        this.url = 'https://kisslove.net/';
```
27136773: May remove trailing slash

27136773 (commented):
Looks good, but classname collision needs to be addressed before merge

XenonR:
Would close #1658

src/web/mjs/connectors/KissLove.mjs
dodieboy: omg why i keep forgetting to remove slash. fixed that alr, sry

27136773 (approved):
Looks good

PR #1682, opened 2020-05-07
[skscans] fixed website support

27136773:
References fmd-project-team/FMD@476f1eaf


src/web/mjs/connectors/SKScans.mjs
```diff
@@ -0,0 +1,12 @@
+import Genkan from './templates/Genkan.mjs';
+
+export default class SKScans extends Genkan {
+
+    constructor() {
+        super();
+        super.id = 'skscans';
+        super.label = 'SK Scans (Sleeping Knight)';
+        this.tags = [ 'webtoon', 'high-quality', 'english', 'scanlation' ];
```
XenonR, line 9: I find the tag `webtoon` and `scanlation` contradicting, also there seem to be no webtoons available on this source.

XenonR (commented):
Looks good.

src/web/mjs/connectors/SKScans.mjs
27136773: To be honest, i don't know if a cleaned and translated digital media is still called "scanlation" or not, since i guess the term originated from scanning, cleaning and translating physical copies of manga...
I also don't know how to distinguish between all those terms such as cartoon, webtoon, manhwa, manhua ...
I thought webtoons are those less detailed colored korean (originated) online comics and manga the japanese (originated) black and white more detailed comics.
To solve these mystery and define these terms for the tag we may ask the community for information...

XenonR: Point taken. Will keep a community poll in mind.

XenonR (approved):
All good.

PR #1698, opened 2020-05-09
Removed dependencies to HakuNeko's outdated CDN

27136773:


XenonR (approved):
Looks good. Although i do not understand the `const` parts.

27136773:
TuMangaOnline and MangaGo are developed in a private repository and then obfuscated before added to HakuNeko.
This was necessary, because some websites such as MangaGo, TuMangaOnline and JapScan are watching HakuNeko and adapting their protection as soon as HakuNeko bypass them...

PR #1703, opened 2020-05-09
[wanpaman] add website

### LINKED ISSUE #171: Add wanpaman (ONE website)
http://galaxyheavyblow.web.fc2.com/

XenonR:
Closes #171


src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,81 @@
...
+    async _getMangas() {
+        return [
+            {
+                id: this.url,
+                title: this.label
+            }
+        ];
+    }
+
+    async _getChapterList(manga, callback) {
```
27136773: Prefer new interface method `async _getChapters(manga)` over deprecated callback based method

src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,81 @@
...
+        let request = new Request(new URL(manga.id), this.requestOptions);
+
+        Engine.Request.fetchUI( request, 'new XMLSerializer().serializeToString(document);' )
```
27136773: Always return a promise in the injected script, so error handling is assured.
Make better use of fetchUI to perform the logic (the same way you would in the developer console with your browser).

```javascript
let script = `new Promise(resolve => {
    let chapters = [...document.querySelectorAll('body a[href*="imageviewer"]')].map(link => {
        return {
            id: link.pathname,
            title: link.text.trim()
        };
    });
    resolve(chapters);
});`;

let request = ...
// get the chapter list directly ...
return Engine.Request.fetchUI(request, script);
```

src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,81 @@
...
+                }
+                callback(null, chapters.reverse());
+            } )
+            .catch( error => {
+                console.error( error, manga );
+                callback( error, undefined );
+            });
+    }
+
+    async _getPageList(manga, chapter, callback) {
```
27136773: Prefer new interface method `async _getPages(chapter)` over deprecated callback based method

src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,81 @@
...
+                        reject(error);
+                    }
+                }, 3000);
+            });
+        `;
+
+        Engine.Request.fetchUI( request, script )
+            .then( data => {
+                let pages = [];
+                data = this.createDOM(data).querySelectorAll('div.image_item source');
```
27136773: Put this logic into the injected script instead, it already got a ready-to-use DOM from which the chapterlist can be extracted directly (see suggestion for chapters)

27136773 (commented):
I would suggest to utilize `fetchUI` to its fullest, otherwise there wouldn't be much advantage over using `fetchDOM`

src/web/mjs/connectors/WanPaMan.mjs
27136773: You may also consider to get the image data via REST, e.g. `http://galaxyheavyblow.web.fc2.com/fc2-imageviewer/1/3/metadata.json`

src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,61 @@
...
+                };
+            });
+            resolve(chapters.reverse());
+        });`;
+
+        return Engine.Request.fetchUI( request, script );
+    }
+
+    async _getPages(chapter) {
+        let aid = chapter.id.match(/(?:aid=)(\d+)/)[1];
```
27136773: May use URL object to extract certain data from an URL, instead of custom regex,
e.g.
```javascript
let uri= new URL(chapter.id, this.url);
let aid = uri.searchParams.get('aid');
let iid = uri.searchParams.get('iid');
let path = uri.pathname;
// ...
let rest = [path, '', aid, iid, ''].join('/');
```

src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,61 @@
...
+        let iid = chapter.id.match(/(?:iid=)(\d+)/)[1];
+        let pathname = '/' + chapter.id.match(/\/(.+)\//)[1] + '/';
+
+        let request = new Request(
+            new URL(
+                pathname + aid + '/' + iid + '/metadata.json?_=' + Math.round(new Date().getTime()),
```
27136773: `Date.now()` is the shorter static way to get unix timestamp in milliseconds

src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,61 @@
...
+                this.url
+            ),
+            this.requestOptions
+        );
+        let data = await this.fetchJSON(request);
+
+        return data.imageItemList.map(page => {
+            return this.getAbsolutePath(pathname+ + aid + '/' + iid + '/' + page.fileName, this.url);
```
27136773: There are two `+` signs following each other, where one can be removed

27136773 (approved):
Looks good, there is some room for minor improvments ...

src/web/mjs/connectors/WanPaMan.mjs
XenonR: Ohh nice, I really like that.

XenonR:
That was a nice one.

src/web/mjs/connectors/WanPaMan.mjs
```diff
@@ -0,0 +1,58 @@
...
+    async _getPages(chapter) {
+        let uri = new URL(chapter.id, this.url);
+        let rest = [uri.pathname+uri.searchParams.get('aid')];
+        rest.push(uri.searchParams.get('iid'));
+        rest.push( 'metadata.json?_=' + Date.now() );
+
+        let request = new Request( new URL(rest.join('/'), this.url), this.requestOptions );
+        let data = await this.fetchJSON(request);
+
+        return data.imageItemList.map(page => {
```
27136773, line 47: May be safer to work on copy instead modifying the `rest` array,
e.g.
```javascript
    let basePath = rest.slice(0, -1);
    basePath.push(page.fileName);
    // ...
```

Issue #1560, opened 2020-04-20
Suggestion for new site https://copytoon80.com/ 


syleniaa:
Suggestion for new site https://copytoon80.com/  is a good site updated with manhwa raws.

27136773:
Seems to have similar layout as toonkor

PR #1710, opened 2020-05-10
[funbe] add website

### LINKED ISSUE #146: Add support for website: Funbe [KO]
**Site name**: Funbe
**URL**: http://funbe29.com/
**Language**: Korean
**Content**: Webtoons in Korean. This site is quite similar to [Toonkor](https://toonkor.bar/) (ID=toonkor) but the chapters seem to be more frequently updated.
**Free**: Yes, completely
**Require account/login**: No

XenonR:
Closes #146


src/web/mjs/connectors/Funbe.mjs
```diff
@@ -0,0 +1,47 @@
...
+        this.url = 'https://funbe.care';
+        this.language = 'ko';
+    }
+
+    async _getMangas() {
+        let request = new Request(new URL('/망가', this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'div.list-container div.section-item-title a');
+        return data.map(link => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(link.href, this.url).replace('hakuneko://cache/', '/'),
```
27136773, line 19: It seems there is an issue with stripping of the connector origin.
This should not be handled on the connector itself, instead it should be handled in the [base class](https://github.com/manga-download/hakuneko/blob/7fa114a621bc2a737e59f360dff4aa0a0c1bdce5/src/web/mjs/engine/Connector.mjs#L381).

This seems to be a regression, as this worked before without any problems... maybe the last electron update caused this.

27136773: Try to not spread the href attribute manually, instead let the method do it:
`this.getRootRelativeOrAbsoluteLink(link, this.url)`

src/web/mjs/connectors/Funbe.mjs
```diff
@@ -0,0 +1,47 @@
...
+                title: link.getAttribute('alt').trim()
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        let script = `new Promise(resolve => {
+            let pages = [...document.querySelectorAll('div#toon_img img')].map(image => {
+                return new URL(image.getAttribute('src'), document.baseURI).href;
```
27136773: May use `image.src;` it will spread to full qualified USVString based on the document location context
`[...].map(image => image.src);`

src/web/mjs/connectors/Funbe.mjs
```diff
@@ -0,0 +1,47 @@
...
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'table.web_list td.content__title');
+        return data.map(link => {
+            return {
+                id: link.getAttribute('data-role'),
```
27136773: May use auto property `link.dataset.role` to access dataset

src/web/mjs/connectors/Funbe.mjs
```diff
@@ -0,0 +1,47 @@
...
+                title: link.getAttribute('alt').trim()
```
27136773: May use auto property `link.alt` to access property alt

27136773 (approved):
Looks good, may add some improvements

src/web/mjs/connectors/Funbe.mjs
XenonR: I tried that first, but it will return links starting with `hakuneko://cache/`. My test results are that `this.createDOM(data)` is fine but `this.createDOM(data).querySelectorAll('body a')` isn't. Where does `querySelectorAll()` originate from?

27136773: I checked out your branch and it worked fine for me, seems to be an isolated issue...

image

XenonR: Fyi. I'm still trying to resolve this, but something caches the connector itself and changes are not reflected to the running connector (allthough the right contents are shown in the devlopment source view of chrome).

XenonR: The function call itself already looks wrong on my end:
image
And I already see this directly after `fetchDOM`:
image
@ronny1982 can you please test if everything works without the `replace` at your end? I'll remove it then.

27136773:
Funbe uses the exactly same template as Toonkor.
Currently the template from Toonkor is not yet separated, but i would still recommend to use inheritance, so no need to implement Funbe at all:
```javascript
import Toonkor from './Toonkor.mjs';

export default class Funbe extends Toonkor {

    constructor() {
        super();
        super.id = 'funbe';
        super.label = 'Funbe';
        this.tags = [ 'manga', 'korean', 'hentai' ];
        this.url = 'https://funbe.live';
    }
}
```

Funbe also has the same problem of constantly changing domains as Toonkor has, there is no solution yet...

XenonR:
Hmm, I understand the reason for inheritance, but it is impossible to know every website (there are over 600 connectors).

27136773:
I discovered this just by chance...
Related: #1560

Especially asian websites are looking very similar to me, but most of them are implemented completely different so this isn't easy

Ever4engel:
Funbe's new adress https://funbe.live 

this website should use autodetect domain.

27136773:
I extracted the templated from Toonkor (gnuboard5) and also added some similar websites incl. Funbe.
Getting the current URL from Twitter for Toonkor and Funbe is still on the todo list...
See PR #1797

XenonR:
Good stuff. Withdrawing my PR.

PR #1711, opened 2020-05-10
Delete RawLh

Ever4engel:
Duplication of Loveheaven (LHScan)


27136773 (commented):
The related icon should also be removed to prevent cluttering

PR #1721, opened 2020-05-12
[Kuimh] add website support

dodieboy:


src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,72 @@
...
+        let id = uri.pathname + uri.search;
+        let title = data[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let mangaList = [];
+        let request = new Request(this.url + this.path, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryMangasPageCount);
+        let pageCount = parseInt(new RegExp(this.pathMatch).exec(data[0].href)[1]);
```
27136773: May use regex literal instead of converting a constant string into a regex
`paresInt(data[0].href.match(/page=(\d+)/)[1])`

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+export default class Kuimh extends Connector {
+    constructor() {
+        super();
+        super.id = 'kuimh';
+        super.label = '酷爱漫画 (Kuimh)';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.kuimh.com';
+
+        this.path = '/booklist?page=1/';
```
27136773: May be simplified to `/booklist?page=` (which automatically loads the first page when used as is)

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,72 @@
...
+        let pageCount = parseInt(new RegExp(this.pathMatch).exec(data[0].href)[1]);
+        for(let page = 1; page <= pageCount; page++) {
+            let mangas = await this._getMangasFromPage(page);
+            mangaList.push(...mangas);;
+        }
+        return mangaList;
+    }
+
+    async _getMangasFromPage(page) {
+        let request = new Request(new URL(`/booklist?page=${page}`, this.url), this.requestOptions);
```
27136773: When using the simplified path, may be changed to `this.path + page`

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,72 @@
...
+        let data = await this.fetchDOM(request, 'ul.mh-list li div.mh-item-detali h2.title a');
```
27136773: To be consistent may put the constant string into a field (as done with `this. queryMangasPageCount ` or `this.queryPages`)

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,72 @@
...
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.text.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(new URL(this.url + manga.id), this.requestOptions);
+        let data = await this.fetchDOM(request, 'div#chapterlistload ul#detail-list-select li a');
```
27136773: To be consistent may put the constant string into a field (as done with `this. queryMangasPageCount ` or `this.queryPages`)

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,72 @@
...
+                language: ''
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryPages);
+        return data.map(element => {
+            let src = (element.dataset['echo'] || element.src).trim();
```
27136773: May use `this.getAbsolutePath(element.dataset.echo || element, request.url)` which should magically returned the full qualified USVString

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,72 @@
...
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, 'section.banner_detail div.info h1');
+        let id = uri.pathname + uri.search;
+        let title = data[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let mangaList = [];
+        let request = new Request(this.url + this.path, this.requestOptions);
```
27136773: May use `new URL(this.path, this.url)` instead

27136773 (commented):
The lint error needs to be fixed before merge.
Otherwise looks good, there is some room for minor improvments.

dodieboy:
so sry, did so many commit

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,68 @@
...
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.text.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(new URL(this.url + manga.id), this.requestOptions);
```
27136773: May utilize the base parameter, otherwise the URL object would not have any advantage over the composed string (it would fail in case manga.id is already a full qualified USVString)
`new URL(manga.id, this.url)`

src/web/mjs/connectors/Kuimh.mjs
```diff
@@ -0,0 +1,68 @@
...
+export default class Kuimh extends Connector {
+    constructor() {
+        super();
+        super.id = 'kuimh';
+        super.label = '酷爱漫画 (Kuimh)';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.kuimh.com';
+
+        this.path = '/booklist?page=';
+        this.pathMatch = 'page=(\\d+)';
```
27136773, line 13: May prefer regex literal instead of string
`/page=(\d+)/`

27136773 (approved):
Looks good

src/web/mjs/connectors/Kuimh.mjs
dodieboy: i tried `/page=(\d+)/` and `/page=(\\d+)/` before but it will give error `Cannot read property '1' of null`

27136773: Thats weird, may check this out next time some work is needed in this connector...

PR #1741, opened 2020-05-17
[MangaKisa] add website support

dodieboy:


src/web/mjs/connectors/MangaKisa.mjs
```diff
@@ -0,0 +1,54 @@
...
+export default class MangaKisa extends Connector {
+    constructor() {
+        super();
+        super.id = 'mangakisa';
+        super.label = 'MangaKisa';
+        this.tags = [ 'manga', 'webtoon', 'english' ];
+        this.url = 'https://mangakisa.com';
+
+        this.path = '/all-manga';
+        this.queryMangas = 'div.notmain div.lisbox a.an:not([href="/switch"]):not([href="/zion"])';
```
27136773, line 13: Nice find, wasn't aware that `:not` can be chained

27136773 (approved):
Looks good
May try to reduce icon size in future

PR #1744, opened 2020-05-17
Resize some connector icon

dodieboy:
Some connector icon i forget to resize, so i took the time to resize some of the connector icon


27136773 (approved):
Looks good

PR #1709, opened 2020-05-10
LineWebtoon fix age verification

XenonR:
Fixes #1635 

Tested with:
- LineWebtoonEN
- LineWebtoonES
- LineWebtoonFR
- LineWebtoonID
- LineWebtoonTH
- LineWebtoonZH
- LineWebtoonTranslate

PR #1768, opened 2020-05-24
[Line Webtoon] Fix ageGate error (#1767)

### LINKED ISSUE #1767: [Line Webtoon] Connector not working
Yes

Yes

**Describe the bug**
I can't load the chapter list. 

1. Clipboard :  https://www.webtoons.com/en/action/hardcore-leveling-warrior/list?title_no=1221
2. update manga list

When I press F12 to see what's happening I saw the error 302 : it redirect to https://www.webtoons.com/en/ageGate 
With this message in console : 
> https://www.webtoons.com/en/action/hardcore-leveling-warrior/list?title_no=1221 TypeError: Cannot read property 'content' of undefined at LineWebtoonEN._getMangaFromURI (LineWebtoon.mjs:24)

 - Version 6.17

dodieboy:
Resolves #1767

### DIFF
--- a/src/web/mjs/connectors/templates/LineWebtoon.mjs
+++ b/src/web/mjs/connectors/templates/LineWebtoon.mjs
@@ -11,6 +11,7 @@ export default class LineWebtoon extends Connector {
         this.url = undefined;
         this.baseURL = 'https://www.webtoons.com';
         this.requestOptions.headers.set('x-referer', this.baseURL);
+        this.requestOptions.headers.set('x-cookie', 'ageGatePass=true');
     }
 
     get icon() {


27136773 (approved):
Looks good

XenonR:
That's strange. In my previous tests it only works for the first request if the cookie is set. But it seems it's required again now.
Maybe some more logic is required in the future since this basically reverts #1709

27136773:
I will merge it now. Further investigation will be done when another problem with the site is reported.
Maybe it is different for the websites using this template, so it should be done per page instead of the template itself.

PR #1807, opened 2020-05-29
[sukima] Add website

### LINKED ISSUE #86: Add Sukima
https://www.sukima.me/book/search/?free=1

XenonR:
Closes #86


src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+
+export default class Sukima extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'sukima';
+        super.label = 'Sukima';
+        this.tags = [ 'manga', 'japanese' ];
+        this.url = 'https://www.sukima.me';
+        this.language = 'japanese';
```
27136773: May remove this unused field

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+            headers: {
+                'Content-Type': 'application/json;charset=utf-8'
+            }
+        });
+        let categorys = await this.fetchJSON(request);
+
+        let uri;
+        let pages;
+        let mangas = [];
+        for (const category of categorys.rows) {
```
27136773, line 22: Category is a variable, so i recommend using `let`... also don't know if `const` is just syntactic JS sugar since it is still re-assignable opposed to behave like a real constant

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+            if (category.more_btn.link.startsWith('/book/search/')) {
+                uri = new URL(category.more_btn.link, this.url);
+                pages = 1;
+                for (let page = 1; page <= pages; page++) {
+                    request = new Request(new URL('/api/v1/search/', this.url), {
+                        method: 'POST',
+                        body: '{"free":["1","2"],"page":'+page+',"sort_by":"0","tag":["'+uri.searchParams.get('tag')+'"]}',
```
27136773: May use `JSON.stringify(body)` where body is the JSON object (better readability + syntax highlighting)
```javascript
let body = {
    free: [1,2],
    page: page,
    // ...
}
```

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+                        headers: {
+                            'Content-Type': 'application/json;charset=utf-8',
+                        }
+                    });
+
+                    let pageContent = await this.fetchJSON(request);
+                    pages = pageContent.max_page;
```
27136773, line 37: I would expect that this is always the same, so it might be requested outside of the for-loop

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+
+                    for (const manga of pageContent.items) {
+                        mangas.push(
+                            {
+                                id: '/api/book/v1/title/'+manga.title_code+'/',
```
27136773: Looks like only the `title_code` is important, the other parts of the path are redundant and may be composed in `_getChapters`

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+            method: 'POST',
+            body: '{}',
+            headers: {
+                'Content-Type': 'application/json;charset=utf-8',
+            }
+        });
+
+        let chapters = [];
+        let books = await this.fetchJSON(request);
+        for (const book of books.contents) {
```
27136773, line 58: May use `let` instead of const

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+                );
+            }
+        }
+        return chapters;
+    }
+
+    async _getPages(manga) {
+        let script = `
+            new Promise((resolve, reject) => {
+                setTimeout(() => {
```
27136773, line 74: Errors in `setTimeout` are not captured by the `Promise`, so i suggested to encapsulate the whole body of the function within the `setTimeout` in a try...catch block where the error does properly reject the promise, otherwise the Promise may never resolve/reject and HakuNeko hangs
```javascript
setTimeout(() => {
    try {
        // ...
    } catch(error) {
        reject(error);
    }
}, 2500)
```

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+                    let shuffle_data = [];
+                    for (page = 0; page < PAGES_INFO.length; page++) {
```
27136773: Declare `let page` variable in case the injected JavaScript on the website needs to be strict

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+                        )
+                    }
+                    resolve(shuffle_data);
+                }, 2500);
+            });
+        `;
+        let request = new Request(new URL(manga.id, this.url));
+        let data = await Engine.Request.fetchUI(request, script);
+
+        return data.map(page => {
```
27136773: May be simplified to `map(page => this.createConnectorURI(page))`

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+                    id: page.id,
+                    blocklen: page.blocklen,
+                    shuffle_map: page.shuffle_map,
+                }
+            );
+        });
+    }
+
+    async _handleConnectorURI(payload) {
+        while (this.lock) {
```
27136773: Is this lock really required, this is a dangerous bottleneck.
A user may click the preview button for a chapter with e.g. 20 images and the method is now locked for at least 20 seconds... After seeing the cover image, the use decides to look at the cover image of the next chapter and clicks the preview which needs to wait for the previous 20 seconds to finish... Furthermore the method will be locked for another amount of time which stacks up.

A better approach would be to check the response status and repeat the method call after a small amount of time if the request failed.
See: https://github.com/manga-download/hakuneko/blob/40fe23c9bcd71bd453a6ebf9a331250e9a9f02c0/src/web/mjs/connectors/templates/MangaEden.mjs#L87

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+            await this.throttle_download( Math.floor(Math.random() * (2000 - 1000)) + 1000);
```
27136773: May use base method `await this.wait(...)`

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+        }
+        this.lock = true;
+
+        let request = new Request(new URL(payload.id), this.requestOptions);
+        let response = await fetch(request);
+        let blob = await response.blob();
+        let bitmap = await createImageBitmap(blob);
+
+        let canvas = document.createElement('canvas');
```
27136773, line 113: May extract the bitmap descrambling code to an async method that returns the descrambled blob

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+        canvas.width = bitmap.width;
+        canvas.height = bitmap.height;
+        const ctx = canvas.getContext('2d');
+
+        let xSplitCount = Math.floor(bitmap.width / payload.blocklen);
+        let ySplitCount = Math.floor(bitmap.height / payload.blocklen);
+        let count = 0;
+        ctx.drawImage(bitmap, 0, ySplitCount * payload.blocklen + 1, bitmap.width, bitmap.height - ySplitCount * payload.blocklen, 0, ySplitCount * payload.blocklen, bitmap.width, bitmap.height - ySplitCount * payload.blocklen);
```
27136773, line 121: Could this draw call be redundant, as the following for-loop seems to draw all image blocks?

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+
+        for (let i = 0; i < xSplitCount; i++) {
```
27136773: May use more expressive variable name, e.g. `column` instead of `i` and `row` instead of `j`

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+
+        let data = await this._canvasToBlob(canvas);
+
+        this.lock = false;
+        return this._blobToBuffer(data);
+    }
+
+    async _canvasToBlob(canvas) {
+        return new Promise(resolve => {
+            canvas.toBlob(data => {
```
27136773, line 140: Nice move making this callback function from the canvas await-able

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+
+    throttle_download(ms) {
+        return new Promise((resolve) => {
+            setTimeout(resolve, ms);
+        });
+    }
+
+    async _getMangaFromURI(uri) {
+        let title_code = uri.href.split('/');
+        while( title_code[title_code.length-1] == '') {
```
27136773: Could filter be used?
`title_code = title_code.filter(code => code !== '');`

Or maybe even a regex if the title_code has an expected pattern, e.g. `\w` letters and numbers
`title_code = uri.pathname.match(/\/(\w+)\/?$/)[1];`

src/web/mjs/connectors/Sukima.mjs
```diff
@@ -0,0 +1,191 @@
...
+            title_code.pop();
+        }
+        title_code = title_code.slice(-1);
+        let id = '/api/book/v1/title/'+title_code+'/';
+
+        let request = new Request(new URL(id, this.url), {
```
27136773: This code is duplicated, maybe introduce a method `async _fetchPOST(uri, body)` which performs a POST request for the given URL and body object, returning the JSON parsed data object

src/web/mjs/connectors/Sukima.mjs
XenonR: I dream of a feature that reads the language from `this.language` instead of `this.tags`. But I guess it is silly to do it now since a lot of other connectors need to be amended as well.

src/web/mjs/connectors/Sukima.mjs
XenonR: Since `const` is block scoped I try to introduce it wherever a variable does not change anymore. With emphasize on 'try'.

src/web/mjs/connectors/Sukima.mjs
XenonR: I'll always start with the assumption of one page and will read the actual max page from the response of page 1. I opted for this approach to only 1 query instead of 2 querys since I'll anyways receive the pages and mangas in the same JSON object.

src/web/mjs/connectors/Sukima.mjs
XenonR: My problem while testing was that I got blocked for several hours after fetching all images 'at once'. And since `this.config` throttle is not applicable I introduced a mechanism to only process one image at a time. I also double checked and the page is indeed only loading the next image when you flip a page. Therefore I added a reasonable delay between pages. It maybe a shortcoming of my implementation but I think it mimics the website usage.

src/web/mjs/connectors/Sukima.mjs
XenonR: Would make sense if I'm aiming for parallel processing, but I'm intentionally wait for the processing to finish on the current image.

src/web/mjs/connectors/Sukima.mjs
XenonR: No, the site only scrambles what fits into a predefined grid. Therefore most of the time the last 'row' is never scrambled and therefore never 'touched'.

src/web/mjs/connectors/Sukima.mjs
XenonR: Necessity is the mother of all invention xD

src/web/mjs/connectors/Sukima.mjs
XenonR: But I believe then I'll not receive a Promise which I can `await`. This would then be promise(settimeout(promise())) which feels terribly wrong. Do I miss something?

src/web/mjs/connectors/Sukima.mjs
27136773: You can still await when calling the method
```javascript
_handleConnectorURI(payload) {
    let blob = await response.blob();
    blob = await this._descrambleImage(blob, blocks, key, ...);
    return blob;
}

async _descrambleImage(imageData, blocks, key, ...) {
    // ...
}
```

src/web/mjs/connectors/Sukima.mjs
27136773: Could it be that the reason for the temp ban was the amount of the requested images, were missing the referer header? Maybe there is no ban when adding the correct referer for each image (e.g. the URI or parts of it from the HTML page of the chapter)
image

src/web/mjs/connectors/Sukima.mjs
27136773: This is how it should look, so the promise is save and will either resolve or reject:
```javascript
    setTimeout(() => {
        try {
            // simulate an unexpected exception ...
            throw new Error('Test the promise rejects correctly ...');
            // resolve the pending promise
            resolve('OK');
        } catch(error) {
            // reject the pending promise
            reject(error);
        }
    }, 2500);
});
```
Without the try-catch block, the `Promise` will be pending forever, because `resolve('OK')` is never reached and the exception is thrown outside of the promise itself, because `setTimeout` is called asynchronous.

Without an asynchronous call, there is no need for a try-catch block, because the Promise will correctly capture and reject any exception automatically:

```javascript
new Promise(resolve => {
    // simulate an unexpected exception ...
    throw new Error('Test the promise rejects correctly ...');
    // => the exception is captured and the promise rejects automatically
    resolve('OK');
});
```

XenonR: Got it! Thanks.

src/web/mjs/connectors/Sukima.mjs
XenonR: This seems to do the trick. Nice catch!

27136773 (approved):

PR #1808, opened 2020-05-29
Disable cache for development

XenonR:
Disable cache for development by default.
Reference: https://www.electronjs.org/docs/api/command-line-switches#--disable-http-cache


27136773 (approved):
Looks good

PR #1844, opened 2020-06-04
show captcha when website is protected

27136773:


27136773:
Some websites are initialized during application startup to resolve the current url (after redirection),
e.g.
- CopyToon
- HeavenManga
- HolyManga
- KooManga
- MangaFreak
- NewToki

It is probably annoying for the user when one or more captchas needs to be solved during startup for websites that may not be used.

27136773:
The timeout for fetchUI is 60 seconds before the window is closed, which may be a bit short in case a captcha needs to be solved. It might helpful to clear the abort timeout function under that condition and spawn a new one with a higher timeout, e.g. with 120 seconds.

PR #1864, opened 2020-06-08
[mangadex] add setting for users with privacy concerns to opt-out from accessing unofficial/private image nodes

27136773:


XenonR (approved):
All good.

PR #1876, opened 2020-06-12
[vizshonenjump] add node_module + website

### LINKED ISSUE #748: Add Site: Viz Media's Shonen Jump Vault
https://www.viz.com/shonenjump

jdownloader 2 does rip the manga using login credentials, but you have to jump from VPN to VPN.

I love this app that you created. It works wonderful for Crunchyroll.

XenonR:
Added node_module [exif-js](https://github.com/exif-js/exif-js)
Add website support for Viz - Shonen Jump (Closes #748)


src/app/package.json
```diff
@@ -11,6 +11,7 @@
   "main": "main.js",
   "dependencies": {
     "@logtrine/logtrine": "latest",
+    "exif-js": "^2.3.0",
```
27136773: Installing new modules requires to deploy a new desktop client, the current 6.1.7 client does not have this module. Furthermore electron is encouraging the usage of interprocess communication, that means `require()` modules directly in the browser is strongly discouraged and therefore may lead to migration issues someday.

I recommend using the browser version of [exif-js](https://cdn.jsdelivr.net/npm/exif-js) instead. Place it alongside the other [browser libraries](https://github.com/manga-download/hakuneko/tree/master/src/web/js) and add a reference to the [index](https://github.com/manga-download/hakuneko/blob/101a97a29b6b11c88980f53a7d6e95326d73ffab/src/web/index.html#L8).

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
+import Connector from '../engine/Connector.mjs';
+const EXIF = require('exif-js');
```
27136773: Pulling in node modules is discouraged as stated in comment above. Use browser version which makes the library available on global scope.

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+        super.label = 'Viz - Shonen Jump';
+        this.tags = [ 'manga', 'english' ];
+        this.url = 'https://www.viz.com/shonenjump';
+        this.requestOptions.headers.set( 'x-referer', this.url );
+        this.config = {
+            /* Since the website requires us to fetch the image data a few seconds after the API call
+            *  we need to ensure we can download the data fast enough.
+            *  Also the website itself only loads a maximum of 4 pages at a time to ensure that.
+            */
+            concurrent: {
```
27136773: I'm not quite sure, must images be downloaded as fast as possible or must they be throttled to not download as fast as possible. Is it really necessary to introduce throttling on the connector level? The downloader already has throttling support which can be used by configuring the `throttling` property.

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+        let script = `
+            new Promise((resolve, reject) => {
+                setTimeout(() => {
+                    try {
+                        window.loadPages = function loadPages() {
+                            var aj_images = [];
+                            for (var page_count = 0; page_count <= pages; page_count++) 
+                                if (!(page_count < 0 || page_count > pages) && "undefined" == typeof pageImages["page" + page_count]) {
+                                    var aj_req = {
+                                        url: pageUrl + "&manga_id=" + manga_id + "&page=" + page_count,
```
27136773: Is my assumption correct, that these links are only valid for a certain amount of time?
In that case i would recommend a different solution and just get the URL when the image is request in `_handleConnectorURI`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+                        title: 'Ch. ' + chapter.href.match(/chapter-(\d+)\//)[1]
+                    };
+                } catch(error) {
+                    throw new Error('Unknown chapter format. Please report at https://github.com/manga-download/hakuneko/issues');
+                }
+            }
+        });
+    }
+
+    async _getPages(chapter) {
```
27136773, line 77: I suggest to use `fetchRegex` instead of `fetchUI`, since it would consume less resources.
Something like this may work:
```javascript
// Some samples for development reference ...
// Mangas: https://www.viz.com/any/shonen-jump/section/45176/more
// Manga: https://www.viz.com/one-piece
// Chapters: https://www.viz.com/shonenjump/chapters/one-piece
// Chapter: https://www.viz.com/shonenjump/one-piece-chapter-1/chapter/5090
// Page: https://www.viz.com/manga/get_manga_url?device_id=3&manga_id=5090&page=1
let uri = new URL(manga.id, this.url); 
let request = new Request(...);
let data = await this.fetchRegex(request, /var\s+pages\s*=\s*(\d+)/g);
let pageCount = parseInt(data[0]);
let chapterID = 5090; // extracted from chapter url
return Array(pageCount-1).fill().map((_, index) => {
    let page = new URL('https://www.viz.com/manga/get_manga_url');
    page.searchParams.set('device_id', 3); // 3 => probably for desktop with higher quality image link
    page.searchParams.set('manga_id', chapterID); // no idea why VIZ called it mangaID ...
    page.searchParams.set('page', index + 1);
    return this.createConnectorURI(page.href); // x-referer => this.url is sufficient, no need for chapter URL
});

File: src/web/mjs/connectors/VizShonenJump.mjs [hunk 6]
```diff
@@ -0,0 +1,200 @@
...
+    async _handleConnectorURI(payload) {
+        while (this.concurrent >= this.config.concurrent.value) {
+            await this.throttle_connections( Math.floor(Math.random() * (2000 - 1000)) + 1000);
+        }
+        this.concurrent += 1;
+
+        try {
+            // API request
+            let request = new Request(new URL(payload.id, new URL(this.url).origin ), this.requestOptions);
+            request.headers.set('x-referer', new URL(payload.referer, new URL(this.url).origin));
```
27136773, line 96: Referer `this.url` is sufficient, no need for chapter URL

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+            request.headers.set('Accept', 'text/plain, */*; q=0.01');
```
27136773, line 97: I think this header is not required, regarding my curl test, only `x-referer` and `user-agent` (automatically added to each request) are necessary.

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+            // Image request
+            request = new Request(new URL(img_url), this.requestOptions);
+            request.headers.set('x-referer', new URL(payload.referer, new URL(this.url).origin));
+            request.headers.set('crossOrigin', 'Anonymous');
+            request.headers.set('Origin', new URL(this.url).origin);
+
+            response = await fetch(request);
+            const blob = await response.blob();
+            const bitmap = await createImageBitmap(blob);
+            const exif = EXIF.readFromBinaryFile(await blob.arrayBuffer());
```
27136773, line 110: Nice, didn't know array buffer can directly be created from blob

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+const EXIF = require('exif-js');
+
+export default class VizShonenJump extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'vizshonenjump';
+        super.label = 'Viz - Shonen Jump';
+        this.tags = [ 'manga', 'english' ];
+        this.url = 'https://www.viz.com/shonenjump';
```
27136773: Do not not append the path here. I'm not sure if and how this will affect copy & paste feature since a pasted manga URL such as https://www.viz.com/one-piece may not match `this.url => https://www.viz.com/shonenjump`...
Better adding the path in `_getMangas()`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
+import Connector from '../engine/Connector.mjs';
+const EXIF = require('exif-js');
+
+export default class VizShonenJump extends Connector {
+
```
27136773, line 5: There is no implementation of `async _getMangaFromURI(uri)`, therefore copy & paste will not work.

27136773 (commented):
I just figured out the websites is region locked and shows different content based on users IP address.
My comments are based on the region locked version of the website, where https://www.viz.com/shonenjump did not show a manga list at all, but there was one available on https://www.viz.com/any/shonen-jump/section/45176/more
Also all free chapters are redirected to MangaPlus instead of showing on VIZ itself.

I suggest some further investigation on how to handle region differences to prevent users complaining that e.g. the manga list is empty, or there are no chapters ...

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        const request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'a.o_chapter-container');
+        data = data.filter(chapter => chapter.innerText.includes('FREE'));
+
+        return data.map(chapter => {
```
27136773, line 56: Avoid using `try catch` for obvious errors that can be handled with conditions
```javascript
let chapterDIV = chapter.querySelector('.disp-id');
if(chapterDIV) {
    return { id: ..., title: chapterDIV... };
}
let chapterNumber = chapter.href.match(/chapter-(\d+)\//);
if(chapterNumber && chapterNumber.length > 1) {
    return { id: ..., title: chapterNumber... };
}
throw new Error(...);

File: src/web/mjs/connectors/VizShonenJump.mjs [hunk 12]
```diff
@@ -0,0 +1,200 @@
...
+                ctx.drawImage(
+                    bitmap,
+                    Math.floor((piece % 8 + 1) * (x_split + 10)), Math.floor((Math.floor(piece / 8) + 1) * (y_split + 10)), // sx, sy
+                    Math.floor(x_split), Math.floor(y_split), // sWidth, sHeight
+                    Math.floor((shuffle_map[piece] % 8 + 1) * x_split), Math.floor((Math.floor(shuffle_map[piece] / 8) + 1) * y_split), // dx, dy
+                    Math.floor(x_split), Math.floor(y_split)); // dWidth, dHeight
+            }
+
+            let data = await this._canvasToBlob(canvas);
+            this.concurrent -= 1;
```
27136773: Should do this in `finally` block instead of `try` and `catch`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+            {
+                id: page.url,
+                referer: chapter.id
+            }
+        ));
+    }
+
+    async _handleConnectorURI(payload) {
+        while (this.concurrent >= this.config.concurrent.value) {
+            await this.throttle_connections( Math.floor(Math.random() * (2000 - 1000)) + 1000);
```
27136773: Base class already provides a wait method `this.wait()`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,200 @@
...
+        }
+        this.concurrent += 1;
+
+        try {
+            // API request
+            let request = new Request(new URL(payload.id, new URL(this.url).origin ), this.requestOptions);
```
27136773: Just use `this.url` instead of `new URL(this.url).origin`, the URL constructor will magically resolve the path correctly

src/web/mjs/connectors/VizShonenJump.mjs
XenonR: After issuing the API call we retrieve the image url, but this is only valid fro a few seconds. So if we would retrieve all image urls we require to download them all at once (start immediately). Therefore later pages may start to download too late.
My understanding for `throttling` was that it does not work wit `_handleConnectorURI()`?

src/web/mjs/connectors/VizShonenJump.mjs
XenonR: At this point in the communication they are not yet time sensitive. The time sensitive API request is already inside of `_handleConnectorURI()`.

src/web/mjs/connectors/VizShonenJump.mjs
XenonR: Maybe, but I follow what a real browser would send. Therefore I deem `payload.refer` 'more' correct.

src/web/mjs/connectors/VizShonenJump.mjs
XenonR: In my tests I only got a 403 without that header. It is definitely required.

src/web/mjs/connectors/VizShonenJump.mjs
XenonR: Right, stepped into that one again...

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,191 @@
...
+        super();
+        super.id = 'vizshonenjump';
+        super.label = 'Viz - Shonen Jump';
+        this.tags = [ 'manga', 'english' ];
+        this.url = 'https://www.viz.com';
+        this.requestOptions.headers.set( 'x-referer', this.url );
+        this.config = {
+            throttle: {
+                label: 'Throttle Requests [ms]',
+                description: 'Enter the timespan in [ms] to delay consecuitive HTTP requests.\nThe website may ban your IP for to many consecuitive requests.',
```
27136773: Its more something like `The image links may expire for too many consecutive requests and therefore the downloads may fail.`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    async _getPages(chapter) {
+        const request = new Request(new URL(chapter.id, this.url));
+        const data = await this.fetchRegex(request, /var\s+pages\s*=\s*(\d+)/g);
+        const pageCount = parseInt(data[0]);
+        const chapterID = chapter.id.match(/chapter\/(\d+)/)[1];
+
+        return Array(pageCount+1).fill().map((_, index) => {
+            let page = new URL('https://www.viz.com/manga/get_manga_url');
```
27136773: May use `new URL('/manga/get_manga_url', this.url)`

27136773 (approved):
Looks good so far, ready for merge

PR #1883, opened 2020-06-14
Delete Mangazuki.info

dodieboy:


27136773 (approved):
No clues about moving to a different domain, so it seems the website is really gone.
Looks good.

PR #1897, opened 2020-06-18
[9anime, kissanime] fixed regression: only map images but not video streams

### LINKED ISSUE #1871: Nothing on 9anime Loads
> **From:** HakuNeko User

**Version:** [6.1.7@aa9bb2](https://github.com/manga-download/hakuneko/commits/aa9bb2)

9anime used to be the only anime site which worked for me (with the exception of anime hentai sites, which I never tested and so cannot be sure if they worked or not), and I was fine with that. Now, no episodes on 9anime load. When I try to load an episode, I get one of two errors: pages.map is not a function, or it failed to load the episode within the given timeout of 60 seconds. In my experience with this issue, the same episode always has the same error when trying to play it. 9anime's manga list is fine, though. Most manga websites work.

Below are the answers to the questions asked in the 'Making an informative help request' section of the troubleshooting guide.
My operating system is PureOS.
I am using desktop client version 6.1.7.
My rev. number is aa9bb2.
This issue is about 9anime not working, but as stated before none of the other non-hentai anime sites worked before this issue occured. I may write a separate help request about them, or I may not.
The settings for HakuNeko are the default ones, with the exception of the Frontend, which is set on 'Ken's Daedal Dark', although the error persists regardless of the Frontend setting.
After the problem occurs, either https://pasteboard.co/JcvvtLm.png or https://pasteboard.co/Jcvwo0V.png appears in the error log (the images are consistent with what appears in the error log, with the exception of the episode links, where the links change depending on the episode), for either a timeout error or a pages.map error respectively.

<details><summary>What has been done?</summary>

☑ Read the Quick Start Guide: https://git.io/hakuneko-quickstart
☑ Tried Troubleshooting: https://git.io/hakuneko-troubleshooting
☑ Done Website Interaction: https://git.io/hakuneko-websiteinteraction
☑ Searched for Similar Issues: https://git.io/hakuneko-issuesearch

</details>

27136773:
Resolves #1871

### DIFF
--- a/src/web/mjs/engine/Chapter.mjs
+++ b/src/web/mjs/engine/Chapter.mjs
@@ -111,7 +111,7 @@ export default class Chapter extends EventTarget {
                         this.pageCache = [];
                         if( !error ) {
                             // HACK: bypass 'i0.wp.com' image CDN to ensure original images are loaded directly from host
-                            this.pageCache = pages.map(page => page.replace(/\/i\d+\.wp\.com/, ''));
+                            this.pageCache = Array.isArray(pages) ? pages.map(page => page.replace(/\/i\d+\.wp\.com/, '')) : pages;
                         }
                         callback( error, this.pageCache );
                     } );


XenonR (approved):
All good

PR #1907, opened 2020-06-19
[serimanga] added website support

### LINKED ISSUE #461: please add this website serimanga.com
https://serimanga.com/?a=a
ı hope you add this website for turkey
thanks for read

27136773:
Resolves #461


src/web/mjs/connectors/SeriManga.mjs
```diff
@@ -0,0 +1,66 @@
...
+        this.tags = [ 'manga', 'webtoon', 'turkish' ];
+        this.url = 'https://serimanga.com';
+    }
+
+    async _getMangaFromURI(uri) {
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, 'div.seri-img div.name');
+        let id = uri.pathname + uri.search;
+        let title = data[0].textContent.trim();
+        return Promise.resolve(new Manga(this, id, title));
```
XenonR: I do not see why this can not be simplified to `return new Manga()`

XenonR (approved):
Looking good.

src/web/mjs/connectors/SeriManga.mjs
27136773: Copy & Paste relict

PR #1919, opened 2020-06-20
[flatmanga] add support for base64 encoded image links

27136773:


XenonR (approved):
The `atob()` `catch` looks strange, but there seems to be no better way.
Looking fine.

27136773:
It would require to have specific domain knowledge to distinguish valid links from base64 encoded text.

PR #1957, opened 2020-06-28
fixes magicScroll when no gap between images

Sheepux:
Current page within view was not correctly identified when there is no gap between images.


27136773 (approved):
Looks good

PR #1958, opened 2020-06-28
[zynk] added template based on mhxk

### LINKED ISSUE #1759: Add website template for 1359mh / 517mh
1 / Hello, now I'm very happy and suddenly there is a website that I want has been added to Hakuneko, I recommend a few more websites maybe more people will need like me: ohmanhua.com, 1359mh.com , www.517mh.net.
2 / Is the chapter chapter issue fixed? For example, there is no need to detect the number of chapters in the title of the manga.

27136773:
Resolves #1759


src/web/mjs/connectors/templates/ZYMK.mjs
```diff
@@ -0,0 +1,111 @@
...
+
+    async _getChapters(manga) {
+        let uri = new URL(manga.id, this.url);
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryChapters);
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
+                title: element.text.trim(),
+                language: ''
```
XenonR, line 88: May remove unused element.

XenonR (approved):
Looking good. Major work👍

PR #1963, opened 2020-06-29
Add Discord presence

### LINKED ISSUE #1925: [Feature]Discord Rich Presence
you know how when you are playing games it says what u are playing 

like taiga(a somewhat similar tool but for anime) that says what you are watching 
but for reading obv so (reading "chapter" of "name of manga" hakuneko) with the hakunoko icon 

As far as i am aware there isn't a way to do this other than manually 
you could set it to say u were playing hakuneko but that its all it would say

https://i.imgur.com/Syp7WYh.png
https://i.imgur.com/podi2Uh.png

XenonR:
Adds Discord 'rich presence' to HakuNeko.
image
image

Presence is disabled by default and can be set to either:
image

Closes #1925


package.json
```diff
@@ -45,5 +45,8 @@
     "test:app": "jest --color --reporters=default --reporters=jest-junit --testPathPattern=/src/app",
     "test:web": "jest --color --reporters=default --reporters=jest-junit --testPathPattern=/src/web",
     "test:e2e": "jest --color --reporters=default --reporters=jest-junit --testPathPattern=/src/__tests__"
+  },
+  "dependencies": {
+    "discord-rpc": "^3.1.1"
```
27136773: Should be in `src/app/package.json`, otherwise it will not be built in the client

src/web/mjs/engine/Chapter.mjs
```diff
@@ -89,6 +89,7 @@ export default class Chapter extends EventTarget {
      * and a reference to the page list (undefined on error).
      */
     getPages( callback ) {
+        document.dispatchEvent(new CustomEvent(EventListener.onChapterGetPages, { detail: this }));
```
27136773: This will fire every time chapter starts downloading.
May use the UI preview button click event instead to fire the event.

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
+const discordPresenceId = '726702836775256094';
+const DiscordRPC = require('discord-rpc');
+DiscordRPC.register(discordPresenceId);
+
+export default class DiscordPresence {
+
+    constructor(settings) {
+        this.rpc = null;
+        this.IpcBytes = 0; // keep track of IPC connection
```
27136773, line 10: No need to use a filed when only used in a single method.
Use a scoped variable in the corresponding function.

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+        // Manga selected
+        document.addEventListener( EventListener.onMangaClicked, this._onMangaClicked.bind(this) );
+        // get chapter pages (Reader/Downloader)
+        document.addEventListener( EventListener.onChapterGetPages, this._onChapterGetPages.bind(this) );
+
+        this.startDiscordPresence();
+    }
+
+    _onSettingsChanged() {
+        switch (this._settings.discordPresence.value) {
```
27136773: Could be simplified
```javascript
this.enabled = this._settings.discordPresence.value !== 'none';
this.enabledHentai = this._settings.discordPresence.value === 'hentai';
```

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+            this.rpc.on('ready', () => {
+                this.status.startTimestamp = + new Date();
+                
+                // some delay for Discord to be receptive
+                setTimeout( () => {
+                    this.updateStatus();
+                }, 2000);
+
+                // activity can only be set every 15 seconds
+                setInterval(() => {
```
27136773: Reference should be stored in class field, so it can be stopped gracefulley

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+                this.statusNew = false;
+            } else if (this.rpc.transport.socket.bytesWritten == this.IpcBytes && this.statusNew) {
+                console.warn('WARNING: DiscordPresence - Lost connection to Discord.');
+                this.stopDiscordPresence();
+            }
+        }
+    }
+
+    stopDiscordPresence() {
+        this.statusNew = false;
```
27136773, line 107: Should also `clearInterval(reference)` that otherwise continues to tick every 15 seconds
May also lead to spawn multiple intervals when not cleared if a user switches on/off/on/off/on/off/... discord in settings

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+                }, 2000);
+
+                // activity can only be set every 15 seconds
+                setInterval(() => {
+                    this.updateStatus();
+                }, 15200);
+            });
+            this.rpc.login({ clientId: discordPresenceId })
+            .catch (error => {
+                if (error.message == 'Could not connect') {
```
27136773: Is this check reliable? Is the text always english and case sensitive?
Maybe check `error.code` if available such as 4001, 4002 and whatever the discord codes are...
https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway-gateway-close-event-codes

src/web/mjs/engine/Enums.mjs
```diff
@@ -7,6 +7,9 @@ export default class Enums {
         return {
             onMangaStatusChanged:     'onMangaStatusChanged', // ...
             onChapterStatusChanged:   'onChapterStatusChanged', // ...
+            onChapterGetPages:        'onChapterGetPages',
```
27136773: May use more consistent naming scheme, e.g.
onSelectConnector, onSelectManga, onSelectChapter

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+            this.statusNew = true;
+            this.startDiscordPresence();
+        } else {
+            this.stopDiscordPresence();
+        }
+    }
+
+    _onSelectConnector(e) {
+        // Hentai check
+        if(e.detail.tags.includes('hentai') || e.detail.tags.includes('porn')) {
```
27136773: Not sure if it is reliable to use the website tags.
From technical point the comparison is case sensitive which will fail for `Hentai`.
There might also be upcoming tags like `Yuri` which will very likely being forgotten to add here...
A website may not be tagged as `hentai`, but still contain some NSFW mangas.

A show all or nothing approach is probably easier to handle from a developer point of view

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+        this.statusNew = true;
+        if (this.enabled && !this.rpc) this.startDiscordPresence();
+    }
+
+    async updateStatus() {
+        if(this.rpc) {
+            if (this.enabled && this.statusNew) {
+                this.IpcBytes = this.rpc.transport.socket.bytesWritten;
+                if( !this.hentai || (this.hentai && this.enabledHentai)) {
+                    this.rpc.setActivity(this.status);
```
27136773, line 89: If supported, may set a custom type, because HakuNeko is not really a game that is played
https://discord.com/developers/docs/topics/gateway#activity-object-activity-structure

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+
+    _onChapterGetPages(e) {
+        // Hentai check
+        if(e.detail.manga.connector.tags.includes('hentai') || e.detail.manga.connector.tags.includes('porn')) {
+            this.hentai = true;
+        } else {
+            this.hentai = false;
+        }
+
+        this.status['details'] = 'Reading ' + e.detail.manga.title;
```
27136773: It might also be an anime, where reading does not fit

27136773 (commented):
May execute `npm run format:web` to auto-fix the lint errors
Some minor adjustments should be made before merge

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+
+        if (this.enabled) {
+            this.statusNew = true;
+            this.startDiscordPresence();
+        } else {
+            this.stopDiscordPresence();
+        }
+    }
+
+    _onSelectConnector(e) {
```
27136773: Use more expressive argument name, e.g. `event`, `sender`, ...

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        this.status['details'] = 'Browsing ' + e.detail.label;
+        if (this.status.state) delete this.status.state;
+        this.status.startTimestamp = + new Date();
+        this.statusNew = true;
+        if (this.enabled && !this.rpc) this.startDiscordPresence();
+    }
+
+    _onMangaClicked(e) {
```
27136773: Use more expressive argument name, e.g. `event`, `sender`, ...

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        this.status['details'] = 'Browsing ' + e.detail.connector.label;
+        this.status['state'] = 'Looking at ' + e.detail.title;
+        this.status.startTimestamp = + new Date();
+        this.statusNew = true;
+        if (this.enabled && !this.rpc) this.startDiscordPresence();
+    }
+
+    _onChapterGetPages(e) {
```
27136773: Use more expressive argument name, e.g. `event`, `sender`, ...

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -0,0 +1,167 @@
...
+        if (this.rpc) {
+            this.rpc.clearActivity();
+            this.rpc.destroy();
+            this.rpc = null;
+        }
+    }
+
+    async startDiscordPresence() {
+        if (!this.rpc) {
+            this.rpc = new DiscordRPC.Client({ transport: 'ipc' });
```
27136773: Could drop `if` when using short circuiting ||
`this.rpc = this.rpc || new ...`

27136773:
Additional use case:
Should HakuNeko continue to update discord status, even when not in foreground?
In this case it may overwrite the status from a game that is currently focused...

src/web/mjs/engine/DiscordPresence.mjs
Sheepux: How about "Viewing", should match both cases :)

src/web/mjs/engine/Chapter.mjs
XenonR: This is intended, since I personally never read a manga in HakuNeko I would like to state my latest 'activity'.

src/web/mjs/engine/DiscordPresence.mjs
XenonR: Unfortunately discord-rpc does not expose an error code:
image
Also this is an error that triggers if there is no IPC connection to Discord possible, so no Discord API can be leveraged. But i changed it to `error.message.match(/[A-z]+/g).join('').toUpperCase() == 'COULDNOTCONNECT'` to make a it a little bit more robust.

src/web/mjs/engine/DiscordPresence.mjs
XenonR: What else to use :) That's what tags are for. Changed comparison to `e.detail.tags.map(t => t.toLowerCase()).includes('hentai')` and `e.detail.tags.map(t => t.toLowerCase()).includes('porn')`

src/web/mjs/engine/DiscordPresence.mjs
XenonR: Yeah i saw that. Unfortunately discord-rpc doesn't support it (again):
image

XenonR:
> 
> 
> Additional use case:
> Should HakuNeko continue to update discord status, even when not in foreground?
> In this case it may overwrite the status from a game that is currently focused...

It doesn't, I keep track if the latest 'activity' was already send with `this.statusNew`. I will only update if there is a new 'activity' to report.

src/web/mjs/engine/DiscordPresence.mjs
XenonR: I'm not sure if this is possible without waiting. By keeping track globally I can check the IPC state one interval later which basically allows 15 seconds to pass.

src/web/mjs/engine/DiscordPresence.mjs
XenonR: I thought the interval is attached to `this.rpc`. Can you please point me in the right direction?

src/web/mjs/engine/DiscordPresence.mjs
27136773: OK i understand the intention now

src/web/mjs/engine/DiscordPresence.mjs
27136773: Regex might be more convenient for case insensitive (partial) match
Full with Single: `/^sample$/i.test('SaMpLe')`
Partial with Multi: `/(hentai|porn)/i.test(tag)`

src/web/mjs/engine/DiscordPresence.mjs
27136773: Introduce a field e.g. `this.statusInterval = null;`
Then assign it in the method `this.statusInterval = setInterval(this.updateStatus.bind(this, 15200));`
To stop interval `this.clearInterval(this.statusInterval);`

src/web/mjs/engine/DiscordPresence.mjs
27136773: Also a nice example for short Regex.test
`if(/could\s*not\s*connect/i.test(error.message)) { ... error }`

27136773:
> > Additional use case:
> > Should HakuNeko continue to update discord status, even when not in foreground?
> > In this case it may overwrite the status from a game that is currently focused...
> 
> It doesn't, I keep track if the latest 'activity' was already send with `this.statusNew`. I will only update if there is a new 'activity' to report.

This contradicts the `Chapter.getPages()` based trigger which will update the status in discord whenever a new chapter is downloaded. So adding hundred chapters to HakuNeko then minimize and open up a game, will lead to HakuNeko updating discord status every 15 seconds, even when minimized.

XenonR:
> This contradicts the `Chapter.getPages()` based trigger which will update the status in discord whenever a new chapter is downloaded. So adding hundred chapters to HakuNeko then minimize and open up a game, will lead to HakuNeko updating discord status every 15 seconds, even when minimized.

True. I finally found the development info of Discord:
> Q: What happens if someone has more than one game running that supports Rich Presence?
Due to recent changes in our infrastructure for support of multi-activities, the behavior of multiple connected Rich Presence apps has changed from what it was before. Previously, whichever application was focused would be the presence that was shown. With the recent changes, **the application that connected first is now displayed**.
[Source](https://discord.com/developers/docs/rich-presence/faq#q-what-happens-if-someone-has-more-than-one-game-running-that-supports-rich-presence)

So it doesn't even matter if it updates or not.

And rich presence != presence. Where rich presence > presence. So as long as an rich presence app runs it will take precedence.

27136773 (approved):
Seems it would be ready to go

PR #1965, opened 2020-06-30
[kissmanga] re-added manga list

27136773:


src/web/mjs/connectors/KissManga.mjs
```diff
@@ -23,8 +24,27 @@ export default class KissManga extends Connector {
     }
 
     async _getMangas() {
-        let msg = 'This website does not provide a manga list, please copy and paste the URL containing the chapters directly from your browser into HakuNeko.';
-        throw new Error(msg);
+        let mangaList = [];
+        let request = new Request(this.url + '/MangaList?page=', this.requestOptions);
```
XenonR: May drop unused search parameters. `https://kissmanga.com/MangaList` seems sufficient.

XenonR (approved):
May drop unused search parameters. Otherwise ok.

src/web/mjs/connectors/KissManga.mjs
27136773: done

PR #2008, opened 2020-07-10
Added support for ManhuaES

### LINKED ISSUE #1961: please add website manhuaes.com
**Name of the website**
manhuaes

- Site: https://manhuaes.com/
- Manga List: https://manhuaes.com/category-comics/manga/
- Manga example: https://manhuaes.com/comics/emperor-lingwu/
- Chapter online viewer example: https://manhuaes.com/chapter/emperor-lingwu-chapter-66/

**Languages**
English

newaccforhc:
Fixes #1961 

Basically the site was nearly same as mavimanga... So, I just changed a few things in it..
**NOTE:** The Mangalist feature may need to be updated in future because, currently the site only contains 2 pages... So, there wasn't anything like last page.


src/web/mjs/connectors/Manhuaes.mjs
```diff
@@ -0,0 +1,59 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class ManhuaES extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'manhuaes';
+        super.label = 'Manhua ES';
+        this.tags = [ 'manga', 'English' ]; //should I add 'Webtoon' tag as well?
```
27136773: When the website provides mangas, then set `manga`
When the website provides webtoons, then set `webtoon`
When the website provides both, then add `manga` and `webtoon`

Use lowercase for tags, otherwise there will be two groups `english` and `English` with the latter only containing this single website

src/web/mjs/connectors/Manhuaes.mjs
```diff
@@ -0,0 +1,59 @@
...
+        let data = await this.fetchDOM(request, 'h1.title-detail');
+        let id = uri.pathname;
+        let title = data[0].textContent;
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let request = new Request(new URL('/category-comics/manga/', this.url), this.requestOptions);
+        let pages = await this.fetchDOM(request, 'ul.pagination li:nth-last-child(1) > a');
+        pages = Number(pages[0].pathname.match(/\/(\d+)\/$/)[1]);
```
27136773: It seems to be easier to use `Number(pages[0].text)` instead of extracting the number via regex from the `href` attribute

27136773 (approved):
Ready for merge

PR #2012, opened 2020-07-10
Fix SmackJeeves

### LINKED ISSUE #1964: Smack Jeeves is not working
> **From:** HakuNeko User

**Version:** [6.1.7@d8c08b](https://github.com/manga-download/hakuneko/commits/d8c08b)

When selecting Smack Jeeves as a connector it errors out on an update list. Using the Clipboard and downloading a single URL also fails.

<details><summary>What has been done?</summary>

☑ Read the Quick Start Guide: https://git.io/hakuneko-quickstart
☑ Tried Troubleshooting: https://git.io/hakuneko-troubleshooting
☑ Done Website Interaction: https://git.io/hakuneko-websiteinteraction
☑ Searched for Similar Issues: https://git.io/hakuneko-issuesearch

</details>

XenonR:
~~Adds node_module [he](https://github.com/mathiasbynens/he) for DOM independent html entity decoding.~~
Closes #1964


src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
-     *
-     */
     constructor() {
         super();
         super.id = 'smackjeeves';
         super.label = 'Smack Jeeves';
         this.tags = [ 'webtoon', 'english' ];
-        this.url = 'http://www.smackjeeves.com';
+        this.url = 'https://www.smackjeeves.com';
+        this.he = require('he');
```
27136773: `he` should be available as global (on window), if `require` is used something seems not to be as intended.
`require` is only for installed electron packages.

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
+                } catch(error) {
+                    reject(error);
                 }
-            } );
+            });
+        `;
+        const request = new Request(new URL(url), this.requestOptions);
+        const data = await Engine.Request.fetchUI(request, script);
+
+        return data.navigation.items;
```
27136773: May use `data.navigation.items.map(item => item.val)`, since the val is the only thing that is used later on

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
-                callback( error, undefined );
-            } );
+    async _getMangas() {
+        const categories = await this._getCategories( new URL('/discover?type=genre', this.url));
+
+        let manga_list = [];
+        for (const category of categories) {
+            let url = new URL('/api/getTitleListByGenreDiscover', this.url);
+            url.searchParams.set('genre', category.val);
+            url.searchParams.set('order', 'newarriva');
```
27136773: Could this be a typo newarriva`l`, or mistake on website

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
+
+            let pages = 1;
```
ronny1982, line 39: Each category only has a single page? This would make the following for loop obsolete.

src/web/index.html
```diff
@@ -14,6 +14,7 @@
     <script src="./js/ass.min.js" defer></script>
     <script src="./js/sql.min.js" defer></script>
     <script src="./js/exif-js.min.js" defer></script>
+    <script src="./js/he.min.js" defer></script>
```
27136773: No need for this dependency, HakuNeko can utilize the underlying Browser to do that with 3 lines of code:
https://github.com/manga-download/hakuneko/blob/a361183c2e96075650f8d83f7861bb3c81a96ad3/src/web/mjs/connectors/MangaLife.mjs#L26

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
+            for (let page = 1; page <= pages; page++) {
+                url.searchParams.set('page', page);
+                const data = await this._fetchPOST(url);
+                pages = data.result.totalPageCnt;
+
+                manga_list.push(...data.result.list.map(element => {
+                    return {
+                        id: this.getRootRelativeOrAbsoluteLink(element.titleUrl, this.url),
+                        title: this.he.decode( element.title ).trim()
```
27136773: Don't use `he` as described above

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
-                callback( null, chapterList );
-            } )
-            .catch( error => {
-                console.error( error, manga );
-                callback( error, undefined );
-            } );
+        return data.result.list.map(element => {
+            return {
+                id: element.articleUrl,
+                title: this.he.decode( element.articleTitle ).trim()
```
27136773: Same here, drop `he` and use a small local method to replace HTML entities

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
+            url.searchParams.set('genre', category.val);
+            url.searchParams.set('order', 'newarriva');
+
+            let pages = 1;
+            for (let page = 1; page <= pages; page++) {
+                url.searchParams.set('page', page);
+                const data = await this._fetchPOST(url);
+                pages = data.result.totalPageCnt;
+
+                manga_list.push(...data.result.list.map(element => {
```
27136773: May use e.g. `item` instead of `element`, since the latter may make developers believe this is an HTML element (which wouldn't require to deal with HTML entities)

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
+        const data = await this._fetchPOST(url);
 
-                }
-                callback( null, chapterList );
-            } )
-            .catch( error => {
-                console.error( error, manga );
-                callback( error, undefined );
-            } );
+        return data.result.list.map(element => {
```
27136773: May use e.g. `item` instead of `element`

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
-                let pageLinks = data.map( element => this.createConnectorURI( this.getAbsolutePath( element.value, request.url ) ) );
-                callback( null, pageLinks );
-            } )
-            .catch( error => {
-                console.error( error, chapter );
-                callback( error, undefined );
-            } );
+    async _getPages(chapter) {
+        const script = `
+            new Promise((resolve, reject) => {
```
27136773: Can be simplified to `new Promise(resolve => resolve(cmnDate.comicData));`
A promise will reject exceptions automatically (as long as they are not within an async scope, e.g. setTimeout)

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
+                    resolve(cmnData.comicData);
+                } catch(error) {
+                    reject(error);
+                }
+            });
+        `;
+        const request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        const data = await Engine.Request.fetchUI(request, script);
+
+        return data.map(element => {
```
27136773: May embed the `map` into the injected script and return `fetchUI` directly

src/web/mjs/connectors/SmackJeeves.mjs
```diff
@@ -1,130 +1,99 @@
...
-        /*
-         * TODO: only perform requests when from download manager
-         * or when from browser for preview and selected chapter matches
-         */
-        return this.fetchDOM( request, 'source#comic_image' )
-            .then( data => super._handleConnectorURI( this.getAbsolutePath( data[0], request.url ) ) );
+    async _fetchPOST( uri) {
+        const request = new Request(new URL(uri.pathname, this.url), {
+            method: 'POST',
+            body: uri.searchParams.toString(),
```
27136773: Placing the form data into the GET parameters of the URL is misleading, i would suggest to separate parameter concerns (what if you have to fetch an URL that actually has some required GET parameters?):

```javascript
async _fetchPOST(uri, form) {
    fetch(uri, {
        body: form.toString()
    });
}

async foobar() {
    let uri = new URL('http://foo.bar?genre=bla')
    let form = new URLSearchParams();
    form.set('token', 123);
    form.set('page', 123);
    let data = await this._fetchPOST(uri, form);
    // ...
```

27136773 (commented):
The `he` dependency should be removed, then it is ready to go

src/web/mjs/connectors/SmackJeeves.mjs
XenonR: If it is, it was the websites programmers typo 😂

src/web/mjs/connectors/SmackJeeves.mjs
XenonR: What's the idea here? Is it better practice or easier readable for others?

src/web/mjs/connectors/SmackJeeves.mjs
XenonR: I like that. `.map()` is still a thing where i learn something everytime i use it.

src/web/mjs/connectors/SmackJeeves.mjs
XenonR: You're right. Can be misleading. Modified.

src/web/mjs/connectors/SmackJeeves.mjs
ronny1982: Readability, i was confused at first because the loop seemed to run only once and then noted that the `pages` variable in the condition of the for loop got changed dynamically

27136773: Readability:
```javascript
let bike = new Car();
let car = new Car();
```

27136773 (approved):
Really sleepy right now, but i would still dare to say it is ready for merge

PR #2021, opened 2020-07-11
[mangalivre] add website support

### LINKED ISSUE #159: Add support manga MangaLivre
> **From:** HakuNeko User

Portuguese language:
https://mangalivre.com/

XenonR:
Closes #159
Note: site is geoblocked


src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+        super.label = 'Manga Livre';
+        this.tags = [ 'manga', 'portuguese', 'webtoon' ];
+        this.url = 'https://mangalivre.com';
+    }
+
+    async _getMangas() {
+        let page = 1;
+        let morePages = true;
+
+        let request, data;
```
27136773: No need for ahead declaration, there is no performance/memory gain and in case of refactoring it needs to be changed in two places instead of one.

src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+        let mangas = [];
+
+        while ( morePages ) {
+            try {
+                request = new Request(new URL('/series/index/nome/todos?page=' + page, this.url), this.requestOptions);
```
27136773: May use side effects `new Request('...' + page++, ...);` will use current value of page and increment the value afterwards.

src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+                data = await this.fetchDOM(request, 'div.seriesList ul.seriesList li > a.link-block');
+                page++;
+
+                for (const manga of data) {
+                    mangas.push(
+                        {
+                            id: manga.pathname.match(/\d+$/)[0],
+                            title: manga.title.match(/^Ler\s(.*)\sOnline$/) ? manga.title.match(/^Ler\s(.*)\sOnline$/)[1].trim() : manga.title
```
27136773: Redundant code, may use a local variable for the duplicated regex result

src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+        let request, data;
+        let mangas = [];
+
+        while ( morePages ) {
+            try {
+                request = new Request(new URL('/series/index/nome/todos?page=' + page, this.url), this.requestOptions);
+                data = await this.fetchDOM(request, 'div.seriesList ul.seriesList li > a.link-block');
+                page++;
+
+                for (const manga of data) {
```
27136773: May utilize `map` instead of loop

src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+        while ( morePages ) {
+            request = new Request(url, {
+                headers: {
+                    'X-Requested-With': 'XMLHttpRequest',
+                    'Accept': 'application/json, text/javascript, */*; q=0.01'
+                }
+            });
+            data = await this.fetchJSON(request);
+            page++;
+            url.searchParams.set('page', page);
```
27136773: May use side effect of increment as described previously

src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+
+            if ( data.chapters) {
+                for (const chapter of data.chapters) {
```
27136773: May utilize `map`

src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+        // Order matters !
+        url.searchParams.set('page', 1);
+        url.searchParams.set('id_serie', online_id);
+
+        const request = new Request(new URL(url.href), this.requestOptions);
+        const data = await this.fetchJSON(request);
+
+        return new Manga(
+            this,
+            data.chapters[0].id_serie,
```
ronny1982, line 108: Isn't this the same as `online_id`

src/web/mjs/connectors/MangaLivre.mjs
```diff
@@ -0,0 +1,113 @@
...
+
+    async _getMangaFromURI(uri) {
+        const online_id = uri.pathname.match(/\d+$/)[0];
+        let url = new URL('/series/chapters_list.json', this.url);
+        // Order matters !
+        url.searchParams.set('page', 1);
+        url.searchParams.set('id_serie', online_id);
+
+        const request = new Request(new URL(url.href), this.requestOptions);
+        const data = await this.fetchJSON(request);
```
ronny1982, line 104: Is this request required

27136773 (approved):
Can be merged

src/web/mjs/connectors/MangaLivre.mjs
XenonR: Didn't think I would like `.map` better. But the code looks much better when written as `.map`

PR #2024, opened 2020-07-11
[furyosquad] Add website support

### LINKED ISSUE #311: Add FuryoSquad
http://www.furyosquad.com

XenonR:
Closes #311


src/web/mjs/connectors/FuryoSquad.mjs
```diff
@@ -0,0 +1,50 @@
+import Connector from '../engine/Connector.mjs';
+
+export default class FuryoSquad extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'furyosquad';
+        super.label = 'Furyo Squad';
+        this.tags = [ 'manga', 'french', 'high-quality' ];
+        this.url = 'http://www.furyosquad.com/';
```
27136773: May remove trailing slash

src/web/mjs/connectors/FuryoSquad.mjs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    async _getMangas() {
+        const categories = ['/', '/en-cours', '/termines', '/stoppes'];
+
+        let request,data;
```
27136773: May declare variables when they are initialized (inside the loop).
No need to declare variables outside for re-use, there is no noticeable performance/memory gain.

src/web/mjs/connectors/FuryoSquad.mjs
```diff
@@ -0,0 +1,50 @@
...
+                title: chapter.title.trim()
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        const request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        const data = await this.fetchDOM(request, 'div.fs-reader-page source');
+
+        return data.map(page => page.src);
```
27136773: Is `page.src` always full qualified name (absolute) or should it be wrapped into `this.getAbsolutePath(page, this.url)`

27136773 (approved):
Ready for merge

src/web/mjs/connectors/FuryoSquad.mjs
XenonR: Uncertain, i missed that.

PR #2030, opened 2020-07-12
[readmanga] updated domain

27136773:


XenonR (approved):
how do they change on a monthly basis?

PR #2035, opened 2020-07-12
Futabanet MangaList Added & Latest Chapter Naming fixed

newaccforhc:
Previously, all latest chapters used to show "NEW" as their title instead of chapter number... So, fixed it.
Also, added manga list feature.

### DIFF
--- a/src/web/mjs/connectors/futabanet.mjs
+++ b/src/web/mjs/connectors/futabanet.mjs
@@ -20,14 +20,14 @@ export default class futabanet extends SpeedBinb {
     }
 
     async _getMangas() {
-        let request = new Request(this.url + '/list/monster/works', this.requestOptions);
+        let request = new Request(this.url + '/list/monster/serial', this.requestOptions);
         let pages = await this.fetchDOM(request, 'li.m-pager__last a');
         pages = Number( new URL(pages[0].href).searchParams.get('page') );
 
         let data;
         let mangas = [];
         for (let page = 0; page <= pages; page++) {
-            request = new Request(this.url + '/list/monster/works' + '?' + 'page' + '=' + page);
+            request = new Request(this.url + '/list/monster/serial' + '?' + 'page' + '=' + page);
             data = await this.fetchDOM(request, 'div.m-result-list__item a');
             mangas.push( ...data.map(element => {
                 return {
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/futabanet.mjs
```diff
@@ -20,20 +20,33 @@ export default class futabanet extends SpeedBinb {
     }
 
     async _getMangas() {
-        //https://futabanet.jp/list/monster/works?page=1
-        let msg = 'Manga list is not yet implimented for this website, please copy and paste the URL containing the chapters directly from your browser into HakuNeko.';
-        throw new Error(msg);
+        let request = new Request(this.url + '/list/monster/serial', this.requestOptions);
```
27136773: May use `new URL('/list/monster/serial', this.url)`

src/web/mjs/connectors/futabanet.mjs
```diff
@@ -20,20 +20,33 @@ export default class futabanet extends SpeedBinb {
...
+        let pages = await this.fetchDOM(request, 'li.m-pager__last a');
+        pages = Number( new URL(pages[0].href).searchParams.get('page') );
+
+        let data;
+        let mangas = [];
+        for (let page = 0; page <= pages; page++) {
+            request = new Request(this.url + '/list/monster/serial' + '?' + 'page' + '=' + page);
```
27136773: May be wrapped in `new URL()`
Also combine constant strings `'a?b='` instead of `'a' + '?' + 'b' + '='`

src/web/mjs/connectors/futabanet.mjs
```diff
@@ -20,20 +20,33 @@ export default class futabanet extends SpeedBinb {
...
         let request = new Request(new URL(manga.id, this.url), this.requestOptions);
         let data = await this.fetchDOM(request, 'section.detail-sec.detail-ex div.detail-ex__btn-item a');
         return data.map(element => {
-            let title = element.querySelector('span');
+            let title = element.querySelector('span:not(.new)');
             return {
                 id: this.getRootRelativeOrAbsoluteLink(element, request.url),
-                title: title.innerText.replace(manga.title, '').trim(),
-                language: ''
+                title: title.innerText.replace(/\(\d+.\d+(.*)\)$/, '').trim(),
```
27136773: No need for grouping if the result is not used, remember a `.` matches any character and `\.` matches a period literally, i'm not sure what was intended for the `.` between the `\d+`
`/\(\d+\.\d+.*\)$/`

27136773 (approved):
Ready for merge after feedback was received

src/web/mjs/connectors/futabanet.mjs
newaccforhc: Ohh.. yeah.. Simply using `.` is a wrong thing.. I need to use `\.`
I'll correct it later.

The text I want to remove: `(07.10更新！)`

PR #1886, opened 2020-06-14
MagicScroll and FullZoom ShortKeys

Sheepux:
Using spacebar, if the scroll down goes lower than the image, it will stop at the bottom of this image (if the scroll is not enougth to go lower than the image because of zoom then it will be a normal scroll). 
When you've reached the bottom of an image, pressing on spacebar again will scroll down to the top of the next image.

Also adding shorcuts (contributes to #1859 )
- FullImageWidth zoom * (zoom 100)
- DefaultImageWidth zoom / (zoom 75)

PR #2038, opened 2020-07-13
FEAT-MagicScrollv2 improvement and auto next chapter

Sheepux:
Code from PR #1886 reorganised to handle multiple images on screen (manwha with a lot of cuts). Use-cases described in image :
MagicScroll

Introduced the first elements for the AutoNextChapter feature : 
- When reaching the bottom page and trying to go to the next image will display (todo) a notice that you have to press the MagicScroll shortcut again
- Pressing it again within 4s will trigger the next chapter function

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/pages.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/pages.html
@@ -538,11 +538,11 @@
                 if(images[images.length-1].getBoundingClientRect().bottom-window.innerHeight < 0)
                 {
                     // Should we go to next chapter because we previouysly reached the end of page ?
-                    if (this.autoNextChapter) { this.requestChapterUp(); return; };
+                    if (this.autoNextChapter) { return this.requestChapterUp(); };
                     // Prepare for next chapter
                     this.autoNextChapter = true;
                     // Todo: Find a way to popup that you have to press spacebar again within 4s
-                    setInterval(function(){ this.autoNextChapter=false; }, 4000);
+                    setTimeout(function(){ this.autoNextChapter=false; }, 4000);
                     return;
                 }
                 // Lets stay on current page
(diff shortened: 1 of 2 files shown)


src/web/lib/hakuneko/frontend@classic-dark/pages.html
```diff
@@ -532,35 +534,55 @@
              */
             scrollMagic(defaultDistance) {
                 let images = this.$.container.querySelectorAll('.image');
+                // Are we at the end of the page
+                if(images[images.length-1].getBoundingClientRect().bottom-window.innerHeight < 0)
+                {
+                    // Should we go to next chapter because we previouysly reached the end of page ?
+                    if (this.autoNextChapter) { this.requestChapterUp(); return; };
```
27136773: Can safe a colon with `return this.requestChapterUp();`

src/web/lib/hakuneko/frontend@classic-dark/pages.html
```diff
@@ -532,35 +534,55 @@
...
+                    // Prepare for next chapter
+                    this.autoNextChapter = true;
+                    // Todo: Find a way to popup that you have to press spacebar again within 4s
+                    setInterval(function(){ this.autoNextChapter=false; }, 4000);
```
27136773: May prefer closure instead of function, not quite sure if this is correctly scoped in function context...
`setInterval(() => this.autoNextChapter = false, 4000)`

27136773: It seems many interval calls could be leaking, each of them ticking every 4 seconds and never getting cleared, is this intended, or should this be `setTimeout`?

Sheepux: Hmmm.... was probably drunk when writing this... setTimeout was indeed intended

27136773 (commented):
Only checked dark mode, same comments apply to light mode.
May cross check the `setInterval` behavior...

27136773 (approved):
Looks good

PR #2042, opened 2020-07-13
[mangafest] add website support

### LINKED ISSUE #317: add support for website [Mangafest]
https://mangafest.net/category/manga/

XenonR:
Closes #317


src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    async _getMangas() {
+        let request = new Request(new URL('/category/manga/', this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'nav.navigation.pagination div.nav-links a.page-numbers:not(.next)');
+        const pages = parseInt(data[data.length-1].pathname.match(/\d+\/$/));
+
+        let mangas = [];
+        for (let page = 1; page <= pages; page++) {
+            request = new Request(new URL('/category/manga/page/'+page+'/', this.url), this.requestOptions);
```
27136773: May use string interpolation
```javascript
`/category/manga/page/${page}/`
```

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+            data = await this.fetchDOM(request, 'div.post_image a');
+            mangas.push(...data.map(manga => {
+                return {
+                    id: this.getRootRelativeOrAbsoluteLink(manga.pathname, request.url),
```
27136773: May be safer to use `this.url` instead of `request.url`, in case both are different (cross reference), then `getPages` would not work as it assumes `this.url` as base

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+                };
+            }));
+        }
+
+        return mangas;
+    }
+
+    async _getChapters(manga) {
+        // website provides mangas as whole
+        return [
```
27136773: Can be simplified to `return [ manga ];`

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+            }
+        ];
+    }
+
+    async _getPages(manga) {
+        let request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'div#ngy2p');
+        const flickr_data = JSON.parse(data[0].getAttribute('data-nanogallery2-portable'));
+
+        let url = new URL('https://api.flickr.com/services/rest/?&method=flickr.photosets.getPhotos&api_key=2f0e634b471fdb47446abcb9c5afebdc&format=json');
```
27136773: For readability the parameters `method`, `api_key` and `format` might be set with `url.searchParams.set()` as well instead of appending them to the USVString

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+        url.searchParams.set('user_id', flickr_data.userID);
+        url.searchParams.set('photoset_id', flickr_data.photoset);
+        url.searchParams.set('page', 1); // Returns 500 pictures per page. Since these are hobbyist mangas I assume none will be over 500 pages.
+
+        request = new Request(url);
+        data = await this.fetchRegex(request, /^jsonFlickrApi\((.*)\)/g);
```
27136773: Is greedy intentional?
It may swallow any closing brackets into the group.

When dealing non-greedy (aka lazy) cases i prefer to define the ending character by _all characters but this_ regex (as an alternative to the lazy operator)
e.g. `/^jsonFlickrApi\(([^)]+)\)/g`

https://regex101.com/r/vFON21/2
https://regex101.com/r/vFON21/1

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+        data = JSON.parse(data);
+
+        if (data.stat != 'ok') throw new Error('Flickr API - Gallery query failed.');
+
+        let pages = [];
+        url = new URL('https://api.flickr.com/services/rest/?&method=flickr.photos.getSizes&format=json&api_key=2f0e634b471fdb47446abcb9c5afebdc');
```
27136773: Redundant string constant, may declare local variable when used more then once

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+        for (const page of data.photoset.photo) {
+            url.searchParams.set('photo_id', page.id);
+            request = new Request(url);
+            data = await this.fetchRegex(request, /^jsonFlickrApi\((.*)\)/g);
```
27136773: Same as mentioned above, may also consider to store regex in local variable when used more then once

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+            data = JSON.parse(data);
+
+            if (data.stat != 'ok') throw new Error('Flickr API - Image query failed.');
+
+            data = data.sizes.size.sort(this.sortSizes)[0];
```
27136773: When height is supposed to be a positive number, sort can be simplified to `sort((a, b) => b.height - a.height)`

src/web/mjs/connectors/Mangafest.mjs
```diff
@@ -0,0 +1,94 @@
...
+
+    async _getMangaFromURI(uri) {
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, 'head title');
+        let id = uri.pathname;
+        let title = data[0].text.split(' – ')[0].trim();
+        return new Manga(this, id, title);
+    }
+
+    sortSizes( first, second ) {
```
27136773: Maybe obsolete when inlining the a closure into sort as described above

27136773 (approved):
May revise if the Regex works as intended (greedy), then ready to merge

src/web/mjs/connectors/Mangafest.mjs
XenonR: ahh careless mistake

src/web/mjs/connectors/Mangafest.mjs
XenonR: Yes, but i added a `$` as last character in the meantime. Flickr always wraps the response in `jsonFlickrApi(...)` for whatever reason. So i now use `/^jsonFlickrApi\((.*)\)$/` which will prevent consuming the closing bracket.

src/web/mjs/connectors/Mangafest.mjs
XenonR: Nice! Really nice.

PR #2048, opened 2020-07-14
[novelcool] add website support

### LINKED ISSUE #528: Add website support [novelcool]
Hello,

Could you please add these two websites: https://es.novelcool.com/ and http://mangaina.com/.

Thank you,

XenonR:
Closes #528


src/web/mjs/connectors/NovelCool.mjs
```diff
@@ -0,0 +1,75 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class Mangafest extends Connector {
```
27136773, line 4: Classname does not match meta data

src/web/mjs/connectors/NovelCool.mjs
```diff
@@ -0,0 +1,75 @@
...
+    async _getMangas() {
+        let morePages = true;
+        let page = 1;
+
+        let mangas = [];
+        while (morePages) {
+            let request = new Request(new URL(`/category/index_${page}.html`, this.url), this.requestOptions);
+            let data = await fetch(request);
+
+            if ( data.url == `https://es.novelcool.com/category/index_${page++}.html` ) {
```
27136773, line 23: Prefer `===` to include type equality check as well

src/web/mjs/connectors/NovelCool.mjs
27136773: May create a template similar to WordPressMadara instead and make implementations for each support language

src/web/mjs/connectors/NovelCool.mjs
```diff
@@ -0,0 +1,75 @@
...
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        let page_links = await this.fetchDOM(request, 'select.sl-page');
+        page_links = [...page_links[0].querySelectorAll('option')].map(link => link.value);
+
+        let pages = [];
+        for( const link of page_links) {
```
27136773, line 59: `_getPages` should not scrape to all pages, consider using `this.createConnectorURI(...)` such as in [LELScan.mjs](https://github.com/manga-download/hakuneko/blob/4d0af6153b291747f862b9833b2ce09e6c91ed40/src/web/mjs/connectors/LELScan.mjs#L82), but with async/await instead of promises

src/web/mjs/connectors/NovelCool.mjs
```diff
@@ -0,0 +1,75 @@
...
+            pages.push( this.getRootRelativeOrAbsoluteLink( data[0].src, this.url) );
+        }
+
+        return pages;
+    }
+
+    async _getMangaFromURI(uri) {
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, 'div.bookinfo-info h1.bookinfo-title');
+        let id = uri.href;
```
27136773, line 71: Should this really be the absolute URL, this contradicts `_getMangas` where the ID is only the relative path of the URL

XenonR:
Will incorporate your suggestions. Closing PR in favor of template based approach.

PR #2067, opened 2020-07-16
[japscan] updated image detection

### LINKED ISSUE #2049: [Japscan] Small / unusual picture size get the download in error
Yes

**Describe the bug**
Some manga on japscan don't work each time i try (cancel at the end)

**To Reproduce**
1)Try to download "Air Gear" Volume 8 or Volume 9 for exemple or akane-chan overdrive volume 2
2)Wait for the download to complete
3)It'll crash at the end

 - OS : Windows10 64 bits
 - HakuNeko Stable and nightly versions, 32bits and 64bits

Seem to be related to the scan picture size.
All the manga with this bug have a very small / unusual picture size in the manga (https://www.japscan.co/lecture-en-ligne/air-gear/volume-8/109.html or https://www.japscan.co/lecture-en-ligne/akane-chan-overdrive/volume-2/4.html for exemple)

Screen Shot 014

Screen Shot 015

27136773:
Changeset: https://github.com/manga-download/hakuneko-plugins/commit/adf68a9ca916491bc3e42e123198e887dc027113
Resolves #2049


Sheepux (approved):
/LGTM
Unable to assert all possible cases.

satsukiyatoshi:
A simple way to avoid the canceling of the dl would be to skip small picture but not cancel the whole download ?
As realy small picture are always logos or stamp

27136773:
This would require to completely change the concept of how JapScan currently works and is way to much effort

satsukiyatoshi:
So i think a 64x64 is a safe value, i can't imagine a logo smaller than that ^^

satsukiyatoshi:
Found another limit witch make japscan crash in picture detection, some manga have an empty page, no picture at all, then the whole volume / chapter crash
for exemple : https://www.japscan.co/lecture-en-ligne/appelez-moi-hotesse/13/23.html

27136773:
This behavior is expected, if an image is missing the chapter download is marked as failed.
The image is missing on the website as well, so this seems to be an issue on JapScan website itself (i doubt they did it intentionally because their own users would recognise this as a broken page).

satsukiyatoshi:
ok, but is there any way to finish download with a warning / log / popup ? so user can get the chapter / volume then check the result / error log.

27136773:
No, if this would be allowed, then users would start complaining that images are missing in a downloaded chapter

satsukiyatoshi:
I get you point, but why to set it up as an hidden option, or command line argument ... so the option will be only visible in the manual with a disclaimer ?
As it's the only bug so far with japscan, to be able to ignore the missing picture would be great (i checked those volume online, there're complete, it's only a bad indexed empty page.
Thanks

Sheepux:
The feature request would be a new button in case of errors "retry and ignore any errors (warning content might be missing int he downloaded chapter)"

satsukiyatoshi:
Would be usefull too, to do it this way

PR #2072, opened 2020-07-17
[fenixscan] added website support

27136773:


XenonR (approved):
Good to go. The different domainname and website title are confusing, but it is what it is.

PR #2073, opened 2020-07-17
[netease] add origin header, evaluating authorization

### LINKED ISSUE #389: Add support for Bilibili Comics
https://manga.bilibili.com

Separated from #388

### LINKED ISSUE #2043: [bilibili / NetEase Comic] Connector not working
Yes

Yes

**Describe the bug**
unable to update manga list

**To Reproduce**
1. NetEase Comic
2. update not working

**Screenshots**

 - Version [master (yesterday)]
the issue was first noted by SiliuS#4699 on the discord but he didn't open a ticket
also given as the comment says on #389 shouldn't the name be changed to "Bilibili Manga" from "NetEase Comic" as "NetEase Comic" no longe exists

27136773:
Resolves #2043


src/web/mjs/connectors/NetEaseComic.mjs
```diff
@@ -6,6 +6,7 @@ export default class NetEaseComic extends Connector {
     constructor() {
         super();
         super.id = 'neteasecomic';
+        // TODO: rename website?
```
XenonR: I would say yes if existing bookmarks don't break.

XenonR (approved):
Also looking good.

src/web/mjs/connectors/NetEaseComic.mjs
27136773: Has been renamed. Only ID is critical because it will break all existing bookmarks, so it should not be changed.

27136773:
I cannot purchase a chapter due to Alipay only accept payments from China.
Therefore it cannot be verified if the chapter filtering correctly includes purchased chapters or not (after manual user login).

PR #2082, opened 2020-07-18
[heromanhua] added website support

27136773:


Sheepux (approved):
/LGTM

However, are we really adding a 12 mangas site?

27136773:
The simplicity of implementation is tempting and i think its a scanlation source

Issue #1757, opened 2020-05-21
[Site Request] GoComics


maxtron95:
**Name of the website**
GoComics

**Website urls (examples below)**
- Site: https://www.gocomics.com/
- Manga List: https://www.gocomics.com/comics/a-to-z
- Manga example: https://www.gocomics.com/garfield
- Chapter online viewer example: https://www.gocomics.com/garfield/2020/05/20

**Languages**
English

**Additional details**
I viewed Page Info in Firefox and I found that the comic image files are being pulled from `https://assets.amuniversal.com/...`

MikeZeDev:
There is no concept of chapters on this website and no proper chapter list for a comic therefore i fail to see how it can be added.

PR #2087, opened 2020-07-18
[gocomics] added website support

speedycatfish:
Fixes issue #1757


src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+            throttle: {
+                label: 'Throttle Requests [ms]',
+                description: 'Enter the timespan in [ms] to delay consecuitive HTTP requests.\nThe website may ban your IP for too many consecutive requests.',
+                input: 'numeric',
+                min: 0,
+                max: 5000,
+                value: 500
+            }
+        };
+        this.lastChapter = '';
```
27136773, line 22: This variable can be declared locally, no need to make it a field of the class

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
+                title: title.innerText.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let chapter = manga.id;
+        this.lastChapter = manga.id;
```
27136773, line 54: Use `let lastChapter = manga.id`

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+        let request = new Request(this.url + manga.id, this.requestOptions);
```
27136773, line 55: May use `new URL(manga.id, this.url)` instead of `this.url + manga.id`

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+        let startData = await this.fetchDOM(request, 'nav.gc-calendar-nav a.fa-backward');
+        let startDate = this._getDate(startData[0].pathname);
+        let endDate = this._getDate(manga.id);
+        let chapterList = [];
+        let title = manga.id.slice(1, manga.id.search(/\/\d\d\d\d\/\d?\d\/\d?\d$/));
```
27136773, line 60: May simplify regex, use regex101 online editor for testing and verification
e.g. `/.\d{4}.\d{1,2}.\d{1,2}$/`

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+        let chapterNum = 1;
```
27136773, line 61: May optionally be placed in the loop to express its relation/scope
`for(let chapterNum = 1, let year = startDate.year; year <= endDate; ...)`

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+        return {
+            year: parseInt(dateString.match(/\d\d\d\d/)[0]),
+            month: parseInt(dateString.match(/\/\d?\d\//)[0].slice(1, -1))
+        };
+    }
+
+    async _getPages(chapter) {
+        this.getPagesCount++;
+        let startPageCount = this.getPagesCount;
+        let title = chapter.manga.id.slice(1, chapter.manga.id.search(/\/\d\d\d\d\/\d?\d\/\d?\d$/));
```
27136773, line 94: May simplify regex as mentioned above

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+    _getDate(path) {
+        let dateString = path.match(/\d\d\d\d\/\d?\d\/\d?\d$/)[0];
+        return {
+            year: parseInt(dateString.match(/\d\d\d\d/)[0]),
+            month: parseInt(dateString.match(/\/\d?\d\//)[0].slice(1, -1))
+        };
+    }
+
+    async _getPages(chapter) {
+        this.getPagesCount++;
```
27136773, line 92: Using a field for local logic will probably break the logic as soon as this function is called more then ones at the same time (e.g. a user is clicking through the chapter previews without waiting)
Make this a local variable and remove the field from the class
`let getPagesCount = 1`

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+
+    _getDate(path) {
+        let dateString = path.match(/\d\d\d\d\/\d?\d\/\d?\d$/)[0];
+        return {
+            year: parseInt(dateString.match(/\d\d\d\d/)[0]),
+            month: parseInt(dateString.match(/\/\d?\d\//)[0].slice(1, -1))
+        };
+    }
+
+    async _getPages(chapter) {
```
27136773, line 91: This method seems quite complex and i have trouble to understand it.
This is a sign that the design may be improved fro better readability.

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+        this.getPagesCount++;
+        let startPageCount = this.getPagesCount;
+        let title = chapter.manga.id.slice(1, chapter.manga.id.search(/\/\d\d\d\d\/\d?\d\/\d?\d$/));
+        let json = await (await fetch(this.url + '/calendar/' + title + chapter.id)).json();
+        let pages = [];
+        for (let date of json) {
+            await this.wait(this.config.throttle.value);
+            await this._waitFor(() => startPageCount == this.getPagesCount);
+            let request = new Request(this.url + '/' + title + '/' + date, this.requestOptions);
+            let imageData = await this.fetchDOM(request, 'div.comic__container div a picture source');
```
27136773, line 101: Looks like all HTML pages containing an image are requested and parsed when a chapter is requested, which is a quite long operation for a requesting a simple list of links.

There is a mechanic called `Connector URIs` to solve such cases
The basic idea is that `_getPages()` return a list of links, where each link must return an image.
It doesn't matter if the link actually is an image or not, it might be an HTML page containing the image, or a link to an encrypted binary blob, or ...
So when a link is not an image it is possible to return a protocol handler link instead which works like a callback. Whenever something requests the image, the `_handleConnectorURI(param)` method is called which must return the image.

E-Hentai is one of the many examples, for further information ask about this on discord
https://github.com/manga-download/hakuneko/blob/852e6c60321514d30a343c6014c5978e645d51e7/src/web/mjs/connectors/EHentai.mjs#L35

src/web/mjs/connectors/GoComics.mjs
```diff
@@ -0,0 +1,116 @@
...
+
+    async _getPages(chapter) {
+        this.getPagesCount++;
+        let startPageCount = this.getPagesCount;
+        let title = chapter.manga.id.slice(1, chapter.manga.id.search(/\/\d\d\d\d\/\d?\d\/\d?\d$/));
+        let json = await (await fetch(this.url + '/calendar/' + title + chapter.id)).json();
+        let pages = [];
+        for (let date of json) {
+            await this.wait(this.config.throttle.value);
+            await this._waitFor(() => startPageCount == this.getPagesCount);
```
27136773, line 99: This condition seems always to be true and therefore is redundant
See L98 `let startPageCount = this.getPagesCount`

Also if i understand correctly it says something like
```javascript
if(startPageCount !== this.getPagesCount) {
    await this.wait(400);
}
```
In that case i would just write it that way instead of nesting closure functions

27136773 (changes requested):
Suggest further discussion about the adaption of the `Connector URI` mechanism before merge

27136773:
Some additional notes:

Current implementation seems to work with single page images:
https://www.gocomics.com/angry-birds/2020/05/22

But does it support long strips as well:
https://www.gocomics.com/comics/lists/1660283/calvin-and-hobbes-valentines-day

Also it seems that there is no concept of "chapters" on the website, so everyday a new image is added, the user has to re-download the full chapter to get the new image

27136773:
Is this PR still active?

27136773:
Will close for now, maybe re-opened when work is continued

PR #2086, opened 2020-07-18
[manganel] cross-link support for manganel and its affiliates

27136773:
Resolves #2074
Resolves #2103 

Drawbacks:
- Individual CSS queries are combined and defined in `MangaNel.mjs`
- Using absolute URLs as IDs will render bookmarks useless when domain changes

----

# Verification

The following sections provide some use cases and their expected results

## MangaNel

<details><summary>Show Test Cases</summary>

### MangaList

The manga list should load correctly, the following mangas in the list should successfully provide a chapter list
- [x] Tower of God
- [x] Solo Leveling
- [x] The Ghostly Doctor

### Copy & Paste

When copy & paste the following URLs in HakuNeko, they should successfully load a manga provided by `MangaNel` and all chapters of the manga should load the preview images correctly
- [x] https://manganelo.com/manga/tower_of_god_manga (Tower of God | old link)
- [x] https://chap.manganelo.com/manga-aa89094 (Tower of God | new link)
- [x] https://chap.manganelo.com/manga-dr117685 (Solo Leveling)
- [x] https://m.manganelo.com/manga-cs116938 (The Ghostly Doctor)

</details>

## MangaBat

<details><summary>Show Test Cases</summary>

### MangaList

The manga list should load correctly, the following mangas in the list should successfully provide a chapter list
- [x] Tower of God
- [x] Solo Leveling
- [x] The Ghostly Doctor

### Copy & Paste

When copy & paste the following URLs in HakuNeko, they should successfully load a manga provided by `MangaBat` and all chapters of the manga should load the preview images correctly
- [x] https://mangabat.com/manga/tower_of_god_manga (Tower of God | old link)
- [x] https://read.mangabat.com/read-ii14105 (Tower of God | new link)
- [x] https://read.mangabat.com/read-lz42696 (Solo Leveling)
- [x] https://m.mangabat.com/read-hx385616 (The Ghostly Doctor)

</details>

## MangaIro

<details><summary>Show Test Cases</summary>

### MangaList

The manga list should load correctly, the following mangas in the list should successfully provide a chapter list
- [x] Tower of God (Cross-Domain-Link | MangaNel)
- [x] Solo Leveling (Cross-Domain-Link | MangaNel)
- [x] The Ghostly Doctor

### Copy & Paste

When copy & paste the following URLs in HakuNeko, they should successfully load a manga provided by `MangaIro` and all chapters of the manga should load the preview images correctly
[...]

PR #2094, opened 2020-07-20
[shonenjumpplus][coreview] read json data from html page

hyww:
shonenjumpplus.com returns 404 on `${chapter id}.json` requests. read json data from html pages instead.

Sorry I am not sure if this works for all coreview based sites. Maybe this fix should be applied only to shonenjumpplus.

### DIFF
--- a/src/web/mjs/connectors/templates/CoreView.mjs
+++ b/src/web/mjs/connectors/templates/CoreView.mjs
@@ -17,6 +17,7 @@ export default class CoreView extends Connector {
         this.queryChapters = 'feed entry';
 
         this.queryPages = 'source.page-image[data-src]';
+        this.queryEpisodeJSON = '#episode-json';
     }
 
     async _getMangas() {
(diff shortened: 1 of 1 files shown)


27136773 (changes requested):
Looks like a good addition, unfortunately i cannot approve this without a confirmation of manual testing for each affected connector.
It would be sad if this fixes ShonenJump+, but breaks all other 8 websites.
The second approach would be the mentioned overwrite of `_getPages()` in `ShonenJumpPlus` only.

Sample Checklist for manual Testing (See also https://github.com/manga-download/hakuneko/pull/2086#issue-451976901):

- [ ] ShonenJump+
Sample: https://shonenjumpplus.com/episode/13933686331681119000
- [ ] TonariNoYoungJump
Sample: https://tonarinoyj.jp/episode/13933686331675753623

...

----
**CoreView based websites**

image

27136773:
Probably related #2096

newaccforhc:
@ronny1982 I've tested this with each and every coreview connector supported by HakuNeko (The list in your Screenshot) and it works... You can approve it.

27136773 (approved):
Functionality confirmed by @newaccforhc

PR #2120, opened 2020-07-30
[webnovel] prefer high resolution images

### LINKED ISSUE #2118: [Feature] Add width parameter to Webnovel Comics to fetch HQ images
Webnovel by default serves the images with width 800px. however current comics have max width of 900px and some old mangas released around last december go upto 1080px quality.
In getContent api call, adding the width to searchParams in  _getPageList method will let you fetch highest available. set this value to something like 2000.

27136773:
Resolves #2118


Sheepux (approved):
Looks Genius To Me

PR #2162, opened 2020-08-11
[funlist] added website support

27136773:
References fmd-project-team/FMD#1106


Sheepux (approved):
5 mangas website ? (but well ... it's madara)
I guess you don't want to be "behind" FMD 😄 
/LGTM

27136773:
Re-checking if the website has less ads and provides direct download links to think about merging.
Unfortunately website is down right now => postponed decision.

Yes WP-Madara is tempting because it is so easy, but there are websites in FMD which are not supported in HakuNeko such as https://guya.moe

27136773:
uBlock shows 11~12 blocked ad items (layers/scripts)
there are no visible direct download links
=> think it is OK to add website to HakuNeko

Issue #2117, opened 2020-07-30
[Mangas.in] Continuously changing to defeat HakuNeko


rausina:
**Did you read [the troubleshooting guide](https://github.com/manga-download/hakuneko/wiki/Troubleshooting)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes

**Describe the bug**
all mangas download 1kb .cbz files

**To Reproduce**
1. Mangas.in
2. All mangas
3. All
4. Copy any relevant error message that appears in a popup or (advanced) within the console (press F12)

**Screenshots**
![E72RaUA](https://user-images.githubusercontent.com/65744036/88867685-31fe9680-d1dc-11ea-9d15-a956303f4e16.png)
Chapter download after the problem and before 
![image](https://user-images.githubusercontent.com/65744036/88867862-a89b9400-d1dc-11ea-9424-0d9d11c3e3a8.png)

**HakuNeko (please complete the following information):**
 - Version [e.g. 6.17]

27136773:
Please provide a sample manga / chapter for reproduction

rausina:
The file is empty
[⟶ #116.2                      Capítulo 116.2.zip](https://github.com/manga-download/hakuneko/files/5004459/116.2.Capitulo.116.2.zip)

27136773:
Reproducible with: `29 to JK`, `#29 Polvo de Estrellas y Chocolate`

Next time please provide a sample manga & chapter, the developers don't have time to search through thousands of mangas and chapters looking for one that didn't work...

rausina:
Sorry for that, it is easier for me complain than to give solutions. Thank for all work you are doing.
Now it just doesn't load the mangas that had those problems
![image](https://user-images.githubusercontent.com/65744036/89096475-ffec6080-d3a4-11ea-8cc0-1c4584e4d944.png)
![image](https://user-images.githubusercontent.com/65744036/89096502-21e5e300-d3a5-11ea-88d7-977e822ad730.png)
![image](https://user-images.githubusercontent.com/65744036/89096619-07f8d000-d3a6-11ea-9596-68375c7737fd.png)
[cache-1596266616039.log](https://github.com/manga-download/hakuneko/files/5009976/cache-1596266616039.log)

27136773:
It seems MangAs entered a battle with HakuNeko, they are tracking changes in this git repository and updating their website accordingly.

Some websites are always trying to fight against HakuNeko:

2017 - KissManga
[...]

PR #2169, opened 2020-08-12
[mangas] fixed chapter list

27136773:
Review: https://github.com/manga-download/hakuneko-plugins/commit/7cbf22abd9fe83f8b52773247a041db686829db6
References #2117


Sheepux (approved):
Thanks for the cross ref.
/LGTM

The code is actually fun to read to see how you bypass the tricks 😃

PR #2208, opened 2020-08-21
[mangahere] use text based manga list

### LINKED ISSUE #2203: [Mangahere] Connector suggestion for improvement
Yes

**Describe the issue**
Currently, HakuNeko uses https://www.mangahere.cc/directory/*numbers*.htm to scan Mangahere for about 143 pages which takes some time to go through. Can the connector just use the page https://www.mangahere.cc/mangalist/ to get all listings all at once saving time when refreshing MangaHere?

nightly-20200705.1 with master@57a756

27136773:
[mangafox] migrated to new API

### DIFF
--- a/src/web/mjs/connectors/MangaFox.mjs
+++ b/src/web/mjs/connectors/MangaFox.mjs
@@ -16,15 +16,14 @@ export default class MangaFox extends Connector {
         this.scriptSource = 'chapterfun.ashx';
     }
 
-    /**
-     *
-     */
     get script() {
         return `
             new Promise(resolve => {
                 if(isbarchpater) {
+                    // webtoon (all images on a single page)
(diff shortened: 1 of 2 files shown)


Sheepux (approved):
/LGTM list if refreshed in lesss than 10s
No issue with previous bookmarked manga on connector

Found some issues with mangahere returning HTTP 408 but it doesn't seem related to this change (as i found it in the old connector as well).
Will investigate this one in another PR as the error is "blank" and therefore confusing for the user when such case occurs, will try to intercept the issue and ask the user to retry (or do it for him? could be some sort of throttling?).

27136773:
May add the 3rd parameter for `fetchDOM(request, query, 3)` which is retry count and could be triggered on 4xx and 5xx HTTP error codes

PR #2214, opened 2020-08-22
[MangaHere][MangaFox] Remove ad on last page

Sheepux:
Rework #1543
Detects if the last image seems to be a global included ad by its format Height=563 lWidth==1000
Edge case, the last image is actually this size (low probability as the site is almost always including this ad).

Previous PR found a pattern in the image names but i couldn't find a good solution to detect it.

### DIFF
--- a/src/web/mjs/connectors/MangaFox.mjs
+++ b/src/web/mjs/connectors/MangaFox.mjs
@@ -101,7 +101,21 @@ export default class MangaFox extends Connector {
     async _getPagesDesktop(chapter) {
         let uri = new URL(chapter.id, this.url);
         let request = new Request(uri, this.requestOptions);
-        return Engine.Request.fetchUI(request, this.script);
+
+        var pages = await Engine.Request.fetchUI(request, this.script);
+        //Get Meta informations from last Image
+        let lastImage = await new Promise(resolve => {
+            const img = new Image();
+            img.onload = function() {
+                resolve(img);
+            };
+            img.src = pages[pages.length-1];
+        });
+        //Check if the last image seems to be the usual ad
+        if(lastImage.naturalHeight==563 && lastImage.naturalWidth==1000) {
+            pages.pop();
+        }
+        return pages;
     }
 
     async _getPageListMobile(chapter) {
--- a/src/web/mjs/connectors/MangaFox.mjs
+++ b/src/web/mjs/connectors/MangaFox.mjs
@@ -109,7 +109,7 @@ export default class MangaFox extends Connector {
             img.onload = function() {
                 resolve(img);
             };
-            img.src = pages[pages.length-1];
+            img.src = pages.slice(-1);
         });
         //Check if the last image seems to be the usual ad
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/MangaFox.mjs
```diff
@@ -101,7 +101,21 @@ export default class MangaFox extends Connector {
...
+        //Get Meta informations from last Image
+        let lastImage = await new Promise(resolve => {
+            const img = new Image();
+            img.onload = function() {
+                resolve(img);
+            };
+            img.src = pages[pages.length-1];
+        });
+        //Check if the last image seems to be the usual ad
+        if(lastImage.naturalHeight==563 && lastImage.naturalWidth==1000) {
```
27136773: In case natural values are number types,  a strong type comparison with `===` may be performed (instead of non-typed comparison `==`)
`if(lastImage.naturalHeight === 563 && lastImage.naturalWidth === 1000)`

src/web/mjs/connectors/MangaFox.mjs
```diff
@@ -101,7 +101,21 @@ export default class MangaFox extends Connector {
     async _getPagesDesktop(chapter) {
         let uri = new URL(chapter.id, this.url);
         let request = new Request(uri, this.requestOptions);
-        return Engine.Request.fetchUI(request, this.script);
+
+        var pages = await Engine.Request.fetchUI(request, this.script);
```
27136773: Usage of `let` may be preferred over `var` (since `var` still implies this global feeling)

27136773 (approved):
Looks good

src/web/mjs/connectors/MangaFox.mjs
```diff
@@ -101,7 +101,21 @@ export default class MangaFox extends Connector {
...
+        //Get Meta informations from last Image
+        let lastImage = await new Promise(resolve => {
+            const img = new Image();
+            img.onload = function() {
+                resolve(img);
+            };
+            img.src = pages[pages.length-1];
```
27136773: May use `pages.slice(-1)`, it leaves the original array untouched (in contrast to `splice` or `pop`)

Sheepux: Well, it was the purpose to modify the original array (to remove 1 element)

27136773: But you don't want to remove the image before it was loaded, you want to remove it later
`pages.slice(-1)` is equivalent to `pages[pages.length-1]` it provides the last element without modifying the `pages` array

Sheepux: That's why i have an await on the promise, I'm sure the image has been loaded :)

Sheepux: Oh, i understand "where you wanted to slice ! I was thinking you wanted to remove the pop

27136773: Seems GitHub comments are not as good as in Azure DevOps, where the line(s) with comment are clearly highlighted

PR #2222, opened 2020-08-25
[comicmedu] add support

ghost:


27136773 (changes requested):
Please make one pull request per topic (one for comicmedu, another for newtype)
Please fix lint errors first (`npm run lint` to check locally)

Congratulations for PR 2222

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+export default class COMICMeDu extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'comicmedu';
+        super.label = 'COMIC MeDu (こみっくめづ)';
+        this.tags = [ 'manga', 'japanese' ];
+        this.url = 'http://comic-medu.com';
+
+        this.path = [ '/wk' ];
```
27136773, line 13: May just use `path = '/wk'` instead of an array with a single element

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+        return mangas.map(series => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(series.querySelector(this.queryMangaLink), this.url),
+                title: series.querySelector(this.queryMangaTitle).textContent
+            };
+        });
+    }
+
+    async _getMangas() {
+        let uri = new URL(this.url + this.path[0]);
```
27136773, line 43: May use 2 argument constructor `new URL(this.path[0], this.url)`

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+        let request = new Request(uri, this.requestOptions);
+        let seriesPages = await this.fetchDOM(request, this.queryMangaListCount);
```
27136773, line 45: May be better to extract the page numbers from the last list element, it is very unlikely that there will be 500 list items shown when there are 500 pages.

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+        let totalPages = seriesPages.length;
+        let mangas = await this.fetchDOM(request, this.queryManga);
```
27136773, line 47: May be removed to prevent duplicate code (same thing is already done in the loop)

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+        let mangaList = await this._mapMangas(mangas);
```
27136773, line 48: May be removed to prevent duplicate code (same thing is already done in the loop)
`let mangaList = [];`

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+        for(let page = 2; page <= totalPages; page++) {
```
27136773, line 49: May start from page `1` to avoid processing the first page separately outside of the loop (avoid code duplication)

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+            uri.searchParams.set('page', page);
+            request = new Request(uri, this.requestOptions);
+            mangas = await this._mapMangas(await this.fetchDOM(request, this.queryManga));
+            mangaList.push(...mangas);
+        }
+        return mangaList;
+    }
+
+    async _getChapters(manga) {
+        let uri = new URL(this.url + manga.id);
```
27136773, line 59: May use 2 argument constructor new URL(manga.id, this.url)

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryChapters);
+        if(data.length == 0) { // series is a short story; only one chapter
```
27136773, line 62: May use type safe comparison operator `===`

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+            data = await this.fetchDOM(request, this.queryShortChapter);
+            data[0].textContent = manga.title;
+        }
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
```
27136773, line 68: May use `this.url` instead of `request.url`, as this is re-composed in `_getPages` also with `this.url`

src/web/mjs/connectors/COMICMeDu.mjs
```diff
@@ -0,0 +1,80 @@
...
+                title: element.textContent
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let uri = new URL(this.url + chapter.id);
```
27136773, line 75: May use 2 argument constructor new URL(chapter.id, this.url)

27136773 (approved):
May use their twitter icon instead:
https://pbs.twimg.com/profile_images/936228460644597761/YEez7vHj_reasonably_small.jpg

No problems, verified with random manga

PR #2225, opened 2020-08-25
[comic newtype] fullsize images

ghost:


src/web/mjs/connectors/NewType.mjs
```diff
@@ -77,6 +77,10 @@ export default class NewType extends Connector {
         let data = await this.fetchDOM(request, 'div#viewerContainer');
         let link = this.getAbsolutePath(data[0].dataset.url, request.url);
         data = await this.fetchJSON(new Request(link, this.requestOptions));
-        return data.map(image => this.getAbsolutePath(image, request.url));
+        return data.map(image => {
+            if (image.includes("/h1200"))
```
27136773, line 81: May just strip of the `/h1200` appendix, so the whole `if` condition is no longer required, e.g.
```javascript
image = image.split('/h1200').shift();
// or
image = image.replace(/\/h1200.*/, '');
```

27136773 (approved):
No problems

PR #2274, opened 2020-09-12
FIX: Disable the right panel when reader disabled

Sheepux:
Adding the start page introduced a regression in the interface: the right panel isn't removed when "Enable Reader" option was disabled and therefore creating a UI behavior issue.
The start page will only be displayed when "Enable Reader" is enabled (default hakuneko configuration).  Fits with the purpose of the start page: help new users.


27136773:
There is a also a scrolling issue with the background which might be addressed as well

Screen Shot 2020-09-13 at 1 15 08 AM

27136773 (approved):
Looks good

Sheepux:
Background scrolling fixed

PR #2300, opened 2020-09-15
warning message about slow connectors

### LINKED ISSUE #2071: [Feature] warning users about slow downloads/list updates on sites with request throttle
not really a problem but pushing people away from the more problematic connectors (with throttles) would probably ease the amount of complaints dew to connectors being slow and captcha issues.

upon selecting a connector with a throttle it would popup something like "this connector has a request throttle that may result in slow download speeds, disabling the throttle may result in a ip ban from the connector" a "do not show again" box would be nice too

 - Your current HakuNeko Version [e.g. 6.17]

Sheepux:
resolves #2071 
not specific to slow connectors but a global warning that users can see while waiting.
Seems good enougth to me as a quick solution.


27136773 (approved):
Looks good

PR #2301, opened 2020-09-15
fix start page theme

Sheepux:
Found a color from a previous test was hiding and i didn't see it while testing. Light theme is no longer dark.
Adjusted a few things so that the defaut page doesn't enable the scrollbar.


27136773 (approved):
Looks good

PR #2304, opened 2020-09-16
[WordPressMadaraNovel] Implement _getMangaFromURI

### LINKED ISSUE #2290: Manga Name Error while Using Clipboard method[Bug]
Yes

**Describe the bug**
When i copied the manga url and the clicked the clipboard icon it loaded the manga which i wanted but there was a bug and the manga name didnt show. I tried multiple time but the name didnt show properly. I restarted multiple times too and usend tother websites to test. unfortunately the name didnt show.

**To Reproduce**
1. Go to 'https://wuxiaworld.site/novel/kill-the-hero-comic/'
2. Copy the url : 'https://wuxiaworld.site/novel/kill-the-hero-comic/'
3. Open HakuNeko and Click Clipboard Icon

**Expected behavior**
I expected to see the manga name!

**Screenshots**
error name png

 - OS: [Windows10]
 - HakuNeko Version [6.17] Stable Latest

**Additional context**

XenonR:
Closes #2290

### DIFF
--- a/src/web/mjs/connectors/templates/WordPressMadara.mjs
+++ b/src/web/mjs/connectors/templates/WordPressMadara.mjs
@@ -1,4 +1,5 @@
 import Connector from '../../engine/Connector.mjs';
+import Manga from '../../engine/Manga.mjs';
 
 export default class WordPressMadara extends Connector {
 
@@ -11,6 +12,7 @@ export default class WordPressMadara extends Connector {
         this.queryMangas = 'div.post-title h3 a, div.post-title h5 a';
         this.queryChapters = 'li.wp-manga-chapter > a';
         this.queryPages = 'div.page-break source';
+        this.queryTitleForURI = 'head meta[property="og:title"]';
     }
 
     _createMangaRequest(page) {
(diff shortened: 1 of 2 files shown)


27136773 (commented):
Would it make more sense to move the method into the base class `WordPressMadara`, so all Madara based connectors could use this and not just Madara Novel?

Many Madara (Novel) based websites may not work correctly after this update, because the CSS query may be different for each website, but i don't see this as show stopper, as fetching the manga list is usually faster for these websites then using copy & paste.

XenonR:
Idk tbh. How similar are the novel vs. non-novel sites? Relying on your experience here. If you say it makes more sense to integrate it into `WordPressMadara` I can do so.

27136773:
Seems most Madara and MadaraNovel websites provide title by the last `og:title` meta tag, so i would go with this
`[...document.querySelectorAll('head meta[property="og:title"]')].pop().content`

XenonR:
> 
> 
> Seems most Madara and MadaraNovel websites provide title by the last `og:title` meta tag, so i would go with this
> `[...document.querySelectorAll('head meta[property="og:title"]')].pop().content`

Works too and is certainly more universal. Some sites do add the site name after the title, but that should be fine as long as it is consistent (per conenctor).

27136773 (approved):
Looks good

PR #2332, opened 2020-09-20
[Hato] updated domain

dodieboy:


27136773 (changes requested):
The website incl. the icon should be removed instead of being renamed. It would be too confusing having two connectors for KissLove. If users have bookmarks from HATO, they will be lost or need to be manually moved to the existing KissLove connector.

dodieboy:
my bad, never see there is kisslove connector alr

PR #2347, opened 2020-09-22
New theme

devem-software:
Add new theme
Remove fontawesome icons
Add material design icons


devem-software:
Fix Settings.mjs -> add support for modern-theme

src/web/mjs/engine/Settings.mjs
```diff
@@ -57,9 +57,10 @@ export default class Settings extends EventTarget {
             input: types.select,
             options: [
                 { value: 'frontend@classic-light', name: 'Classic (Light)' },
-                { value: 'frontend@classic-dark', name: 'Ken\'s Daedal Dark' }
+                { value: 'frontend@classic-dark', name: 'Ken\'s Daedal Dark' },
+                { value: 'frontend@modern', name: 'Modern <devemloop>' }
             ],
-            value: 'frontend@classic-light'
+            value: 'frontend@modern'
```
Sheepux: I think we should stay on classic-light by default -for a while- so that more "early adopters" can test it thouroughly.

src/web/css/materialdesignicons/.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
```
27136773: Do not check in user configuration files => delete this file or add entry to `.gitignore`

27136773:
Remove `materialdesignicons/css/materialdesignicons.css` and use the included
minified version `materialdesignicons/css/materialdesignicons.min.css` instead.

Remove `materialdesignicons/css/materialdesignicons.css.map` and use the included
minified version `materialdesignicons/css/materialdesignicons.min.css.map` instead.

27136773:
There are way too many changes in `src/web/index.html`, this file should be kept as is.
The update notification should not be styled as this is generic for all frontends.

src/web/index.html
```diff
@@ -14,161 +18,233 @@
     <script src="./js/ass.min.js" defer></script>
     <script src="./js/sql.min.js" defer></script>
     <script src="./js/exif-js.min.js" defer></script>
+    <link rel="stylesheet" href="./css/materialdesignicons/mdi.css" />
```
27136773: Use this directly in the theme of the new frontend, the index should not apply any theme based styles

27136773:
The directory `src/web/lib/hakuneko/frontend@modern/css/materialdesignicons/fonts/` seems to contain duplicates of all fonts that are already in `src/web/css/materialdesignicons/css/`, is this necessary or can one of them be removed?

27136773 (commented):
Briefly checked all the files, i think the `index.html` should only contain changed that are technically necessary to load the themes. It should not add any style or theme itself.

devem-software:
ready, everything is fixed

src/web/index.html
```diff
@@ -1,250 +1,175 @@
...
-    />
-
+    <link rel="icon" href="./img/logo.png">
+    <meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
     <script src="./js/protobufjs.min.js"></script>
     <script src="./js/crypto-js.min.js"></script>
     <script src="./js/oauth-1.0a.min.js"></script>
     <script src="./js/jszip.min.js" defer></script>
-    <script src="./js/pdfkit.standalone.js" defer></script>
+    <script src="./js/jspdf.min.js" defer></script>
```
27136773: `jspdf` was replaced by `pdfkit`, so it should not be reverted

src/web/index.html
```diff
@@ -1,250 +1,175 @@
...
     <script src="./js/hls.light.min.js" defer></script>
     <script src="./js/ass.min.js" defer></script>
     <script src="./js/sql.min.js" defer></script>
-    <script src="./js/exif-js.min.js" defer></script>
```
27136773: `exif` is required by one of the connectors to extract information from an image required for scraping and therefore should not be removed

src/web/index.html
```diff
@@ -148,6 +148,7 @@
         }
 
         #default_menubar {
+            width: 100%;
```
27136773, line 152: `menubar` is a block element (div) and therefore setting a width of 100% is not explicitly required

src/web/index.html
```diff
@@ -111,11 +110,12 @@
     </script>
     <style>
         html, body {
+            display:flex;
```
27136773, line 114: Is there any benefit in using flex?

27136773:
Please be aware that local storage might be cleared when the HakuNeko electron client is updated.
Since it is only used for `frontend/modern` related data, there is probably no harm even when the stored data is lost.

That said, it is not expected that the HakuNeko electron client will ever be updated again (in favor of upcoming HakuNeko NW)

27136773 (commented):
Can be merged after fixing the included scripts in `index.html`

src/web/index.html
devem-software: All the modern style uses flexbox, for this reason it is required that the body and the html have this tag for correct operation, the flex display does not affect the other styles

src/web/index.html
devem-software: the setting is done because the body has the flex property in its css

src/web/mjs/connectors/VerComicsPorno.mjs
```diff
@@ -1,55 +1,66 @@
-import Connector from '../engine/Connector.mjs';
```
27136773: This file is not related to this PR and should be updated in a separate PR

src/web/mjs/connectors/TvYManga.mjs
```diff
@@ -33,6 +33,7 @@ export default class TvYManga extends Connector {
     async _getChapters(manga) {
```
27136773: This file is not related to this PR and should be updated in a separate PR

devem-software:
### Added
- The i18n file is added to the engine for the translation into multiple languages
- The lang folder is added with the dictionaries for future translations, at the moment it supports English and Spanish
- Complete translation into Spanish in the components
- In the customization menu options to modify the language are added

### Fixed
- Some fixes in the style sheets

devem-software:
I need feedback for this PR, thanks

Sheepux:
I'm starting to get enougth time. I'll be testing this one for the next days. So i will probably need around 2 weeks (testing & some code review). 
Having Ronny take a second look at the code will also help.
Make sure to mark the discussions as resolved when you've done some changes about it

27136773:
Localization is global, but the content is only usable for the new fronend, i suggest a mechanism to distinguish localization and made it only for the new frontend.

This may not be necessary, since there is no other localization, but it would be better to prevent collisions in the future, e.g. when another frontend or the error messages from the engine may be translated as well...

27136773:
I don't see the themed integration in the startup UI anymore, which was a major complain so if the rest of the frontend is working as expected it could be merged

Sheepux:
Windows 10 x64 20H2

Unable to get the theme to display.
Steps done to reproduce:
* npm install
* npm start
* HN shows up with my current theme
* Change theme to modern in menu
* Close HN
* npm start
* white screen and nothing happens

Console:
> Loading Frontend: frontend@modern

 and then error messages in console:

> GET hakuneko://cache/lib/hakuneko/frontend@modern/polymer/polymer-element.html net::ERR_NOT_IMPLEMENTED

and

> Uncaught TypeError: Cannot read property 'general' of undefined
>     at I18n.translate (I18n.mjs:33)
>     at HTMLElement.i18n (script.js:178)
>     at runMethodEffect (property-effects.html:818)
>     at Function._evaluateBinding (property-effects.html:2738)
>     at Object.runBindingEffect [as fn] (property-effects.html:581)
>     at runEffectsForProperty (property-effects.html:162)
>     at runEffects (property-effects.html:128)
>     at HTMLElement._propagatePropertyChanges (property-effects.html:1692)
>     at HTMLElement._propertiesChanged (property-effects.html:1656)
>     at HTMLElement._flushProperties (properties-changed.html:341)

devem-software:
i will fix this bug next week

src/web/mjs/engine/I18n.mjs
```diff
@@ -0,0 +1,38 @@
+export default class I18n {
```
Sheepux, line 1: Class wasn't working: rewrote it (you could have used a node lib)
`
export default class I18n {

    constructor() {
        this.lang = localStorage.getItem("lang") ?? "en";
        this.dic = undefined;
        this.getDictionary(`lib/hakuneko/lang/${this.lang}.json`);
    } 

    async getDictionary(url) {
        try {
            console.log('Loading translation dictionary: ',url);
            const langfile = await fetch(url);
            const tr = await langfile.json();
            this.dic = tr;
            return tr;
        } catch (e) {
            console.error('Unable to load dictionary', url);
        }
    }

    translate(key, def = '') {
        var tr = key.split('.').reduce(function(obj, key) {
            return obj && obj[key];
        }, this.dic);
        return tr ?? def;
    }
}
`

src/web/lib/hakuneko/frontend@modern/app.js
```diff
@@ -0,0 +1,86 @@
...
+            "saved",
+            this.onSettingsSaved.bind(this)
+        );
+    }
+
+    /**
+     *
+     */
+    getAppearance() {
+        let appearance = "";
```
Sheepux, line 30: This is missing a default value and results (when you have never used the modern theme) in a broken blank page. You should set a default color (indigo or blue ?)

src/web/mjs/engine/I18n.mjs
```diff
@@ -0,0 +1,38 @@
...
+    constructor() {
+        this.fs = require("fs");
+        this.path = require("path");
+        this.lang =
+            localStorage.getItem("lang") === null
+                ? "es"
```
Sheepux: "es" shouldn't be the default language but "en"

src/web/mjs/engine/I18n.mjs
```diff
@@ -0,0 +1,38 @@
+export default class I18n {
+    constructor() {
+        this.fs = require("fs");
+        this.path = require("path");
+        this.lang =
+            localStorage.getItem("lang") === null
```
Sheepux, line 6: can simplify default value with ??
`this.lang = localStorage.getItem("lang") ?? "en";`

src/web/lib/hakuneko/frontend@modern/components/menuComponent/script.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    setLanguage(e) {
+        localStorage.setItem("lang", e.model.item.value);
+        this.windowReload();
+    }
+
+    setTheme(e) {
+        localStorage.setItem("themeColor", e.model.item.value);
+        this.windowReload();
```
Sheepux, line 113: Would be nice to find a way to change the color theme without reloading (did it in haruneko).
Switching the base css load did the trick with svelte
`<link rel="stylesheet" href="css/{currentTheme}.css">`

Sheepux:
Not resolving the points listed ?

devem-software:
I need feedback pls

Sheepux:
Make sure to use the "resolve conversation" so that we know that the issues have been solved.
Also you can use the "request review" when you've handled all "current issues" to let us know that you need new feedback (we are not checking PR directly).

This weekend, even if there is still 1 blocking point pending (the default theme color) , i'll take some time to test the rest.

devem-software:
Request review

src/web/lib/hakuneko/frontend@modern/about.md
```diff
@@ -0,0 +1,67 @@
+# Modern Theme
```
27136773, line 1: What is the purpose of this file?
I recommend to track TODOs as GitHub issues and changes are already traceable via the commit history in git.

src/web/mjs/connectors/VerMangasPorno.mjs
```diff
@@ -0,0 +1,69 @@
+import Connector from "../engine/Connector.mjs";
```
27136773, line 1: This file is not related to a change in the frontend, i recommend to make a separate PR (furthermore i think this was already added)

src/web/mjs/HakuNeko.mjs
```diff
@@ -79,6 +81,10 @@ export default class HakuNeko {
         return this._enums;
     }
 
+    get I18n() {
```
27136773, line 84: A mechanic for localization on a global level should be done in a separate PR and from my point of view the current version of HakuNeko is very unlikely to implement localization at all.

If the new frontend supports localization, i recommend a scoped solution only for the new frontend.

src/web/lib/hakuneko/frontend@modern/app.css
```diff
@@ -0,0 +1,104 @@
...
+  flex-direction: column;
+  width: 100%;
+}
+.app {
+  display: flex;
+  flex-direction: row;
+  height: 100%;
+  width: 100%;
+  overflow: hidden;
+  background: var(--base);
```
27136773, line 12: Variable name is not expressive, someone who is going to change this in the corresponding file probably have no idea where this is used.
Variables should better describe their usage, e.g. `--main-panel-background-color`

src/web/lib/hakuneko/frontend@modern/app.css
```diff
@@ -0,0 +1,104 @@
...
+  display: none;
+}
+.mangas {
+  display: flex;
+  flex-direction: row;
+  flex: 1;
+  margin-bottom: 1rem;
+  overflow: hidden;
+  border-radius: var(--border-radius);
+  border: 1px solid var(--accent);
```
27136773, line 33: Variable name is not expressive.

src/web/lib/hakuneko/frontend@modern/components/bookmarksComponent/script.js
```diff
@@ -0,0 +1,119 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable no-undef */
+
+class HakunekoBookmarks extends Polymer.Element {
+    /**
```
27136773, line 5: Remove empty comments (also all other places in file)

src/web/lib/hakuneko/frontend@modern/components/menuComponent/script.js
```diff
@@ -0,0 +1,181 @@
...
+        ];
+
+        this.languages = [
+            { label: this.i18n('settings.languages.english'), value: 'en'},
+            { label: this.i18n('settings.languages.spanish'), value: 'es'},
+        ]
+    }
+
+    getLanguageSelected(value) {
+        return localStorage.getItem("lang") === value ? "selected" : "";
```
27136773, line 57: Key name is to common for global scope, use something like `frontend.modern.language` instead of `lang`
Also be aware that local storage is not persistent, it can be wiped randomly by the chrome engine

src/web/lib/hakuneko/frontend@modern/components/menuComponent/script.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    getCornerSelected(value) {
+        return localStorage.getItem("corner") === value ? "selected" : "";
```
27136773, line 61: Key name is to common for global scope, use something like `frontend.modern.corner` instead of `corner`
Also be aware that local storage is not persistent, it can be wiped randomly by the chrome engine

src/web/lib/hakuneko/frontend@modern/components/menuComponent/script.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    getThemeSelected(theme) {
+        return localStorage.getItem("themeColor") === theme ? "selected" : "";
```
27136773, line 65: Key name is to common for global scope, use something like `frontend.modern.themeColor` instead of `themeColor`
Also be aware that local storage is not persistent, it can be wiped randomly by the chrome engine

src/web/lib/hakuneko/frontend@modern/components/menuComponent/script.js
```diff
@@ -0,0 +1,181 @@
...
+        this.windowReload();
+    }
+
+    setCorner(e) {
+        localStorage.setItem("corner", e.model.item.value);
+        this.windowReload();
+    }
+
+    windowReload() {
+        window.location.reload();
```
27136773, line 122: Use `if(confirm('Change requires reload, reload now?')) { ... }` to show a dialog asking the user before reload, they may not want to do it now.

src/web/lib/hakuneko/lang/en.json
```diff
@@ -0,0 +1,181 @@
+{
```
27136773, line 1: If the localization really is supposed to be used globally (for the HakuNeko engine and all frontends), then all keys for the modern frontend should be placed inside
```json
{
  "frontend" : {
    "modern": {
      // keys go here ...
    }
  }
}
```

src/web/lib/hakuneko/lang/es.json
```diff
@@ -0,0 +1,181 @@
+{
```
27136773, line 1: If the localization really is supposed to be used globally (for the HakuNeko engine and all frontends), then all keys for the modern frontend should be placed inside
```json
{
  "frontend" : {
    "modern": {
      // keys go here ...
    }
  }
}
```

27136773 (changes requested):
My major concern would be the localization. I think it should only be done in a local scope for the modern frontend, but not be added globally to the HakuNeko engine.

May be further discussed in discord for clarification...

Also please resolve all comments which are completed or rejected (with a proper explanation why a change is not necessary), so it can be verified by the commentator.

27136773:
I'm going to close this PR ☹️

Currently we don't put much effort into the current version of HakuNeko in favor of the upcoming [HaruNeko](https://github.com/manga-download/haruneko) prototype, which is planned to replace the current version sometime in the future.

Would be awesome if you are still interested in frontend development for the new version as well. Since this is a greenflied project, there is alot more room for new ideas. You may check the (incomplete) [Frontend Devleoper Guide](https://github.com/manga-download/haruneko/blob/master/doc/frontend-development-guide.asciidoc) or ask in discord for further info/help.

PR #2382, opened 2020-09-27
[sektekomik] update connectors

manh21:


src/web/mjs/connectors/SekteKomik.mjs
```diff
@@ -8,9 +8,9 @@ export default class SekteKomik extends WordPressMangastream {
         super.label = 'SEKTEKOMIK.COM';
         this.tags = [ 'webtoon', 'indonesian' ];
         this.url = 'http://sektekomik.com';
-        this.path = '/manga/?list';
+        this.path = '/manga/list-mode';
```
XenonR: May add trailing `/` to avoid redirects (`'/manga/list-mode/'`)

XenonR (approved):
Minor optimization possible. Otherwise looking good.

PR #2385, opened 2020-09-27
[ComicDays] Fix mangalist update

Sheepux:


XenonR (approved):
Even works without `console.log`😉

PR #2464, opened 2020-10-20
MH template and sites that use it

### LINKED ISSUE #2088: [Site Request] caisemh.com
**Name of the website**
CaiseMH

**Website urls** 
- Site: https://www.caisemh.com/
- Manga List: https://www.caisemh.com/booklist?end=0
- Manga example: https://www.caisemh.com/book/622
- Chapter online viewer example: https://www.caisemh.com/chapter/25041

**Languages**
Chinese

**Website relationship**
https://www.kelemh.com/ (Same content with different domain)

**Additional details**
Chapter in the site can only be accessed with mobile view browser. On PC I usually use developer tool in chrome / use responsive design mode in firefox

**Note :**
This is one of the best website containing CN Manhwa RAW.. adding this to Hakuneko will help us scanlator greatly, thank you~

### LINKED ISSUE #2469: [Site Request] 仙漫网 (XianMan123)
**Name of the website**
仙漫网 (XianMan123)

- Site: https://www.xianman123.com/
- Manga List: https://www.xianman123.com/f-1-0-0-0-0-0-1.html
- Manga example: https://www.xianman123.com/woduzidaomu/
- Chapter online viewer example: https://www.xianman123.com/woduzidaomu/1152061.html

**Languages**
Chinese

**Additional details**
Looks like it is using the theme similar to 酷漫屋 (Kuman5).

Robonau:
resolves #2088
resolves #2469
also changed Kuimh to use template


src/web/mjs/connectors/templates/MH.mjs
```diff
@@ -0,0 +1,69 @@
...
+
+    constructor() {
+        super();
+        super.id = undefined;
+        super.label = undefined;
+        this.tags = [];
+        this.url = undefined;
+
+        this.path = '/booklist?page=';
+        this.pathMatch = 'page=(\\d+)';
```
27136773: I'm surprised this works since this is a string.
However i would still recommend to use a regex literal.
`/page=(\d+)/`

src/web/mjs/connectors/templates/MH.mjs
```diff
@@ -0,0 +1,69 @@
...
+        this.queryMangasPageCount = 'div.page-pagination div.pagination li:nth-last-child(2) a';
+        this.queryMangas = 'ul.mh-list li div.mh-item-detali h2.title a';
+        this.queryMangasURL = 'section.banner_detail div.info h1';
```
27136773: Member name implies to be used in `_getMangas`, but in fact this is used to extract a title...
May rename to something like `queryMangaTitle`

src/web/mjs/connectors/templates/MH.mjs
27136773: May also rename to `queryMangasPageCountMatcher`, to imply that this belongs to the logic that extract the manga page count from the pagination link

src/web/mjs/connectors/Caisemh.mjs
```diff
@@ -0,0 +1,23 @@
...
+        this.tags = [ 'manga', 'webtoon', 'chinese', 'hentai' ];
+        this.url = 'https://www.caisemh.com';
+
+        this.queryPages = 'div#cp_img source';
+    }
+
+    async _getPages(chapter) {
+        this.requestOptions.headers.set('x-user-agent', 'Mozilla/5.0 (Linux; Android 9; Pixel) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4026.0 Mobile Safari/537.36');
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryPages);
```
27136773, line 21: Should not await, otherwise any other request made by this connector will use the mobile user agent that is set until the await is completed.

Either use a promise to continue instantly, or even better don't modify the user agent in the global request options and instead declare some new request options, e.g.
```javascript
const request = new Request(new URL(chapter.id, this.url), {
    headers: {
        'x-user-agent': 'Mozilla/5.0 Android X'
    }
});
```

27136773:
Can Kuman55 also be migrated to use the new template?

src/web/mjs/connectors/templates/MH.mjs
Robonau: that's what Kuimh was before, il change anyways

Robonau:
Kuman55 has quite a lot of differences despite looking the same

Robonau:
im actually done now, i don't have any more sites that use the template

src/web/mjs/connectors/Caisemh.mjs
```diff
@@ -0,0 +1,24 @@
...
+        this.tags = [ 'manga', 'webtoon', 'chinese', 'hentai' ];
+        this.url = 'https://www.caisemh.com';
+
+        this.queryPages = 'div#cp_img source';
+    }
+
+    async _getPages(chapter) {
+        const request = new Request(new URL(chapter.id, this.url), {
+            headers: {
+                'x-user-agent': 'Mozilla/5.0 Android X'
```
27136773: Does this work with this user agent?
I was just suggesting with some random string.

src/web/mjs/connectors/templates/MH.mjs
```diff
@@ -0,0 +1,69 @@
...
+export default class MH extends Connector {
+
+    constructor() {
+        super();
+        super.id = undefined;
+        super.label = undefined;
+        this.tags = [];
+        this.url = undefined;
+
+        this.path = '/booklist?page=';
```
27136773: May use a placeholder to allow better control for websites that do not use the page
number as last element, e.g.
`'/booklist?page=%PAGE%';`

src/web/mjs/connectors/templates/MH.mjs
```diff
@@ -0,0 +1,69 @@
...
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryMangaTitle);
+        let id = uri.pathname + uri.search;
+        let title = data[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let mangaList = [];
+        let request = new Request(new URL(this.path, this.url), this.requestOptions);
```
27136773: In case of the placeholder solution change this to e.g.
`new URL(this.path.replace('%PAGE%', 1), this.url)`

src/web/mjs/connectors/templates/MH.mjs
```diff
@@ -0,0 +1,69 @@
...
+        let pageCount = parseInt(data[0].href.match(this.pathMatch)[1]);
+        for(let page = 1; page <= pageCount; page++) {
+            let mangas = await this._getMangasFromPage(page);
+            mangaList.push(...mangas);
+        }
+        return mangaList;
+    }
+
+    async _getMangasFromPage(page) {
+        let request = new Request(new URL(this.path + page, this.url), this.requestOptions);
```
27136773: In case of the placeholder solution change this to e.g.
`new URL(this.path.replace('%PAGE%', page), this.url)`

src/web/mjs/connectors/Kuman5.mjs
```diff
@@ -1,21 +1,14 @@
...
 
-export default class Kuman5 extends Connector {
+export default class kuman5 extends MH {
     constructor() {
         super();
         super.id = 'kuman5';
         super.label = '酷漫屋 (Kuman55)';
         this.tags = [ 'manga', 'chinese' ];
         this.url = 'http://www.kuman55.com';
-    }
```
27136773, line 11: In case of the placeholder solution, overwrite the path, e.g.
`this.path = '/sort/1-%PAGE%.html';`

src/web/mjs/connectors/Kuman5.mjs
```diff
@@ -29,7 +22,7 @@ export default class Kuman5 extends Connector {
 
     async _getMangasFromPage(page) {
```
27136773: In case of the placeholder solution this method no longer needs to be overwritten, the base from template should work

src/web/mjs/connectors/Kuman5.mjs
```diff
@@ -1,21 +1,14 @@
...
-    async _getMangaFromURI(uri) {
-        let request = new Request(uri, this.requestOptions);
-        let data = await this.fetchDOM(request, 'section.banner_detail div.info h1');
-        let id = uri.pathname + uri.search;
-        let title = data[0].textContent.trim();
-        return new Manga(this, id, title);
+        this.queryChapter = 'div#chapterlistload ul#detail-list-select-1 li a';
     }
 
     async _getMangas() {
```
27136773, line 15: Apply the customized chapter IDs before returning the list, e.g.
```javascript
mangaList.forEach(manga => manga.id = `/mulu${manga.id}1-1.html`);
return mangaList;
When this is done, there is n o longer a need to overwrite the `_getChapters` method from the base class.
```

src/web/mjs/connectors/Kuman5.mjs
```diff
@@ -40,7 +33,7 @@ export default class Kuman5 extends Connector {
 
     async _getChapters(manga) {
```
27136773: In case of applying the customized chapter IDs in `_getMangas` method, this method no longer needs to be overwritten, the base from template should work

src/web/mjs/connectors/Caisemh.mjs
Robonau: it works fine, but il change it back to the useragent that i had before

27136773:
Maybe also add #2469

src/web/mjs/connectors/Caisemh.mjs
27136773: It is possible that the server just checked if the user agent contains `ios|android`, so it probably worked with this minimalistic user agent

src/web/mjs/connectors/Kuman5.mjs
09morbab: *`_getChapters`

09morbab:
reoslves #2469

src/web/mjs/connectors/templates/MH.mjs
```diff
@@ -0,0 +1,69 @@
...
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.childNodes[0].nodeValue.trim()
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryPages);
+        return data.map(element => this.getAbsolutePath(element.dataset.echo || element, request.url));
```
27136773: May extend options to cover cases from `iqiyi ` and `mkzhan`, so these websites don't need to overwrite it
`this.getAbsolutePath(element.dataset.src || element.dataset.original || element.dataset.echo || element, request.url)`

src/web/mjs/connectors/mkzhan.mjs
```diff
@@ -0,0 +1,36 @@
...
+
+        this.path = '/category/?page=%PAGE%';
+        this.queryMangasPageCount = 'div#Pagination a:nth-last-child(2)';
+        this.queryMangas = 'div.cate-comic-list div.common-comic-item a.cover';
+        this.queryChapter = 'a.j-chapter-link';
+        this.queryMangaTitle = 'div.de-info__box p.comic-title ';
+        this.queryPages = 'div.rd-article-wr div.rd-article__pic source';
+    }
+
+    async _getPages(chapter) {
```
27136773: May merge logic into template and remove this method

src/web/mjs/connectors/iqiyi.mjs
```diff
@@ -0,0 +1,49 @@
...
+        this.path = '/manhua/category/%E5%85%A8%E9%83%A8_0_9_%PAGE%';
+        this.queryMangasPageCount = 'div.mod-page a:nth-last-child(2)';
+        this.pathMatch = /_0_9_(\d+)/;
+        this.queryMangas = 'ul.cartoon-hot-ul li.cartoon-hot-list a.cartoon-cover';
+        this.queryChapter = 'div.chapter-container ol.chapter-fixhei li a';
+        this.queryPages = 'ul.main-container li.main-item source';
+
+    }
+
+    async _getPages(chapter) {
```
27136773: May merge logic into template and remove this method

src/web/mjs/connectors/mkzhan.mjs
```diff
@@ -0,0 +1,36 @@
+import MH from './templates/MH.mjs';
+
+export default class mkzhan extends MH {
+
+    constructor() {
+        super();
+        super.id = 'mkzhan';
+        super.label = 'mkzhan';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.mkzhan.com/';
```
27136773: Remove trailing `/`

src/web/mjs/connectors/iqiyi.mjs
```diff
@@ -0,0 +1,49 @@
+import MH from './templates/MH.mjs';
+
+export default class iqiyi extends MH {
+
+    constructor() {
+        super();
+        super.id = 'iqiyi';
+        super.label = 'iqiyi';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.iqiyi.com/';
```
27136773: Remove trailing `/`

27136773 (commented):
Ready for merge, as soon as the Lint errors have been fixed

27136773 (approved):
Looks good

PR #2472, opened 2020-10-21
[Site Request] Comic Trail #2465

### LINKED ISSUE #2465: [Site Request] Comic Trail
Comic Trail (コミックトレイル)

- Site: https://comic-trail.jp/
- Manga List: https://comic-trail.jp/series
- Manga example: https://comic-trail.jp/series/ishesan
- Chapter online viewer example: https://comic-trail.jp/pt/ishesan_00001/

Language: JP

This should use the SpeedBinb template.

Robonau:
resolves #2465
this took me way too long


src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,54 @@
...
+import Manga from '../engine/Manga.mjs';
+
+export default class ComicTrail extends SpeedBinb {
+
+    constructor() {
+        super();
+        super.id = 'comictrail';
+        super.label = 'Comic Trail (コミックトレイル)';
+        this.tags = [ 'manga', 'japanese' ];
+        this.url = 'https://comic-trail.jp/';
```
27136773: Remove trailing `/`

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,54 @@
...
+                title: element.children[0].attributes[2].value.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(this.url + manga.id, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryChapters);
+        return data.map(element => {
+            let part = /\d{4}-\d{1,2}-\d{1,2}/.exec(element.children[1].firstElementChild.firstElementChild.children[1].textContent);
```
27136773: May use `test` instead of `exec`

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,52 @@
...
+    constructor() {
+        super();
+        super.id = 'comictrail';
+        super.label = 'Comic Trail (コミックトレイル)';
+        this.tags = [ 'manga', 'japanese' ];
+        this.url = 'https://comic-trail.jp';
+
+        this.queryManga = 'section.section div.columns div.column div > a';
+
+        this.queryChapters = 'section.section > ul.columns > li.column.is-full-mobile.is-half-tablet.chapter-item div.columns.is-mobile.is-gapless.chapter-item__inner';
```
27136773: Query can be simplified
`'ul.chapters li.chapter-item div.chapter-item__body'`

Robonau: it can me simplified but that one sure doesn't work

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,52 @@
...
+    async _getChapters(manga) {
+        let request = new Request(this.url + manga.id, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryChapters);
+        return data.map(element => {
+            let part = "";
+            if (!(/\d{4}-\d{1,2}-\d{1,2}/.test(element.children[1].firstElementChild.firstElementChild.children[1].textContent))){
+                part = ` ${element.children[1].firstElementChild.firstElementChild.children[1].textContent.trim()}`;
+            }
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(`${element.firstElementChild.firstElementChild.pathname}/` || "", this.url),
```
27136773: Get ID with CSS query
`element.querySelector('a.button.is-read').href`

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,52 @@
...
+                title: element.children[1].firstElementChild.firstElementChild.firstElementChild.textContent.trim() + part
```
27136773: Get title with CSS query
`element.querySelector('p.has-text-weight-bold').textContent.trim()`

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,52 @@
...
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(this.url + manga.id, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryChapters);
+        return data.map(element => {
+            let part = "";
+            if (!(/\d{4}-\d{1,2}-\d{1,2}/.test(element.children[1].firstElementChild.firstElementChild.children[1].textContent))){
```
27136773: No sure what the reason behind this is, i would just remove it and get id/title with CSS query.

Robonau: the second line is sometimes a date like "2019-9-11" and sometimes like "part 2", the "part 2" is necessary the date is not

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,52 @@
...
+        let title = data[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let request = new Request(this.url + '/series', this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryManga);
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element.href, request.url),
```
27136773: If element is an `anchor` the `.href` is redundant and can be removed

src/web/mjs/connectors/ComicTrail.mjs
Robonau: so, if (not date) put at end of title

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -0,0 +1,52 @@
...
+                title: element.children[0].attributes[2].value.trim()
```
27136773: Isn't there a CSS query to get the title?
`element.querySelector('... input[title]').title.trim()`
If not, get attribute by name
`.getAttribute('title')` instead of index `[2]`

src/web/mjs/connectors/ComicTrail.mjs
Robonau: yea, you right just grab both with a queryall

src/web/mjs/connectors/ComicTrail.mjs
27136773: I tested it in browser and it worked fine, except it goes one level deeper than the original query and therefore needs also adjustments in the chapter queries, e.g.
`element.child[0].child[3].child[1].textContent` would be `element.child[3].child[1].textContent`

src/web/mjs/connectors/ComicTrail.mjs
27136773: I think i understand the concern about the part, in that case the query could be like this:
`title: [...element.querySelectorAll('p.has-text-weight-bold')].map(p => p.textContent.trim()).join(' ');`

Grab all parts and concatenate them with a space...

src/web/mjs/connectors/ComicTrail.mjs
Robonau: needs avoidance if, if element.querySelector('a.button.is-read') = null 
cause no like null.href

src/web/mjs/connectors/ComicTrail.mjs
Robonau: doesn't like querySelector

Robonau:
looks just as messy, but ehh

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -39,13 +39,15 @@ export default class ComicTrail extends SpeedBinb {
...
         return data.map(element => {
-            let part = "";
-            if (!/\d{4}-\d{1,2}-\d{1,2}/.test(element.children[1].firstElementChild.firstElementChild.children[1].textContent)){
-                part = ` ${element.children[1].firstElementChild.firstElementChild.children[1].textContent.trim()}`;
+            let id = "";
+            if (element.querySelector('a.button.is-read') != null){
+                id = `${element.querySelector('a.button.is-read').href}/`;
             }
+            let title = '';
+            element.querySelectorAll('p.has-text-weight-bold').forEach(element => title=title.concat(element.textContent.trim()));
```
27136773: Do not use outer variable name `element` as scoped inner variable name
`element.querySelectorAll(...).forEach(el => el ...)`

27136773: Should there be a space between the concatenated parts?
Maybe use `el => title += ' ' + el.textContent.trim()`
or `title = [...element.querySelectorAll('p.has-text-weight-bold')].join(' ')` instead

src/web/mjs/connectors/ComicTrail.mjs
```diff
@@ -39,13 +39,15 @@ export default class ComicTrail extends SpeedBinb {
         let request = new Request(this.url + manga.id, this.requestOptions);
         let data = await this.fetchDOM(request, this.queryChapters);
         return data.map(element => {
-            let part = "";
-            if (!/\d{4}-\d{1,2}-\d{1,2}/.test(element.children[1].firstElementChild.firstElementChild.children[1].textContent)){
-                part = ` ${element.children[1].firstElementChild.firstElementChild.children[1].textContent.trim()}`;
+            let id = "";
+            if (element.querySelector('a.button.is-read') != null){
```
27136773: This should never be `null`, so the `if` check is obsolete, otherwise there would be chapters with empty IDs which will throw an exception sooner or later anyway.

Robonau: its there because a lot of chapters don't have IDs right now (u gotta pay for them)
i don't know how to deal with them

Robonau:
if there is a better way to clean an array of undefined values please tell me
or a way for a selector to only select when `a.button.is-read` is inside a child (then i could not `if a.button.is-read` and wouldn't need to clean the array )

27136773:
You are probably looking for the [filter](https://www.w3schools.com/jsref/jsref_filter.asp) method.
```javascript
let chapters = [ /*...*/ ];
// only use chapters with valid id
chapters = chapters.filter(chapter => chapter.id);
```

27136773 (approved):
Looks good

PR #2486, opened 2020-10-23
[wuqimh]

### LINKED ISSUE #1342: Add website wuqimh
http://www.wuqimh.com

Robonau:
resolves #1342


Robonau:
like half the pages on this site are dead.

src/web/mjs/connectors/templates/SinMH.mjs
```diff
@@ -83,7 +84,7 @@ export default class SinMH extends Connector {
             });
         `;
         let request = new Request(this.url + chapter.id, this.requestOptions);
-        let data = await Engine.Request.fetchUI(request, script);
+        let data = await Engine.Request.fetchUI(request, this.scriptPages || script);
```
27136773, line 87: Remove local variable `script` and assign its value to `this.scriptPages`

Robonau: i tried that, couldn't get it to work, i assumed it had something to with it having ${this.api} in the script

09morbab: nvm i guess i did something else wrong whilst i was doing it

Robonau: if `${this.api}` (or in the chapter script `${this.queryChapters}`) is changed by the connector then with the script in `this.scriptPages` it won't get the changed value and will instead get the template value

Robonau: unless this is sorted it can't be moved 
if it is sorted then manhuagui chapters can taken out too (this could be done in the same way i did pages if there is no way around it)

src/web/mjs/connectors/wuqimh.mjs
```diff
@@ -0,0 +1,26 @@
...
+        this.tags = [ 'manga','hentai','chinese' ];
+        this.url = 'http://www.wuqimh.com';
+        this.requestOptions.headers.set('x-referer', this.url);
+
+        this.pathMatch = '/list/p-(\\d+)';
+        this.queryMangasPageCount = 'div.pager-cont span.pager a:nth-last-of-type(2)';
+        this.queryChapters = 'div.chapter-list ul li a';        
+
+        this.scriptPages =`
+            new Promise((resolve, reject) => {
```
27136773: Can be simplified to
`new Promise(resolve => resolve(cInfo.fs.map(img => 'http://' + pageConfig.host.auto + img)));`

Robonau: ok but how did u get to that answer so fast

Robonau: nvm ik how you got to it
im just dumb

src/web/mjs/connectors/templates/SinMH.mjs
27136773: Then leave as is for now

27136773 (approved):
Looks good

PR #2491, opened 2020-10-24
[ManHuaFen] add website support

vinh2201:


src/web/mjs/connectors/ManHuaFen.mjs
```diff
@@ -0,0 +1,16 @@
+import SinMH from './templates/SinMH.mjs';
+
+export default class ManHuaBei extends SinMH {
```
27136773: Class already exists, rename to `ManHuaFen`

vinh2201: I have renamed it. Sorry for this small error.

27136773 (approved):
Looks good

PR #2492, opened 2020-10-24
[TuhaoManhua] Website removal

vinh2201:
Since this website is down, this connector should also be taken down too.


27136773 (approved):
Looks good

PR #2495, opened 2020-10-24
add jpm1234 & manhuagui TLC

### LINKED ISSUE #2490: [Site Request] jpm1234
**Name of the website**
jpm1234

- Site: http://jpm1234.com
- Manga List: http://jpm1234.com/All/
- Manga example: http://jpm1234.com/xiangyaogengjinyibudeliangren/
- Chapter online viewer example: http://jpm1234.com/xiangyaogengjinyibudeliangren/1342300.html

**Languages**
Chinese

**Additional details**
This website is using SinMH template, but the links look different from another websites using this template.

Robonau:
+ some minor changes to SinMH
resolves #2490


src/web/mjs/connectors/jpm1234.mjs
```diff
@@ -0,0 +1,42 @@
...
+
+        this.scriptPages = `
+            new Promise(resolve => resolve(cInfo.fs.map(img => 'http://' + (/\\d$/.test(img) ? pageConfig.host.auto[0].replace('uploads','') : pageConfig.host.auto) + img)));
+        `;
+    }
+
+    async _getMangasFromPage(page) {
+        let request = new Request(this.url + this.pathMatch.replace('(\\d+)', page), this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryMangas , 3);
+        return data.filter(datt => datt.href != "hakuneko:").map(element => {
```
Robonau: should i move `.filter(datt => datt.href != "hakuneko:")` to SinMH.mjs
it would mean i can not overwrite _getMangasFromPage

src/web/mjs/connectors/ManHuaGui.mjs
```diff
@@ -24,23 +22,7 @@ export default class ManHuaGui extends SinMH {
...
-            chapter.title = chapter.title.replace(/\d+p$/, '');
-            return chapter;
-        });
-    }
-
-    async _getPages(chapter) {
-        let script = `
+        this.scriptPages =`
             new Promise((resolve, reject) => {
                 setTimeout(() => reject(new Error('Failed to get page links!')), 5000);
```
27136773: This is redundant.
`fetchUI` provides an additional timeout parameter which defaults to 60 seconds, but can be changed.
That said, there is no reason for timeout at all as the promise will throw as soon as any error occurs in the script.

src/web/mjs/connectors/jpm1234.mjs
27136773: This check should not be necessary, either fix the query to ignore fake links, or filter the list from super
```javascript
const mangas = await super. _getMangasFromPage(page);
return mangas.filter(manga => /* ... */);
```

27136773 (approved):
Looks good

PR #2496, opened 2020-10-25
added webmangatr conector

### LINKED ISSUE #2484: [Site Request] Webmangatr
Name of the website:
Webmangatr

•	Site:  https://webmangatr.com
•	Manga List: https://webmangatr.com/dizi-kategori/webtoon/
•	Manga example:  https://webmangatr.com/dizi/mage-demon-queen/
•	Chapter online viewer example:  https://webmangatr.com/bolum/magedemon-queen-1-sezon-1-bolum/

Language: Turkish

Robonau:
resolves #2484
it's a Movifox theme site but we don't have a template for that so i left a comment at the end encase someone needs it in the future


src/web/mjs/connectors/Webmangatr.mjs
```diff
@@ -0,0 +1,64 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class webmangatr extends Connector {
```
27136773: May use same casing for classname and filename

src/web/mjs/connectors/Webmangatr.mjs
```diff
@@ -0,0 +1,64 @@
...
+
+    constructor() {
+        super();
+        super.id = 'webmangatr';
+        super.label = 'WEBMANGATR';
+        this.tags = ['Turkish' ,'webtoon','hentai'];
```
27136773: All tags should be lower case to prevent duplicates with different casing

src/web/mjs/connectors/Webmangatr.mjs
```diff
@@ -0,0 +1,64 @@
...
+        this.url = 'https://webmangatr.com';
+    }
+
+    async _getMangas() {
+        let mangaList = [];
+        let request = new Request(this.url + '/dizi-arsivi/', this.requestOptions);
+        let data = await this.fetchDOM(request, 'div.col-lg-8 > a:last-of-type');
+        let pageCount = data[0].getAttribute('data-max');
```
27136773: May use `parseInt(data[0].dataset.max)` instead

src/web/mjs/connectors/Webmangatr.mjs
```diff
@@ -0,0 +1,64 @@
...
+        let title = data[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getPages(chapter) {
+        let script = `
+            new Promise(resolve => resolve( [...document.querySelectorAll('p > img[loading]')].map(image => image.dataset.src)))
+        `;
+        let request = new Request(this.url + chapter.id, this.requestOptions);
+        let data = await Engine.Request.fetchUI(request, script);
```
27136773, line 51: Is `fetchUI` really required, looks like this can be done with `fetchDOM`

src/web/mjs/connectors/Webmangatr.mjs
```diff
@@ -0,0 +1,64 @@
...
+        return data.map(page => this.createConnectorURI(page));
```
27136773: It creates a connector URL, but no handler is implemented in this class, so it will use the base handler.
Is this intentional?

src/web/mjs/connectors/Webmangatr.mjs
```diff
@@ -0,0 +1,64 @@
...
+        let request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'div.movie-details > div.name > a');
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
+                title: element.text.replace(manga.title, '').trim(),
+            };
+        });
+    }
+}//Movifox theme/template
```
27136773: Comment about similarities/templates should be placed above the class

src/web/mjs/connectors/Webmangatr.mjs
Robonau: idk but fetchDOM wont return anything using the same selector

27136773: No need for intermediate variable and `await`, may return directly...
`return Engine.Request.fetchUI(request, script);`

27136773 (approved):
Looks good now

PR #2497, opened 2020-10-25
[MHKAN] Website removal

vinh2201:
Since the website https://www.mhkan.com is down, the connector (MHKAN - ManHuaKan) should be taken down too.


27136773 (approved):
Looks good

PR #2499, opened 2020-10-25
[flatmanga] updated template websites

27136773:


Sheepux (approved):
/LGTM
Don't understand the warning about missing semicolon

27136773:
I probably let a PR through that had the warning

PR #2501, opened 2020-10-25
Manga ay

### LINKED ISSUE #2467: [Site Request] Manga Ay
Name of the website:
Manga Ay

•	Site:  https://manga-ay.com
•	Manga List: https://manga-ay.com/seriler
•	Manga example:  https://manga-ay.com/manga/215/musubenai-necktie
•	Chapter online viewer example: https://manga-ay.com/bolum/2048

Language: Turkish
Example site: Manga Dex

Robonau:
resolves #2467


src/web/mjs/connectors/MangaAy.mjs
```diff
@@ -0,0 +1,65 @@
...
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
+                title: element.text.replace(manga.title,'').trim(),
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        let data = (await this.fetchRegex(request, /var dataurl = '([^']*)';\svar page_array = \[([^\]]*),\];\svar server = '([^']*)';/g))[0];
```
27136773: Can `fetchDOM(request, 'div#content source')` be used instead?

Robonau: will work for webtoons but not manga
(manga uses pagination)

27136773: Overwriting `fetchRegex` from the base class makes no sense since it is neither re-used in the base class, nor in this class. Just add and simplify the code directly.
```javascript
    // ...
    let data = await response.text();
    data = data.match(/var dataurl = '([^']*)';\svar page_array = \[([^\]]*),\];\svar server = '([^']*)';/);
    const pages = data[2].split(',');
    // ...
```

27136773 (approved):
Looks good

PR #2510, opened 2020-10-25
Add some websites support for WordpressMadara template

vinh2201:


27136773:
Some of the icons are too large
Icons should be around 64x64 pixel and not larger then ~16KB

vinh2201:
Sorry for the large icons. I have changed them.

27136773 (approved):
Looks good

PR #2514, opened 2020-10-26
117漫画网 (bx117)

### LINKED ISSUE #2509: [Site Request] 117漫画网 (bx117)
**Name of the website**
117漫画网 (bx117)

- Site: http://www.bx117.com
- Manga List: http://www.bx117.com/all/
- Manga example: http://www.bx117.com/gufeng/yuanlaishihuananchenga/
- Chapter online viewer example: http://www.bx117.com/gufeng/yuanlaishihuananchenga/803726.html

**Languages**
Chinese

**Additional details**
Looks like this website use the same template as 6mh7 (http://www.6mh7.com), 友绘漫画网 (seeym) (http://www.seeym.net), but it does not seem that these websites are currently listed in connectors in Hakuneko.

Robonau:
for some reason the mobile version has pages when the desktop version doesn't
resolves #2509
no manga list cause i couldn't find one


src/web/mjs/connectors/Bx117.mjs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    async _getPages(chapter) {
+        let script = `
+            new Promise(resolve => {
+                resolve(new Array(qTcms_page.total).fill().map((_,ind) => new URL(getPicUrlP(qTcms_S_m_murl,ind),qTcms_m_weburl).href));
+            });
+        `;
+        let request = new Request(this.url + chapter.id, this.requestOptions);
+        let data = await Engine.Request.fetchUI(request, script);
```
27136773: No need for intermediate variable, may return promise directly:
`return Engine.Request.fetchUI(...);`

27136773:
The connector is missing the `async _getMangas()` method.
If not used it should still be implemented and throw a useful error message for the user.
e.g.
```javascript
async _getMangas() {
    let msg = 'This website does not provide a manga list, please copy and paste the URL containing the chapters directly from your browser into HakuNeko.';
    throw new Error(msg);
}
```

Robonau:
why is that _getMangas() not the default behavior ?

Robonau:
vinh DMed me with an issue, i'll say if it's resolved
edit: issue resolved

Robonau:
i broke something again

src/web/mjs/connectors/Bx117.mjs
```diff
@@ -0,0 +1,78 @@
...
+            for(let page = 1, run = true; run; page++) {
+                let mangas = await this._getMangasFromPage(page,3,i);
+                mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+            }
+        }
+        return mangaList;
+    }
+
+    async _getMangasFromPage(page,retries,serial ) {
+        const response = await fetch('http://m.bx117.com/statics/qingtiancms.ashx', {
```
27136773: Use `this.fetchDOM` instead of re-implementing the functionality.

Robonau: ngl i totally forgot that is what `this.requestOptions` was

27136773:
The `_getMangas` method must always be implemented by the connector, a default behavior would allow developers to bypass this rule. Since `virtual/abstract` methods are not supported by JavaScript, the approach is to implement it in the base class and throw a `Not Implemented` exception.

src/web/mjs/connectors/Bx117.mjs
```diff
@@ -0,0 +1,70 @@
...
+                mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+            }
+        }
+        return mangaList;
+    }
+
+    async _getMangasFromPage(page,serial) {
+        const request = new Request('http://m.bx117.com/statics/qingtiancms.ashx', {
+            method: 'POST',
+            body: `page=${page}&action=GetWapList&_id=listbody&pagesize=12&order=1&classid1=0&url=%2Fstatics%2Fqingtiancms.ashx&typelianzai=110${serial}`,
```
27136773: May use `URLSearchParams` to improve readability
```javascript
body: new URLSearchParams({
    page: page,
    action: 'GetWapList',
    // ...
}).toString()
```

27136773 (approved):
Looks good

Robonau:
another issue with `_getMangas `  
popped up (it gets a 520 error every now and then)
(DMed by vinh)

vinh2201:
I saw that the line 71 on the file that cause the manga list not loading.
image
image

It seems that const.request and const.data in async _getMangasFromPage(page,serial) is not match.

Robonau:
though it's a 500+ error so it should have just retried in fetchDOM no?

Robonau:
weather it works or not , there is no difference in what i send, it's a server thing, the site likely has this same issue
retrying would probably work but for some reason the try catch on `Connector.mjs:560` throws an error

Robonau:
wait, no it tries to retry and fails cause the request object has been used ???

Robonau:
welp, i cant get it to throw one anymore so i can't test possible fixes

Robonau:
maybe ronny knows a fix

27136773:
Seems to be a flaw / limitation of the `fetch` API.
Try cloning the request before it is used `fetch(request.clone())` in [Connector.fetchDOM](https://github.com/Robonau/hakuneko/blob/13273fb9894eef9c23a681ed8fd72f4c1f9d7ff2/src/web/mjs/engine/Connector.mjs#L440)

PR #2516, opened 2020-10-27
added MangaRaw

### LINKED ISSUE #2433: [Site Request] manga raw club
**manga raw**
manga raw

- Site: https://www.manga-raw.club/
- Manga List: https://www.manga-raw.club/browse/
- Manga example: https://www.manga-raw.club/manga/ark/
- Chapter online viewer example: https://www.manga-raw.club/reader/ark-chapter-7/

**Languages**
mostly japanese

**Website relationship*
i don't know any

**Additional details**
they have many mangas that others don't have.

Robonau:
resolves #2433


src/web/mjs/connectors/MangaRaw.mjs
```diff
@@ -0,0 +1,78 @@
...
+            let mangas = await this._getMangasFromPage(page);
+            mangaList.push(...mangas);
+        }
+        return mangaList;
+    }
+
+    async _getMangasFromPage(page) {
+        let request = new Request(`${this.url}/browse/?results=${page}`, this.requestOptions);
+        let data = await this.fetchDOM(request, '.novel-item a',3);
+        await this.wait(250);
```
27136773: Is this wait required?
May move the wait into the `for` loop of `_getMangas`

src/web/mjs/connectors/MangaRaw.mjs
```diff
@@ -0,0 +1,78 @@
...
+        let data = await this.fetchDOM(request, 'source.loading');
+        return data.map(element => this.getAbsolutePath(element.src, request.url));
+    }
+
+    async _getChapters(manga) {
+        let script = `
+        new Promise(resolve => {
+            let chaplist = [];
+            showDivEn();
+            [...document.querySelectorAll('ul#raw > li > a')].map(data => {
```
27136773: Either assign chapterlist directly with `map` or use `forEach` instead of `map`.
The `map` function should not be used as a replacement for a loop.
```javascript
chaptersEn = [...document.querySelectorAll(...)].map(element => ...);
chaptersRaw = [...document.querySelectorAll(...)].map(element => ...);
resolve([ ...chaptersEn, ...chaptersRaw ]);
```

src/web/mjs/connectors/MangaRaw.mjs
```diff
@@ -0,0 +1,78 @@
...
+                    title: data.title.trim().replace('-eng-li',' [en]'),
+                    language: 'en'
+                })
+            })
+            showDivRaw();
+            [...document.querySelectorAll('ul#raw > li > a')].map(data => {
+                chaplist.push({
+                    id: data.pathname,
+                    title: data.title.trim(),
+                    language: 'raw'
```
27136773, line 68: Maybe use `jp`

27136773 (approved):
Looks good

src/web/mjs/connectors/MangaRaw.mjs
Robonau: cant, it has manhua that are in korean. so i figure `raw` was the best bet

Robonau: cant, some are manhua and thus are in korean, so i figure raw is more appropriate

src/web/mjs/connectors/MangaRaw.mjs
Robonau: yea , it got a 40x error without it (i think it was 403 i don't remember)

PR #2555, opened 2020-11-01
add Dmzj

### LINKED ISSUE #2550: [Site Request] 动漫之家DMZJ
**Name of the website**
动漫之家DMZJ

- Site: https://www.dmzj.com/
- Manga List: https://www.dmzj.com/category
- Manga example: https://www.dmzj.com/info/njsjm.html
- Chapter online viewer example: https://www.dmzj.com/view/beibingjiaomeimeiaidesiquhuolai/108820.html#@page=1

**Languages**
Chinese

**Website relationship*
this website has an alternative domain whose information is listed below. It seems that if you visiting the website's manga list using different domains, you would get different results. So i suggest you use two domain at same time, and return the results both of them. You can also treat them as two different websites. If you do so, i suggest you use the name "动漫之家（新版）" for www.dmzj.com and "动漫之家（旧版）" for manhua.dmzj.com
- Alternative domain: https://manhua.dmzj.com/
- Manga List: https://manhua.dmzj.com/tags/category_search/0-0-0-all-0-0-0-1.shtml#category_nav_anchor
- Manga example: https://manhua.dmzj.com/sxlhyaria/
- Chapter online viewer example: https://manhua.dmzj.com/bukejiedewodeyiqie/77189.shtml#@page=1

**Additional details**
Nothing to complement

Robonau:
first half of #2550


Robonau:
im not gonna do the second half of the issue (the second site links to the first for some chapters) idk if i should resolve it or not

manh21:
icon?

src/web/mjs/connectors/Dmzj.mjs
```diff
@@ -0,0 +1,36 @@
...
+        this.queryMangas = 'h3 > a';
+
+        this.queryPagesScript =`
+            new Promise(resolve => {
+                resolve(picArry.map(el => new URL(el,img_prefix).href));
+            });
+        `;
+    }
+
+    async _getChapters(manga) {
```
27136773: Is this method required, i would expect the one from the base class to work correctly

Robonau: i figured that as well, but i wasn't able to get it to work

27136773 (approved):
Looks good

src/web/mjs/connectors/Dmzj.mjs
Robonau: nvm

27136773: When injecting the script from the base class into the browser it works fine

image

PR #2560, opened 2020-11-01
[Qimiaomh] add website support

### LINKED ISSUE #2309: [Site Request] Qimiaomh
**Name of the website**
Qimiaomh

- Site: https://www.qimiaomh.com/
- Manga List: https://www.qimiaomh.com/list-1------updatetime--1.html
- Manga example: https://www.qimiaomh.com/manhua/4715.html
- Chapter online viewer example: https://www.qimiaomh.com/manhua/4715/28.html
**Languages**
Chinese
**Website relationship*

**Additional details**

manh21:
Resolve #2309


src/web/mjs/connectors/Qimiaomh.mjs
```diff
@@ -0,0 +1,84 @@
+import Connector from '../engine/Connector.mjs';
+
+export default class EpikManga extends Connector {
```
27136773: Class already exist

src/web/mjs/connectors/Qimiaomh.mjs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    async _getMangasFromPage(page) {
+        const uri = new URL('/list-1------updatetime--' + page + '.html', this.url);
+        const request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryMangas);
+        let checkData = await this.fetchDOM(request, 'div.paging ul li a');
+        let result = {
+            data: null,
+            run: true
```
27136773: When `run` is true it means it should stop running, this is not intuitive.
Either rename the field, or invert the value.

src/web/mjs/connectors/Qimiaomh.mjs
```diff
@@ -0,0 +1,84 @@
...
+                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
+                title: element.text.trim()
+            };
+        });
+
+        result.run = checkData.find(element => {
+            return element.href == "javascript:void(0)";
+        });
+
+        if(page === 1)result.run = null;
```
27136773: May use short-circuiting to combine conditions instead of a separate conditional check
`result.run = page === 1 || checkData.find(...);`

src/web/mjs/connectors/Qimiaomh.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+        return result;
+    }
+
+    async _getMangas() {
+        let mangaList = [];
+        for(let page = 1, run = true; run; page++) {
+            let result = await this._getMangasFromPage(page);
+            result.run ? run = false : mangaList.push(...result.data) ;
```
27136773: As mentioned above, it is not intuitive to abort the loop when `result.run` is `true`

27136773: Shouldn't the result still be pushed to the manga list for the current result?
If i read it correctly it just indicates that their is no next page, but the current page may still hold valid results...

src/web/mjs/connectors/Qimiaomh.mjs
```diff
@@ -0,0 +1,84 @@
...
+        }
+        return mangaList;
+    }
+
+    async _getChapters(manga) {
+        const uri = new URL(manga.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryChapters);
+        console.log(data);
```
27136773: Remove debug logging from productive code

src/web/mjs/connectors/Qimiaomh.mjs
27136773: Maybe it is possible to just provide an empty manga list if there are no further results, this would make this whole tracking of `result.run` obsolete, something like this:
`result.length ? mangaList.push(...result) : run = false;`

manh21: Yes, the current result is valid but the next result won't be valid.
example: 
open page 1000 by URL it will display the last page which is page 185 (no redirect)

manh21: > Maybe it is possible to just provide an empty manga list if there are no further results, this would make this whole tracking of `result.run` obsolete, something like this:
> `result.length ? mangaList.push(...result) : run = false;`

will cause infinite loops

27136773: Yes, this would require `_getMangasFromPage` to return an empty list, if the end is reached (instead of setting the run flag)

src/web/mjs/connectors/Qimiaomh.mjs
manh21: need to override check explicitly for the first page

27136773 (approved):
Seems to work and can be merged

src/web/mjs/connectors/Qimiaomh.mjs
manh21: yeah, either way still need a flag to determine the last page

manh21:
Ready for merge

PR #2582, opened 2020-11-05
Alphapolis

### LINKED ISSUE #93: Add mangapolice
https://www.alphapolis.co.jp/manga/official/44000215

Robonau:
resolves #93


Robonau:
an improvement was suggested in dms

src/web/mjs/connectors/Alphapolis.mjs
```diff
@@ -0,0 +1,61 @@
...
+import Manga from '../engine/Manga.mjs';
+
+export default class Alphapolis extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'alphapolis';
+        super.label = 'ALPHAPOLIS (アルファポリス)';
+        this.tags = ['manga', 'japanese', 'hentai'];
+        this.url = 'https://www.alphapolis.co.jp/';
```
27136773: Remove trailing `/`

src/web/mjs/connectors/Alphapolis.mjs
```diff
@@ -0,0 +1,64 @@
...
+        `;
+        const request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        const data = await Engine.Request.fetchUI(request, script);
```
27136773: No need for `data`, may use `return Engine.Request.fetchUI` directly

27136773 (approved):
Looks good

src/web/mjs/connectors/Alphapolis.mjs
```diff
@@ -0,0 +1,63 @@
...
+        `;
+        const request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        return await Engine.Request.fetchUI(request, script);
```
27136773, line 50: No need to use await, a `Promise` is fully compatible with async, in fact async function promisifies the result:
"_If the result is a promise, return it, if not wrap the result in a promise and return it_"

PR #2587, opened 2020-11-06
add manmanapp

### LINKED ISSUE #390: Add support for Manman Comic 漫漫漫画
https://www.manmanapp.com

Separated from #388

Robonau:
resolves #390


src/web/mjs/connectors/manmanapp.mjs
```diff
@@ -0,0 +1,79 @@
...
+        super.id = 'manmanapp';
+        super.label = 'Manman Comic 漫漫漫画';
+        this.tags = [ 'manga', 'webtoon', 'chineses' ];
+        this.url = 'https://www.manmanapp.com';
+    }
+
+    async _getMangaFromURI(uri) {
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, 'li.title');
+        console.log(data)
```
27136773: Remove debug info from productive code, also prefer chrome debugger over console logging

src/web/mjs/connectors/manmanapp.mjs
```diff
@@ -0,0 +1,79 @@
...
+                page: page
+            })
+        });
+        const datt = await this.fetchJSON(request);
+        return datt.code == 1 ? datt.data.map(element => {
+            return {
+                id: '/comic/detail-'+element.id+'.html',
+                title: element.title
+            };
+        }): '';
```
27136773: Should return empty array `[]` instead of empty string

src/web/mjs/connectors/manmanapp.mjs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    async _getChaptersFromPage(page,id) {
+        let request = new Request(new URL('/works/comic-list-ajax.html', this.url), {
+            method: 'POST',
+            body: new URLSearchParams({
+                id: id,
+                sort: 0,
+                page: page
+            })
```
27136773, line 62: Does this work without `.toString()`?
Does this work without setting `content-type` header to `application/x-www-form-urlencoded`?

src/web/mjs/connectors/manmanapp.mjs
```diff
@@ -0,0 +1,79 @@
...
+                id: '/comic/detail-'+element.id+'.html',
+                title: element.title
+            };
+        }): '';
+    }
+
+    async _getPages(chapter) {
+        const request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        const data = await this.fetchDOM(request,'source.man_img');
+        return data.map(ele => ele.src)
```
27136773: Is `ele.src` always and absolute path or should it be wrapped into `this._getAbsolutePath(ele, request.url)`?

src/web/mjs/connectors/manmanapp.mjs
Robonau: ik how to use console log 
idk how to use chrome debugger properly

src/web/mjs/connectors/manmanapp.mjs
Robonau: they are numbers why would i have to `.toString()`
and yes it works without setting `content-type` header to `application/x-www-form-urlencoded`

src/web/mjs/connectors/manmanapp.mjs
Robonau: it is, but il do it for consistency's sake

src/web/mjs/connectors/manmanapp.mjs
Robonau: its `content-type: application/x-www-form-urlencoded;charset=UTF-8` without setting it

src/web/mjs/connectors/manmanapp.mjs
27136773: 1. in HakuNeko press F12
2. switch to sources
3. navigate to the file you want to debug and open it
4. set a breakpoint at the beginning of the line
5. interact with the application, e.g. select manga/chapter
6. wait to reach breakpoint
7. hover cursor over variables to see values or use console.log to dump data
8. use the debugger controls such as step-in, step-over, continue to proceed

See also: https://developers.google.com/web/tools/chrome-devtools/javascript

image

src/web/mjs/connectors/manmanapp.mjs
27136773: `URLSearchParams` is an object, but body is expected to use a string, thats why usually `URLSearchParams` is required to being serialized with `URLSearchParams.toString`.

If it is working, then there is no problem as it would probably being serialized internally when creating the request.

27136773 (approved):
Looks good

src/web/mjs/connectors/manmanapp.mjs
Robonau: ty

PR #2588, opened 2020-11-06
lint that was missed on the lint update

Robonau:
maybe mac lint is different to win lint?
i don't think it breaks anything


27136773 (approved):
Looks good

PR #2595, opened 2020-11-07
[DM5] add website support

### LINKED ISSUE #253: Add support for website: dm5
https://www.dm5.com/

manh21:
Resolve #253


src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class Dm5 extends Connector {
```
27136773, line 6: Add a comment that they use MangaFox reader

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+                }, 500);
+            });
+        `;
+        const uri = new URL(manga.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        return await Engine.Request.fetchUI(request, script);
+    }
+
+    async _getPages(chapter) {
+        const script = `
```
27136773, line 61: This could possibly be simplified, take a look how this is done in MangaFox.
From a brief view the best option is to use the mobile domain `m.dm5.com` and then just access the `newImgs` variable on the chapter page.

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+                                type: 'GET',
+                                error: function (msg) {
+                                },
+                                success: function (msg) {
+                                    data = eval(msg);
+                                }
+                            });
+                            data.length ? pageList.push(...data) : run = false;
+                        }
+                        resolve(pageList);
```
27136773, line 95: Looks to me like this is resolved before all the async ajax requests are completed and therefore images might be missing.

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+                setTimeout(() => {
+                    try {
+                        let button = document.querySelector('#checkAdult');
+                        if(button) {
+                            button.click();
+                        }
+                        let chapterList = [...document.querySelectorAll('${this.queryChapters}')].map(element => {
+                            return {
+                                id: new URL(element.href, window.location).pathname,
+                                title: element.text.replace(${regex}, '').trim()
```
27136773: The regex has no special characters and is only used once, so it could be directly embedded

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.title.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        // https://regex101.com/r/vD6dIl/2/
+        const regex = /（(.*?)）/gim;
```
27136773: May use more strict regex `/（[0-9]+P）/gi`

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+            new Promise((resolve, reject) => {
+                setTimeout(() => {
+                    try {
+                        let button = document.querySelector('#checkAdult');
+                        if(button) {
+                            button.click();
+                        }
+                        let chapterList = [...document.querySelectorAll('${this.queryChapters}')].map(element => {
+                            return {
+                                id: new URL(element.href, window.location).pathname,
```
27136773: May use `element.pathname` directly without wrapping in URL

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+        super();
+        super.id = 'dm5';
+        super.label = 'DM5 漫画';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.dm5.com';
+        this.path = '/manhua-list-p%PAGE%/';
+
+        this.queryMangas = 'div.box-body ul li div.mh-item div.mh-item-detali h2.title a';
+        this.queryChapters = 'div#chapterlistload ul#detail-list-select-1 li a';
+        this.queryPages = 'div#showimage div#cp_img img#cp_image';
```
27136773: Remove unused field `queryPages`

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+                        resolve(chapterList);
+                    } catch(error) {
+                        reject(error);
+                    }
+                }, 500);
+            });
+        `;
+        const uri = new URL(manga.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        return await Engine.Request.fetchUI(request, script);
```
27136773: Can this be done with `fetchDOM`?
Haven't found a chapter with adult protection, but i'm pretty sure this should work as well...

src/web/mjs/connectors/Dm5.mjs
manh21: jquery ajax configured to be not async `async: false,` [See](https://stackoverflow.com/questions/18366118/when-to-use-async-false-and-async-true-in-ajax-function-in-jquery)

src/web/mjs/connectors/Dm5.mjs
manh21: Adult content handler `https://www.dm5.com/manhua-satanophany/`

27136773: Try to [set cookie](https://github.com/manga-download/hakuneko/blob/138a68892b0f046fe79350644846542f72ad349e/src/web/mjs/connectors/templates/SinMH.mjs#L11) `isAdult=1` in request options and check if using `fetchDOM` will work

src/web/mjs/connectors/Dm5.mjs
27136773: Maybe also note a reference to SinMH since it has partial similarities

src/web/mjs/connectors/Dm5.mjs
27136773: Then this may take very long depending on the number of pages.
If the user didn't get a response within some seconds and/or switches chapters fast, requests may stack up and degrade performance.

I recommend using `createConnectorURI` approach which will basically extract the images in `handleConnectorURI` when they are requested.
See: https://github.com/manga-download/hakuneko/wiki/Developer-Website-Tutorial#6-get-protected-image-links

A matching use case might be LireScan, where it parses out the HTML pages (e.g. `/ch17/p01.html`, `/ch17/p02.html`) and then in the handler extracts the image link from the HTML page and returns the image data

src/web/mjs/connectors/Dm5.mjs
```diff
@@ -0,0 +1,125 @@
...
+                            let data;
+                            if ($("#dm5_key").length > 0) {
+                                mkey = $("#dm5_key").val();
+                            }
+                            ajaxResult = $.ajax({
+                                url: 'chapterfun.ashx',
+                                async: false,
+                                data: { cid: DM5_CID, page: dmpage, key: mkey, language: 1, gtk: 6, _cid: DM5_CID, _mid: DM5_MID, _dt: DM5_VIEWSIGN_DT, _sign: DM5_VIEWSIGN },
+                                type: 'GET',
+                                error: function (msg) {
```
27136773: This may lead to missing images without the user noticing it, may `reject` the promise in case an image could not be extracted

src/web/mjs/connectors/Dm5.mjs
manh21: yes, it is working

27136773 (approved):
Looks ggod

PR #2598, opened 2020-11-07
fixed the static json request Top toon

Robonau:
i couldn't get it all working inside fetchUI, but it really only needed the value of jsonFileUrl


27136773 (approved):
Looks good

Issue #2578, opened 2020-11-04
[Site Request] https://ww8.readonepiece.com


tsentered:
**Name of the website**
ReadOnePiece.com

Please add ReadOnePiece.com if you get a chance.

Thank you.

manh21:
See #2381

tsentered:
Thanks, at least, for the consideration and the tip about mangacruzers. Not sure if its permanent, but their website is currently down.

PR #2605, opened 2020-11-07
[mangacruzers] revise website as collection of (sub)domains

27136773:
Since MangaCruzers is down, the domains has been added hard coded to the connector.
References #2578


Sheepux (approved):
/LGTM
Warning const invasion ! 👍 
Hopefully they wont add new mangas/domains frequently

27136773:
Lets see what happens... users will probably report when one of them is no longer working

PR #2606, opened 2020-11-08
add mangaz (back to not working)

### LINKED ISSUE #34: Add Manga Library Z
https://www.mangaz.com/

Robonau:
the _getPages script works on site but wont fetch in HN (res never gets a value) (request options don't do anything) (ignore the commented out code, i haven't gotten around to it)

i've spent like 3 hours trying to get it to work
edit:
working now 
resolves #34


Robonau:
a promise.all fixed it

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -11,36 +11,38 @@ export default class Mangaz extends Connector {
...
-    // }
+    async _getMangaFromURI(uri) {
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, 'li.title');
+        const id = uri.pathname + uri.search;
+        const title = data[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
 
     async _getPages(chapter) {
```
27136773, line 47: The method `_getPages` is supposed to return links to images, but in this case the method will download all images of the chapter at once and store them in memory.
This can become very slow if a chapter has dozens of images and is also memory inefficient especially for users with low end systems.

There is a Connector Handler mechanism designed for such use cases.
See: https://github.com/manga-download/hakuneko/wiki/Developer-Website-Tutorial#get-protected-image-links
Sample: [SundayWebry](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/SundayWebry.mjs#L54-#L78)

In general `_getPages` would return a list of protocol links generated in the following way:
```javascript
async _getPages(chapter) {
    return [...].map(... => {
        this._createConnectorURI({
            url: ... // link to the encrypted image data
            iv: ... // integration vector used for decrypting the image
            key: ... // key used for decrypting the image
        });
    });
}

    // download encrypted image data from payload.url
    // decrypt image data with CryptoJS AES using payload.iv and payload.key
    // return image data (mime typed buffer)
}
```

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,83 @@
...
+        const request = new Request(new URL(chapter.id.replace('navi', 'virgo/view'), this.url), this.requestOptions);
+        return (await Engine.Request.fetchUI(request, script)).map(ele => this.createConnectorURI(ele));
+    }
+
+    async _getChapters(manga) {
+        const request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        const data = await this.fetchDOM(request, 'body');
+        return data[0].querySelector("li.box") ? [...data[0].querySelectorAll("li.box")].map(ele => {
+            return{
+                id:ele.querySelector('button').dataset['url'],
```
27136773: Replace `navi` with `virgo/view` if necessary

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,83 @@
...
+                title:ele.querySelector('span').textContent.trim()
+            };
+        }): [{
+            id:data[0].querySelector('button').dataset['url'],
```
27136773: Replace `navi` with `virgo/view` if necessary

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,83 @@
...
+                        let a = forge.aes.startDecrypting(b.Enc.key, b.Enc.iv);
+                        await a.update(forge.util.createBuffer(data));
+                        return 'data:image/jpg;base64,' + a.output.toString();
+                    }catch(error){
+                        reject(error)
+                    }
+                })));
+            });
+        `;
+        const request = new Request(new URL(chapter.id.replace('navi', 'virgo/view'), this.url), this.requestOptions);
```
27136773: If possible the replacement should be done in the chapter extraction, i would expect a correct link/id at this point

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,83 @@
...
+        const data = await this.fetchDOM(request, 'li.title');
+        const id = uri.pathname + uri.search;
+        const title = data[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getPages(chapter) {
+        const script = `
+            new Promise(async (resolve,reject) => {
+                let g = JCOMI.namespace("JCOMI.document")
```
27136773, line 50: Use more expressive variable names when used in larger scopes, avoid single characters
Bad: `let a = b.c(d)`
Good: `let minified = text.remove(spaces)`

27136773:
Missing logo for added website

src/web/mjs/connectors/mangaz.mjs
Robonau: i was just using the same ones as the site did to avoid confusion while writing it

Robonau:
i have no idea how to decrypt it to an array buffer

outside the method used by the site i have no idea how to decrypt it at all

Robonau:
ive spent like 12 hours on getting pages, imma give it a rest for today

package.json
```diff
@@ -46,5 +46,8 @@
     "test:app": "jest --color --reporters=default --reporters=jest-junit --testPathPattern=/src/app",
     "test:web": "jest --color --reporters=default --reporters=jest-junit --testPathPattern=/src/web",
     "test:e2e": "jest --color --reporters=default --reporters=jest-junit --testPathPattern=/src/__tests__"
+  },
+  "dependencies": {
+    "node-forge": "^0.10.0"
```
27136773: Added dependencies in the package file are not available in the deployed clients.
Use `CryptoJS.AES.decrypt` instead of adding another package for cryptography

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,90 @@
...
+                let b = g.getDoc()
+                let img = g.getImages().map(ele => g.getLocationDir('enc') + ele.file + "?vw=" + encodeURIComponent(JCOMI.namespace("JCOMI.config").getVersion()))
+                resolve({img:img,b:b});
+            });
+        `;
+        const request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+        const data = await Engine.Request.fetchUI(request, script)
+        return data.img.map(ele => this.createConnectorURI({
+            url:this.getAbsolutePath(ele, request.url),
```
27136773, line 59: Shouldn't this be `ele.img`

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,90 @@
...
+
+    async _handleConnectorURI(payload) {
+        let response = await fetch(payload.url)
+        let encrypted = await response.text();
+        let key = payload.key;
+        let iv = payload.iv;
+        let data = CryptoJS.AES.decrypt(encrypted, key, {iv:iv});
+        return {
+            mimeType: response.headers.get('content-type'),
+            data: data
```
27136773: Isn't data still a base64 encoded string at this point?
Therefore it probably needs to be converted to `Uint8Array` => https://stackoverflow.com/a/41106346

src/web/mjs/connectors/mangaz.mjs
Robonau: nah it maps data.img (data.img being an array filled with urls)

package.json
Robonau: i had already changed to CryptoJS.AES.decrypt, but i can't get it to output anything useful

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,90 @@
...
+        return data.img.map(ele => this.createConnectorURI({
+            url:this.getAbsolutePath(ele, request.url),
+            key:data.b.Enc.key,
+            iv:data.b.Enc.iv
+        }));
+    }
+
+    async _handleConnectorURI(payload) {
+        let response = await fetch(payload.url)
+        let encrypted = await response.text();
```
27136773: Use `.arrayBuffer()` instead of `.text()`

src/web/mjs/connectors/mangaz.mjs
Robonau: it doesn't output a string , i don't even know if its outputting the right thing

src/web/mjs/connectors/mangaz.mjs
Robonau: i changed from .arrayBuffer() to .text() because i get literally nothing with .arrayBuffer() after the decrypt

Robonau: note: encrypted.arrayBuffer()

encrypted.text()

src/web/mjs/connectors/mangaz.mjs
```diff
@@ -0,0 +1,90 @@
...
+        let key = payload.key;
+        let iv = payload.iv;
+        let data = CryptoJS.AES.decrypt(encrypted, key, {iv:iv});
```
27136773: Convert `encrypted`, `iv` and `key` into the correct types that are required for CrypoJS

src/web/mjs/connectors/mangaz.mjs
27136773: Since the decrypted data is a `WordArray` it needs to be converted to `Uint8Array`

src/web/mjs/connectors/mangaz.mjs
27136773: See Type Definitions:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a81a59ac3f035e891bca0f4c5fff40e630d79065/types/crypto-js/index.d.ts#L274

src/web/mjs/connectors/mangaz.mjs
27136773: You need to convert `ArrayBuffer` into `WordArray` or something else that CryptoJS understands.
I'm not used to CryptoJS, so i cannot help any further with that... all i can say that the parameters needs to be converted to a type that works with CryptoJS and the decrypted results needs to be converted from `WordArray` to `Uint8array`

WordArrays can be created from various sources, e.g.
CryptoJS.enc.Base64.parse(...)
CryptoJS.enc.Hex.parse(...)

src/web/mjs/connectors/mangaz.mjs
Robonau: problem is i don't know what they are or what they need to be

Robonau: i have tried everything shown in every example for 5 pages of google

Robonau: i have not gotten a single thing that works in the slightest

Robonau: the only thing that has given something at least decrypted properly was the forge version

Robonau: i can't find a single example of someone decrypting something encrypted with forge, with CrypoJS

src/web/mjs/connectors/mangaz.mjs
Robonau: i have tried every combination of every different way used by anyone on google and the CryptoJS docs, i have no idea how to get anything without forge

src/web/mjs/connectors/mangaz.mjs
Robonau: i have tried every combination of every different way used by anyone on google and the CryptoJS docs, i have no idea how to get anything without forge

src/web/mjs/connectors/mangaz.mjs
Robonau: a couple of the other way around, but they are not usefull

Robonau:
forge is also faster when compared to CryptoJS

Robonau:
ive spend near 12 hours on just trying to get CryptoJS to decrypt something encrypted by forge 
i give up, close this if there is nothing else

src/web/mjs/connectors/mangaz.mjs
27136773: Try with this example (you can throw this directly in HakuNeko's console for testing)
```javascript
(async function handleConnectorURI(payload) {
    const response = await fetch(payload.url);
    const encrypted = CryptoJS.lib.WordArray.create(await response.arrayBuffer());
    const iv = CryptoJS.enc.Base64.parse(btoa(payload.iv));
    const key = CryptoJS.enc.Utf8.parse(payload.key);
    let decrypted = CryptoJS.AES.decrypt({ ciphertext: encrypted }, key, { iv: iv });
    decrypted = decrypted.toString(CryptoJS.enc.Utf8);
    decrypted = Uint8Array.from(atob(decrypted), char => char.charCodeAt(0));
    return {
        mimeType: response.headers.get('content-type'),
        data: decrypted
    };
})({
    url: 'https://cf.mangaz.com/Books/158/158634/xb_D86f9/1_Maccf.xb?vw=6.5.2',
    iv: atob('fy9utLFsZVoKgjxayZ1fzw=='),
    key: 'a76819c5740c6ce6'
});
```

27136773: If you really want to use forge you need to download the [minified distribution](https://cdnjs.cloudflare.com/ajax/libs/forge/0.10.0/forge.min.js) to `./src/web/js` and include it in `./src/web/index.html`.

However, i don't see a point in using two different encryption libraries.

Furthermore i performed a benchmark with an encrypted 200 KB image where `forge` decryption time was ~`25 ms` and crypto-js was ~ `30 ms`, both are very fast and the difference is negligible compared to the time it takes to download the image data from the remote server.

Benchmark Scripts: https://pastebin.com/ve6ZNeM5

manh21: **Ronny Solution is working** 

image

manh21:
Icon?

27136773 (approved):
Looks good

PR #2618, opened 2020-11-09
[Komiku] update getMangas

manh21:
Cover all Mangas (1800) instead of 1000

### DIFF
--- a/src/web/mjs/connectors/Komiku.mjs
+++ b/src/web/mjs/connectors/Komiku.mjs
@@ -21,21 +21,25 @@ export default class Komiku extends WordPressMangastream {
         let mangaList = [];
         for(let page = 1, run = true; run; page++) {
             let mangas = await this._getMangasFromPage(page);
-            mangas.length === 20 ? mangaList.push(...mangas) : run = false;
+            mangas.length ? mangaList.push(...mangas) : run = false;
         }
         return mangaList;
     }
 
     async _getMangasFromPage(page) {
-        const uri = new URL(this.path + page + '/', this.url);
-        const request = new Request(uri, this.requestOptions);
-        const data = await this.fetchDOM(request, this.queryMangas);
-        return data.map(element => {
-            return {
-                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
-                title: element.text.trim()
-            };
-        });
+        try {
+            const uri = new URL(this.path + page + '/', this.url);
+            const request = new Request(uri, this.requestOptions);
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Komiku.mjs
```diff
@@ -9,21 +9,31 @@ export default class Komiku extends WordPressMangastream {
...
         this.queryPages = 'div#Baca_Komik img[src]:not([src=""])';
     }
 
     async _getMangas() {
-        let request = new Request(new URL(this.path, this.url), this.requestOptions);
-        let data = await this.fetchDOM(request, this.queryMangas);
+        let mangaList = [];
+        for(let page = 1, run = true; run; page++) {
+            let mangas = await this._getMangasFromPage(page);
+            mangas.length === 20 ? mangaList.push(...mangas) : run = false;
```
27136773: Does this work when the last page has only 17 mangas, will it still be added?
What happens if the last page always has 20 mangas which are repeated, should there be an additional abort limit such as `&& page < 100`?

manh21: is there a way to handle 404 in Hakuneko other than throw an error to the user? 
[http://komiku.co.id/manga/page/92/](http://komiku.co.id/manga/page/92/)  404 is expected and intentional

manh21: found solution end up using try and catch
```
        try {
            const uri = new URL(this.path + page + '/', this.url);
            const data = await this.fetchDOM(request, this.queryMangas);
            return data.map(element => {
                return {
                    id: this.getRootRelativeOrAbsoluteLink(element, request.url),
                    title: element.text.trim()
                };
            });
        } catch(error) {
            console.log(error);
            return [];
        }
    }
```

src/web/mjs/connectors/Komiku.mjs
```diff
@@ -9,23 +9,37 @@ export default class Komiku extends WordPressMangastream {
...
+            const request = new Request(uri, this.requestOptions);
+            const data = await this.fetchDOM(request, this.queryMangas);
+            return data.map(element => {
+                return {
+                    id: this.getRootRelativeOrAbsoluteLink(element, request.url),
+                    title: element.text.trim()
+                };
+            });
+        } catch(error) {
+            return [];
```
27136773: This is probably to check if the last page was reached, but what happens when an error (e.g. 503) occurs right on page 1 or 2, the user will get an incomplete manga list without any indicator of an error.

27136773 (approved):
Looks good

PR #2715, opened 2020-12-02
add tempestfansub

### LINKED ISSUE #2705: [Site Request] tempestfansub
tempestfansub

manga.tempestfansub.com

language : Turkish

Robonau:
resolves #2705


src/web/mjs/connectors/Tempestfansub.mjs
```diff
@@ -0,0 +1,16 @@
...
+
+export default class Tempestfansub extends WordPressMangastream {
+
+    constructor() {
+        super();
+        super.id = 'tempestfansub';
+        super.label = 'Tempestfansub';
+        this.tags = [ 'webtoon', 'manga', 'turkish' ];
+        this.url = 'https://manga.tempestfansub.com';
+        this.path = '/manga/';
```
27136773: Wouldn't it be better to use `/manga/list-mode/` which i think is already the default in the template?

Robonau: my bad, i just copied a different connector

27136773 (approved):
looks good

PR #2743, opened 2020-12-09
AstralLibrary fix

### LINKED ISSUE #2741: [AstralLibrary] Connector not working
No

Yes

**Describe the bug**
Doesn't load manga list.

**Screenshots**
Screenshot (727)
Screenshot (726)
Screenshot (725)

**HakuNeko:**
 - Version 6.17@d0b5e2

Robonau:
dumb fix for dumb problem
resolves #2741

there is no chapter list so i had to guess:
can't deal with decimal chapters 
thinks everything has a 0 chapter (not everything does)

### DIFF
--- a/src/web/mjs/connectors/AstralLibrary.mjs
+++ b/src/web/mjs/connectors/AstralLibrary.mjs
@@ -10,5 +10,28 @@ export default class AstralLibrary extends WordPressMadaraNovel {
         this.url = 'https://astrallibrary.net';
 
         this.novelObstaclesQuery = 'div.ad';
+        this.queryMangas = 'div.item-summary';
+    }
+
+    async _getMangasFromPage(page) {
+        let request = new Request(new URL('page/'+page+1,'https://www.astrallibrary.net/manga-tag/manga/'));
+        let data = await this.fetchDOM(request, this.queryMangas);
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element.querySelector('.list-chapter a'), request.url),
+                title: element.querySelector('a').textContent.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let chap = manga.id.match(/(\d+)\/$/)[1]
+        let arr = new Array(Number(chap)+1).fill(0)
+        return arr.map((_,ind) => {
+            return {
+                id: manga.id.replace(/(\d+)\/$/,ind),
+                title: 'chapter '+ind
+            }
+        });
     }
 }
--- a/src/web/mjs/connectors/AstralLibrary.mjs
+++ b/src/web/mjs/connectors/AstralLibrary.mjs
@@ -7,31 +7,35 @@ export default class AstralLibrary extends WordPressMadaraNovel {
         super.id = 'astrallibrary';
         super.label = 'Astral Library';
         this.tags = [ 'webtoon', 'novel', 'english' ];
-        this.url = 'https://astrallibrary.net';
+        this.url = 'https://www.astrallibrary.net';
 
         this.novelObstaclesQuery = 'div.ad';
-        this.queryMangas = 'div.item-summary';
+        this.queryMangas = 'h3.h5 a';
     }
 
     async _getMangasFromPage(page) {
-        let request = new Request(new URL('page/'+page+1,'https://www.astrallibrary.net/manga-tag/manga/'));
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/AstralLibrary.mjs
```diff
@@ -10,5 +10,28 @@ export default class AstralLibrary extends WordPressMadaraNovel {
         this.url = 'https://astrallibrary.net';
```
27136773: Add `www` subdomain

src/web/mjs/connectors/AstralLibrary.mjs
```diff
@@ -10,5 +10,28 @@ export default class AstralLibrary extends WordPressMadaraNovel {
...
 
         this.novelObstaclesQuery = 'div.ad';
+        this.queryMangas = 'div.item-summary';
+    }
+
+    async _getMangasFromPage(page) {
+        let request = new Request(new URL('page/'+page+1,'https://www.astrallibrary.net/manga-tag/manga/'));
```
27136773: Use `this.url`, in case it changes it can be simply changed in a single place.
I think brackets are required to increment the page number, otherwise the `page` is first appended to the string and the `1` may also be appended, e.g. `'xxx' + 7 + 1` will result in `xxx71` which is probably not the desired outcome.

`new URL('/manga-tag/manga/page/' + (page + 1), this.url)`

src/web/mjs/connectors/AstralLibrary.mjs
```diff
@@ -10,5 +10,28 @@ export default class AstralLibrary extends WordPressMadaraNovel {
...
+        let data = await this.fetchDOM(request, this.queryMangas);
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element.querySelector('.list-chapter a'), request.url),
```
27136773: This will get a link to the latest chapter instead of the manga and may interfere with bookmarked mangas (bookmarks are not detected, bookmarks will never find new chapters, ...).

I suggest to keep the manga list as is and not implement a hack just to bypass the bug on the website...

src/web/mjs/connectors/AstralLibrary.mjs
```diff
@@ -10,5 +10,28 @@ export default class AstralLibrary extends WordPressMadaraNovel {
...
+                title: element.querySelector('a').textContent.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let chap = manga.id.match(/(\d+)\/$/)[1]
```
27136773: Use the original `manga.id` and get the latest chapter number on the fly instead.
```javascript
const uri = new URL(this.url);
uri.searchParams.set('post_type', 'wp-manga');
uri.searchParams.set('s', manga.title);
const request = new Request(uri, this.requestOption);
const data = await fetchDOM(request, 'div.latest-chap span.chapter a');
// from this point follow the same drill by generating the chapters from the latest chapter number ...
```

27136773 (changes requested):
I suggest to not overwrite the method to get the manga list.
Instead utilize the website search to find the manga in the `getChapters` method and extract the latest chapter number from there.
Also needs to fix the lint errors.

src/web/mjs/connectors/AstralLibrary.mjs
27136773: Obsolete since this method should not be overwritten

src/web/mjs/connectors/AstralLibrary.mjs
27136773: Obsolete since this method should not be overwritten

Robonau:
i can't not overwrite the method to get the manga list.
the madara _getMangas dosnt work as they have removed /wp-admin/admin-ajax.php
if i don't overwrite it it will not give a manga list

and i figure if it needs to be done anyways i might as well do it in the _getMangasFromPage

27136773:
In that case it would be best to use the link to the manga instead of the link to a chapter

Robonau:
> In that case it would be best to use the link to the manga instead of the link to a chapter
the link to the manga is literally useless, i would just be using the title to search and get the last chapter that way, it's the same thing with more network trafic

Robonau:
nvm i didn't read the bookmark thing

27136773 (approved):
Looks good

PR #2745, opened 2020-12-11
Update Team1x1.mjs

### LINKED ISSUE #2744: [team1x1] Mangas redirecting to new domain
No

Yes

**Describe the bug**
http://team1x1.com Changed to https://tqneplus.com

**To Reproduce**
1. Name of connector team1x1.com

Robonau:
new url 
get chapters pages

these guy seem to be changing the url frequently, there doesn't seem to be a pattern though so i don't know if we can do anything about it  
resolves #2744

### DIFF
--- a/src/web/mjs/connectors/Team1x1.mjs
+++ b/src/web/mjs/connectors/Team1x1.mjs
@@ -7,56 +7,69 @@ export default class Team1x1 extends Connector {
         super();
         super.id = 'team1x1';
         super.label = 'Team X';
-        this.tags = [ 'webtoon', 'arabic' ];
-        this.url = 'https://team1x11.com';
+        this.tags = ['webtoon', 'arabic'];
+        this.url = 'https://tqneplus.com';
+
     }
 
     async _getMangaFromURI(uri) {
-        let request = new Request(uri, this.requestOptions);
-        let data = await this.fetchDOM(request, 'div.row > div.col-md-9 > h3');
-        let id = uri.pathname + uri.search;
-        let title = data[0].textContent;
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Team1x1.mjs
```diff
@@ -7,56 +7,69 @@ export default class Team1x1 extends Connector {
...
+        const data = await this.fetchDOM(request, 'div.pagination div.wp-pagenavi > a.last');
+        const pageCount = parseInt(data[0].href.split("/").slice(-2, -1)[0].trim());
+        const pageLinks = [... new Array(pageCount).keys()].map(page => request.url + 'page/' + (page + 1) + '/');
+        return await this._getMangaListFromPages(pageLinks);
+    }
+
+    async _getChapters(manga) {
+        let chapterList = [];
+        const request = new Request(this.url + manga.id, this.requestOptions);
+        const data = await this.fetchDOM(request, 'span.current');
```
27136773: May improve query to make it more robust `div.pagination span.current`.

Is `span.current` really the last page? Sounds to me more like it is the current page (which is probably always `1`).

27136773: OK, seems the website automatically sets the last page as the current, e.g.
https://tqneplus.com/manga/the-heavens-list/

image

27136773 (approved):
Looks good

src/web/mjs/connectors/Team1x1.mjs
Robonau: yea, as u said it is the current page and that is the last page, i think it's an arabic right to left thing

PR #2751, opened 2020-12-13
[gntai] removed blogger template

27136773:
References: dazedcat19/FMD2#204

PR #2752, opened 2020-12-13
[xlecx] added website support

27136773:
References: https://github.com/dazedcat19/FMD2/commit/38093e4b06f4e33c296e78577afe5d2febcd676d


src/web/mjs/connectors/XlecX.mjs
```diff
@@ -0,0 +1,52 @@
...
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.text.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        return [ Object.assign({ language: '' }, manga) ];
```
Sheepux, line 44: Same comment as gntai #2751 . how is that _getChapters working ?

27136773: Hentai websites are usually one-shots without chapters, the manga is basically the one shot chapter.
It creates a new object which acts as chapter containing a `language` property and then assigned the `id` and `title` property from the manga to it.

Sheepux (approved):

src/web/mjs/connectors/XlecX.mjs
27136773: Samples for clarification of `Object.assign`:

```javascript
let source = { b: 99, x: 0 };
let target = { a: 1, b: 2 };
Object.assign(target, source);
console.log(source, target);
```

```javascript
let manga = { id: '/comic/my-little-pony', title: 'My Big Pony' };
let chapter = { language: 'pony' };
Object.assign(chapter, manga);
console.log(chapter);
```

PR #2815, opened 2020-12-24
Use generic GraphQL method

fcaylus:
As mentioned in my previous PR, I modified `Luscious`, `ReaderFront` and `MangaHub` to use the new graphQL method


27136773 (approved):
Looks good (assuming each website was tested manually after change)

PR #2834, opened 2020-12-29
fix spelling

phanirithvij:


Sheepux (approved):
Thanks for fixing that. I basically applied french rule to put s on adjectives when the subject is plural :D

PR #2854, opened 2021-01-03
[SinMH] add some various templated websites

vinh2201:
Add some websites support SinMH template.


src/web/mjs/connectors/guoman8.mjs
```diff
@@ -0,0 +1,20 @@
+import SinMH from './templates/SinMH.mjs';
+
+export default class guoman8 extends SinMH {
+
+    constructor() {
+        super();
+        super.id = 'guoman8';
+        super.label = 'guoman8';
```
27136773: May add chinese name `国漫吧`

src/web/mjs/connectors/guoman8.mjs
```diff
@@ -0,0 +1,20 @@
+import SinMH from './templates/SinMH.mjs';
+
+export default class guoman8 extends SinMH {
```
27136773: Prefer PascalCase for class and file names

src/web/mjs/connectors/iceKr.mjs
```diff
@@ -0,0 +1,22 @@
+import SinMH from './templates/SinMH.mjs';
+
+export default class iceKr extends SinMH {
```
27136773: Prefer PascalCase for class and file names

27136773 (approved):
Looks good

vinh2201:
Updated. Please check.

PR #2871, opened 2021-01-06
Scan manga

### LINKED ISSUE #2874: [ScanManga] Connector not working
Yes

Yes

**Describe the bug**
I can read on the site and all but I can neither download chapters or update the manga list

**To Reproduce**
1. ScanManga
2. Solo Leveling
3. Chapitre 134

**Screenshots**
Desktop-screenshot (1)

 - Version [e.g. 6.17@c16c32]

Robonau:
obfuscated as it seems scanmanga is watching
changed _getMangas to "use clipboard method" as it don't work, and no way to get it to

(other changes were cause of vscode document format)

### DIFF
--- a/src/web/mjs/connectors/ScanManga.mjs
+++ b/src/web/mjs/connectors/ScanManga.mjs
@@ -9,7 +9,7 @@ export default class ScanManga extends Connector {
         super.label = 'ScanManga';
         this.tags = ['manga', 'french', 'novel'];
         this.url = 'https://www.scan-manga.com';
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/ScanManga.mjs
```diff
@@ -7,11 +7,9 @@ export default class ScanManga extends Connector {
...
+        var _0x1087 = ['381sfGWON', '916955zZjSQL', '31ElJoip', '708qNyAYT', '2ayDZDy', 'accept-language', 'set', '2RuRNTr', '681974vvHWEX', '71297ucZdkv', '19023wnFPih', 'headers', '1EktpzL', 'requestOptions', '2315QJlMKR', '765166pfLgfQ', 'en-US', '2469XKKoaj']; var _0x3907 = function (_0x376230, _0x20d038) { _0x376230 = _0x376230 - 0x1be; var _0x1087ed = _0x1087[_0x376230]; return _0x1087ed; }; var _0x59f72d = _0x3907; (function (_0x6ca478, _0x265693) { var _0x302037 = _0x3907; while (!![]) { try { var _0x5c9580 = parseInt(_0x302037(0x1bf)) * -parseInt(_0x302037(0x1c4)) + parseInt(_0x302037(0x1c3)) * parseInt(_0x302037(0x1c7)) + parseInt(_0x302037(0x1ce)) + -parseInt(_0x302037(0x1cf)) * -parseInt(_0x302037(0x1c5)) + -parseInt(_0x302037(0x1ca)) * -parseInt(_0x302037(0x1c2)) + parseInt(_0x302037(0x1c9)) * -parseInt(_0x302037(0x1be)) + -parseInt(_0x302037(0x1cd)) * parseInt(_0x302037(0x1cc)); if (_0x5c9580 === _0x265693) break; else _0x6ca478['push'](_0x6ca478['shift']()); } catch (_0x5bf218) { _0x6ca478['push'](_0x6ca478['shift']()); } } }(_0x1087, 0xf353f), this[_0x59f72d(0x1c8)][_0x59f72d(0x1c6)][_0x59f72d(0x1c1)](_0x59f72d(0x1c0), _0x59f72d(0x1cb)));
```
27136773: This obfuscation is quite useless, as it can be unveiled within a few minutes

27136773: May just set header to same value as one of the popular browsers, so it cannot be distinguished if request is from HakuNeko or Firefox, e.g.
`this.requestOptions.headers.set('accept-language', 'en-US,en;q=0.9,de;q=0.8');`

src/web/mjs/connectors/ScanManga.mjs
```diff
@@ -23,23 +21,8 @@ export default class ScanManga extends Connector {
     }
 
     async _getMangas() {
-        let request = new Request(new URL('/scanlation/scan.data.json', this.url), this.requestOptions);
```
27136773, line 26: No need to remove, just set the headers that the website wants, so it will work again...
```javascript
request.headers.set('x-cookie', `_ga=GA1.2.137581646.${parseInt(Date.now()/1000)}; addtl_consent=1~${new Array(647).fill().map(() => Math.floor(41 * Math.random())).join('.')}`);
request.headers.set('x-referer', this.url + '/scanlation/liste_series.html');
request.headers.set('x-requested-with', 'XMLHttpRequest');
```

27136773 (changes requested):
I think updating the headers is no big deal and since this is a very easy protection to bypass, there is no need to obfuscate it at all right now

src/web/mjs/connectors/ScanManga.mjs
Robonau: i mean they previously shut out all there french users to try get around HN

src/web/mjs/connectors/ScanManga.mjs
Robonau: i mean the site itself doesn't know what the website wants
the search  function doesn't work at all

the method used also gets a shit ton of manga and manhwa that aren't viewable (not on site anymore)
i believe it moved to novels only

Robonau: i legit have no idea where u got this from

27136773 (approved):
Looks good

PR #2922, opened 2021-01-12
Update ScanManga.mjs & HeaderGenerator

### LINKED ISSUE #2921: [ScanManga] Connector not working
Yes

Yes

**Describe the bug**
I can read on the site and all but I can neither download chapters or update the manga list

**To Reproduce**
1. Name of connector : ScanManga
2. Name of manga (if connector update is working) : Solo Leveling
3. Name of chapter (if manga update is working) : Chapitre 134
4. Copy any relevant error message that appears in a popup or (advanced) within the console (press F12) : 

**Screenshots**
Desktop-screenshot

 - Version [master@1688c8]

Robonau:
made the accept-language header a random string
resolves #2921


27136773 (commented):
This will not last long, as it can be easily matched with regex, e.g. `\w{2}`

I suggest to write a generator which creates [valid headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language), similar to the [user agent generator](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/engine/UserAgent.mjs).
Maybe merge them to something like HeaderGenerator.

Robonau:
randomly chooses between ~200 (real) langCodes, 1 to 3 times with a 50/50 on having q-factor weighting with a random number from 0 to 1 (1 decimal place)

i wasn't sure where else to put it so it is currently just imported in to scanmanga

Robonau:
still don't know if i should put it in request.mjs (since this is the only place userAgent was used i assume it would be here)
nor where to put it if i should

27136773 (approved):
Looks good for now

PR #2962, opened 2021-01-21
Update ElevenToon.mjs

Robonau:
changed url to current and added _getMangaFromURI

### DIFF
--- a/src/web/mjs/connectors/ElevenToon.mjs
+++ b/src/web/mjs/connectors/ElevenToon.mjs
@@ -1,4 +1,5 @@
 import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
 
 /**
  *
@@ -13,7 +14,14 @@ export default class ElevenToon extends Connector {
         super.id = '11toon';
         super.label = '11toon';
         this.tags = [ 'manga', 'korean' ];
-        this.url = 'http://www.11toon.com';
+        this.url = 'http://www.11toon3.com';
+    }
+
+    async _getMangaFromURI(uri) {
+        const request = new Request(new URL(uri), this.requestOptions);
+        const data = await this.fetchDOM(request, '#cover-info h2');
+        const title = data[0].textContent.trim();
+        return new Manga(this, uri, title);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/ElevenToon.mjs
```diff
@@ -13,7 +14,14 @@ export default class ElevenToon extends Connector {
         super.id = '11toon';
         super.label = '11toon';
         this.tags = [ 'manga', 'korean' ];
-        this.url = 'http://www.11toon.com';
```
27136773, line 16: May add same mechanism as Heaventoon/Holymanga, when domain keeps being changed regulary
https://github.com/manga-download/hakuneko/blob/9d7264bd2e30809a897a17ddc677c78183bbbd8d/src/web/mjs/connectors/HolyManga.mjs#L13-L28

27136773 (approved):
Looks good

Robonau:
cant actually test any of this cause of the no javascript cloudflare thing

src/web/mjs/connectors/ElevenToon.mjs
```diff
@@ -13,14 +13,28 @@ export default class ElevenToon extends Connector {
...
+        /*
+         * sometimes cloudflare bypass will fail, because chrome successfully loads the page from its cache
+         * => append random search parameter to avoid caching
+         */
+        let uri = new URL(this.url);
+        uri.searchParams.set('ts', Date.now());
+        uri.searchParams.set('rd', Math.random());
+        let request = new Request(uri.href, this.requestOptions);
+        let dat = await fetch(request);
+        this.url = new URL(dat.url).origin;
```
Robonau: used fetch to get around the cloudflare issue
cause the response is still from the new url

src/web/mjs/connectors/ElevenToon.mjs
```diff
@@ -1,127 +1,93 @@
...
-                        .then( chapters => chapterList.concat( chapters ) );
-                } else {
-                    return Promise.resolve( chapterList );
-                }
-            } );
+    async _getMangaFromURI(uri) {
+        const request = new Request(new URL(uri), this.requestOptions);
+        const data = await this.fetchDOM(request, '#cover-info h2');
+        const title = data[0].textContent.trim();
+        return new Manga(this, uri, title);
```
27136773: Use `uri.pathname`, the manga id should not be an `URL` object

27136773 (commented):
Can be merged as soon as Lint error and manga id are adjusted

PR #2975, opened 2021-01-23
Dark mode novels

### LINKED ISSUE #2556: [Feature] dark mode for novels
dark mode is almost a necessity for reading 

a dark mode button

```js
[...document.querySelector("body > hakuneko-app").shadowRoot.querySelector("div.content > hakuneko-pages").shadowRoot.querySelectorAll('.image')].map(imgg => imgg.style.filter = 'invert(1)')
```
works (inverts colours) but has to be reverted when going back to manga and redone on restart

**Additional context**

Robonau:
resolves #2556

### DIFF
--- a/src/web/mjs/connectors/ComicoNovel.mjs
+++ b/src/web/mjs/connectors/ComicoNovel.mjs
@@ -16,7 +16,7 @@ export default class ComicoNovel extends Connector {
     }
 
     async _getPages(chapter) {
-        let darkmode =JSON.parse(Engine._settings.NovelDarkmode.value);
+        let darkmode = Engine.Enums.NovelColorProfiles;
         let script = `
             new Promise((resolve, reject) => {
                 let script = document.createElement('script');
--- a/src/web/mjs/connectors/ComicoNovel.mjs
+++ b/src/web/mjs/connectors/ComicoNovel.mjs
@@ -16,7 +16,7 @@ export default class ComicoNovel extends Connector {
     }
 
     async _getPages(chapter) {
-        let darkmode = Engine.Enums.NovelColorProfiles;
+        let darkmode = Engine.Settings.NovelColorProfile();
         let script = `
             new Promise((resolve, reject) => {
                 let script = document.createElement('script');
--- a/src/web/mjs/connectors/ComicoNovel.mjs
+++ b/src/web/mjs/connectors/ComicoNovel.mjs
@@ -16,6 +16,7 @@ export default class ComicoNovel extends Connector {
     }
 
     async _getPages(chapter) {
+        let darkmode =JSON.parse(Engine._settings.NovelDarkmode.value);
         let script = `
             new Promise((resolve, reject) => {
                 let script = document.createElement('script');
@@ -29,14 +30,12 @@ export default class ComicoNovel extends Connector {
                                 // TODO: include figcaption.cn-picture__caption
                             } else {
                                 element.parentElement.style.margin = '0';
-                                if (${Engine._settings.NovelDarkmode.value}){
-                                    [...element.querySelectorAll(":not(:empty)")].forEach(ele => {
-                                        ele.style.backgroundColor = 'black'
-                                        ele.style.color = 'white'
-                                    })
-                                    element.style.backgroundColor = 'black'
-                                    element.style.color = 'white'
-                                }
+                                [...element.querySelectorAll(":not(:empty)")].forEach(ele => {
+                                    ele.style.backgroundColor = '${darkmode.background}'
+                                    ele.style.color = '${darkmode.text}'
+                                })
+                                element.style.backgroundColor = '${darkmode.background}'
+                                element.style.color = '${darkmode.text}'
                                 let canvas = await html2canvas(element);
                                 images.push(canvas.toDataURL('image/png'));
                             }
--- a/src/web/mjs/connectors/ComicoNovel.mjs
+++ b/src/web/mjs/connectors/ComicoNovel.mjs
@@ -30,6 +30,10 @@ export default class ComicoNovel extends Connector {
                             } else {
                                 element.parentElement.style.margin = '0';
                                 if (${Engine._settings.NovelDarkmode.value}){
+                                    [...element.querySelectorAll(":not(:empty)")].forEach(ele => {
+                                        ele.style.backgroundColor = 'black'
+                                        ele.style.color = 'white'
+                                    })
                                     element.style.backgroundColor = 'black'
                                     element.style.color = 'white'
                                 }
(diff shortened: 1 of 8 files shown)


src/web/mjs/engine/Settings.mjs
```diff
@@ -253,6 +253,20 @@ export default class Settings extends EventTarget {
...
             value: 'none'
         };
+
+        this.NovelDarkmode = {
+            label: 'Light/Dark mode for Novels',
+            description: [
+                'choose between light and dark mode for novels',
+                'this may not work for all connectors that support novels'
+            ].join('\n'),
+            input: types.select,
```
27136773, line 264: Maybe a checkbox would be more appropriate

src/web/mjs/engine/Settings.mjs
```diff
@@ -253,6 +253,20 @@ export default class Settings extends EventTarget {
             ],
             value: 'none'
         };
+
+        this.NovelDarkmode = {
+            label: 'Light/Dark mode for Novels',
+            description: [
+                'choose between light and dark mode for novels',
```
27136773: Start with a capital letter
Also a bit misleading, `mode` implies that it is a setting for the reader, that can be changed as the iser pleases, but in fact it determines the colorset in which the images are downloaded (once downloaded in a mode, it cannot be changed afterwards without re-downloading)

src/web/mjs/connectors/ComicoNovel.mjs
```diff
@@ -29,6 +29,10 @@ export default class ComicoNovel extends Connector {
                                 // TODO: include figcaption.cn-picture__caption
                             } else {
                                 element.parentElement.style.margin = '0';
+                                if (${Engine._settings.NovelDarkmode.value}){
```
27136773: Is it safe to assume that
- none of the child elements are applying a different color (e.g. dark grey) which would become almost invisible with black background
- Embedded images will work well with black background

src/web/mjs/engine/Settings.mjs
Robonau: should 

also be a check box?

other background/text colours are viable 
(im aware that it should be different if im gonna make that argument)

src/web/mjs/engine/Settings.mjs
Robonau: what would you suggest instead of mode
"dark mode" is synonymous with dark background and light text

src/web/mjs/connectors/ComicoNovel.mjs
Robonau: 1. fair
2. then don't use dark mode?, images look funky on sites when you force dark mode all the time

Robonau:
note: most of comico novel doesn't work anyways

src/web/mjs/engine/Settings.mjs
27136773: No, this is a selection list, because there can be 1...n options, but i think for light/dark mode there will onl be just the two of them.

src/web/mjs/engine/Settings.mjs
27136773: Maybe something like this:
`Check this option to download novels in dark mode (inverted colors) instead of their original colors.`

src/web/mjs/connectors/ComicoNovel.mjs
27136773: Maybe add a note in the settimg description, that using dark mode will not work on embedded images

src/web/mjs/connectors/ComicoNovel.mjs
```diff
@@ -16,6 +16,7 @@ export default class ComicoNovel extends Connector {
     }
 
     async _getPages(chapter) {
+        let darkmode =JSON.parse(Engine._settings.NovelDarkmode.value);
```
27136773: Looks like redundant code when every consumer of the settings engine has to convert the value into the appropriate format.
I suggest to either make the [UI component](https://github.com/manga-download/hakuneko/blob/master/src/web/lib/hakuneko/frontend%40classic-dark/input.html#L57-L65) work with generic types instead of just strings, or use a string as identifier (object key) and provide a [global map](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/engine/Enums.mjs) that stores the color profiles for the given identifier.

```javascript
static get NovelColorProfiles() {
    return {
        // Sample Access: const colorProfile = Engine.Enums.NovelColorProfiles['dark']
        //                colorProfile.foreground
        dark: {
            foreground: '...',
            background: '...'
        },
        light: {
           // ...
        }
    };
}
```

Robonau: 1st option, idk how to do (though would be the best option)
2nd option, seems like more work for both me(current dev), future dev(wants to add new colour) and, the computer(though im not sure about this)

Robonau:
couldn't figure out a way to deal with the option value only being a string
so i kinda did the global map thing
Engine.Enums.NovelColorProfiles gets the data wanted, but because i didn't want to have to enter data in multiple places i made NovelColorProfiles (in Enums) get the data from an extra value in the NovelColorProfile (in settings) options

so i ended up with
```js
{val: { background: 'black', text: 'white' }, value: 'Dark', name: 'Dark' },
```
for each option, though val isn't used for the ui at all, just the enums

and enums going through the options to see what one is right

src/web/mjs/connectors/ComicoNovel.mjs
Robonau: 1st option doesn't seem possible

src/web/mjs/engine/Enums.mjs
```diff
@@ -12,4 +12,8 @@ export default class Enums {
             onSelectManga:            'onSelectManga'
         };
     }
+
+    static get NovelColorProfiles() {
```
27136773: Maybe more fitting to place this into `Settings.mjs`, since there are no defined constants at all and the options itself are already in settings

27136773: Name is a bit misleading, it provides the currently selected color profile and not a list of color profiles

Robonau: you were the one that made the name
https://github.com/manga-download/hakuneko/pull/2975#discussion_r565007963

Robonau: il swap em anyways

Robonau: yea i was following https://github.com/manga-download/hakuneko/pull/2975#discussion_r565007963 to literally

27136773 (approved):
Looks good

PR #3013, opened 2021-02-01
mangaraw css query fix

Sheepux:
notified on discord.
Robo was thinking it could be a template (didn't look much, just fixed the queries)


Robonau:
i couldn't find what template it was (might be sinmh or mh),it might just be i've worked on the connector before
edit:
given i made the mangaraw connector its likely i was just remembering it

27136773 (approved):
Looks good

PR #3015, opened 2021-02-02
n-z login pages

Robonau:
ive done //dead? on all sites i couldn't connect to

```js
this.links = {
    login: 'loginURL'
};
```
on all that needed it (many use js to have a popup can't be done like this)


27136773 (approved):
Good work

PR #3020, opened 2021-02-03
[lovehug] fix chapter link extraction

### LINKED ISSUE #3028: [LoveHug] Connector not working
Yes

Yes

**Describe the bug**
Error: There was no content found...

 - Version nightly-20200705.1

27136773:
References: dazedcat19/FMD2#309

### DIFF
--- a/src/web/mjs/connectors/templates/FlatManga.mjs
+++ b/src/web/mjs/connectors/templates/FlatManga.mjs
@@ -53,6 +53,9 @@ export default class FlatManga extends Connector {
         let language = dom.querySelector(this.queryChapterLanguage);
         language = language ? language.className.match(this.queryChapterLanguageRX)[1] : this.language;
         return [...dom.querySelectorAll(this.queryChapters)].map(element => {
+            if(element.dataset.href) {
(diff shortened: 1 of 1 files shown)


Sheepux (approved):

shinigamichan:
I noticed this and patched it on my local copy. Isn't the proper way to fix this to change https://github.com/manga-download/hakuneko/blob/23c2e9fb5975b76af6786c3cff86e6aaa5e39343/src/web/mjs/connectors/templates/FlatManga.mjs#L63 to use uri (defined earlier in the function), rather than this.url?

The breakage was caused by the site switching from an absolute-path reference (prefaced with a /) `/$manga/$chapter/` to a relative-path reference `$chapter/`. Lovehug happens to include the manga path in the data-href attribute, but there's no reason a site needs to do this. The error is treating the href as relative to the base url rather than relative to the page the href was extracted from.

27136773:
Yes this will work as long as there are no cross references to other domains and `uri.href` aka `manga.id` ends with a `/`.

Anyway, they are probably in the process of adding some kind of link encryption right now, at least there are traces of this in their JavaScript on the website.

PR #3040, opened 2021-02-05
kakaopage 

Robonau:
i actually don't understand why this isn't able to access the api
it responds with something to do with cors (403)

note: copying the fetch from the site doesn't work either

i could probably fetch from within fetchUI(on the kakaopage webpage) but i'm certain that is something ronny will not like


Sheepux:
CORS are source domain validation, make sure you have the headers origin and/or referer
Origin: http://moo.com
Referer: http://moo.com/test/blahblah.html
(you can convert the PR to a draft while it's still not working)

Robonau:
its a copy paste only connector since there is nowhere you can get all manga in one place

(i forgot you could change Origin)

Robonau:
EvEr4EnGeL is having img being bin issues
we never found a solution and just assumed it was just his network issues

Robonau:
officially gave up on EvEr4EnGeL issue
tried everything i could think of

src/web/mjs/connectors/kakaopage.mjs
```diff
@@ -0,0 +1,71 @@
...
+        this.url = 'https://page.kakao.com';
+        this.requestOptions.headers.set('x-origin', this.url);
+        this.requestOptions.headers.set('x-referer', this.url);
+    }
+
+    async _getMangaFromURI(uri) {
+        uri = new URL(uri);
+        const request = new Request(uri, this.requestOptions);
+        let id = uri.searchParams.get('seriesId');
+        let data = await this.fetchDOM(request, 'h2', 3);
```
27136773: May be a little bit more specific with CSS query to prevent false positives

src/web/mjs/connectors/kakaopage.mjs
```diff
@@ -0,0 +1,71 @@
...
+    async _getChaptersFromPage(manga, page) {
+        let form = new FormData();
+        form.append('seriesid', manga.id);
+        form.append('page', page);
+        form.append('direction', 'desc');
+        form.append('page_size', 20);
+        form.append('without_hidden', true);
+        this.requestOptions.method = 'POST';
+        this.requestOptions.body = form;
+        let request = new Request('https://api2-page.kakao.com/api/v5/store/singles', this.requestOptions);
```
27136773: Instead of manipulating `this.requestOptions`, create a cloned object and assign changed properties,
e.g.
```javascript
const request = new Request('https://api...', {
   ...this.requestOptions,
    method: 'POST',
    body: form
});
```

src/web/mjs/connectors/kakaopage.mjs
```diff
@@ -0,0 +1,71 @@
...
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let form = new FormData();
+        form.append('productId', chapter.id);
+        this.requestOptions.method = 'POST';
+        this.requestOptions.body = form;
+        let request = new Request('https://api2-page.kakao.com/api/v1/inven/get_download_data/web', this.requestOptions);
```
27136773: See comment above.

src/web/mjs/connectors/kakaopage.mjs
```diff
@@ -0,0 +1,71 @@
...
+        this.requestOptions.method = 'GET';
+        delete this.requestOptions.body;
+        let data = await this.fetchJSON(request);
+        return data.downloadData.members.files.map(element => data.downloadData.members.sAtsServerUrl + element.secureUrl);
```
27136773, line 73: Are these absolute URLs or should they be wrapped in `getAbsolutePath(...)`?

27136773 (approved):

src/web/mjs/connectors/kakaopage.mjs
Robonau: they absolute

PR #3089, opened 2021-02-17
chapter format changeable per connector

Robonau:


Robonau:
there is definitely a better way to do this 
but this was the first thing my brain thought of

Robonau:
any idea why this is happening 
https://i.imgur.com/12Kiwjt.png

src/web/mjs/engine/Connector.mjs
```diff
@@ -743,4 +743,11 @@ export default class Connector {
             return;
         }
     }
+
+    _getFormatRegex() {
```
27136773: Consider making this a getter, so it can be accessed like a property
```javascript
get formatRegex() {
    return // ...
}
```

src/web/mjs/engine/Manga.mjs
```diff
@@ -167,8 +168,8 @@ export default class Manga extends EventTarget {
         }
 
         let name = title;
-        let reVol = /\s*(?:vol\.?|volume|tome)\s*(\d+)/i;
-        let reCh = /\s*(?:^|ch\.?|ep\.?|chapter|chapitre|episode|#)\s*([\d.?\-?v?]+)(?:\s|:|$)+/i; // $ not working in character groups => [\s\:$]+ does not work
+        let reVol = formatRegex.volumeRegex;
+        let reCh = formatRegex.chapterRegex; // $ not working in character groups => [\s\:$]+ does not work
```
27136773: Move or remove comment, since it applies to the certain regex

src/web/mjs/engine/Connector.mjs
27136773: Never mind, i see that this should probably be overwritten by specialisation of this class...

27136773: Remove leading `_`, this has the informal meaning of being private

src/web/mjs/engine/Manga.mjs
```diff
@@ -143,8 +143,9 @@ export default class Manga extends EventTarget {
             .then( chapters => {
                 let chapterFormat = Engine.Settings.chapterTitleFormat.value;
                 // de-serialize chapters into objects
+                let formatRegex = this.connector._getFormatRegex();
```
27136773: May skip local variable, as this is only used once

27136773 (approved):
Looks good, minor imrovments possible

src/web/mjs/engine/Manga.mjs
Robonau: it's used once for every chapter
but ok

27136773:
Can be merged after Lint error has been fixed

PR #3091, opened 2021-02-18
Format chapter title regex

Robonau:
this covers the vast majority of chapter names from the single language connectors

### DIFF
--- a/src/web/mjs/engine/Connector.mjs
+++ b/src/web/mjs/engine/Connector.mjs
@@ -775,6 +775,11 @@ export default class Connector {
                 chapterRegex: /\s*(?:^|ch\.?|ep\.?|chapter|chapitre|Cap[ií]tulo|cap|episode|#):?\s*([\d.?\-?v?]+)(?:\s|:|$)+/i, // $ not working in character groups => [\s\:$]+ does not work
                 volumeRegex: /\s*(?:vol\.?|volume|tome)\s*(\d+)/i
             };
+        } else if (this.tags.includes('polish')) {
+            return {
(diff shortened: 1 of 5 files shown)


src/web/mjs/engine/Connector.mjs
```diff
@@ -745,9 +745,56 @@ export default class Connector {
     }
 
     getFormatRegex() {
-        return{
-            chapterRegex:/\s*(?:^|ch\.?|ep\.?|chapter|chapitre|episode|#)\s*([\d.?\-?v?]+)(?:\s|:|$)+/i, // $ not working in character groups => [\s\:$]+ does not work
-            volumeRegex:/\s*(?:vol\.?|volume|tome)\s*(\d+)/i
-        };
+        if (this.tags.includes('vietnamese')) {
```
27136773: Avoid if-else smell
Either merge this into a single regex, or place each language specific regex into the corresponding connector (object oriented approach)

Robonau: i was trying to avoid
```js
            chapterRegex: /\s*(?:^|ch\.?|ep\.?|chapter|chapitre|Bölüm|Chap|Chương|ตอนที่|Kapitel|Capitolo|Rozdział|Глава|Cap[ií]tulo|cap|episode|#):?\s*([\d.?\-?v?]+)(?:\s|:|$)+/i, // $ not working in character groups => [\s\:$]+ does not work
            volumeRegex: /\s*(?:vol\.?|volume|Sezon|Том|Band|Cilt|tome)\s*(\d+)/i
```
also didn't want to add it to every connector manually

Robonau: if the language was separate from the tags i could use Switch case, but that is not the case

27136773 (approved):
Looks good

PR #3111, opened 2021-02-25
[ComicAction] decrease resource usage

Ever4engel:
decrease resource usage


Robonau:
1.don't spam those PR, do a single one
2.lint
3.given its just coreview just change the coreview template rather than editing each connector individually

PR #3139, opened 2021-02-27
resize-able manga & chapters list

Robonau:
only realised i wanted it when i used #2347
could be added to that in the same way

dosnt work properly when you disable reader
(works fine after a refresh)

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/app.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/app.html
@@ -37,6 +37,8 @@
                 display: none;
             }
             .mangas {
+                resize: horizontal;
+                overflow: auto;
                 display: flex;
                 flex-direction: row;
                 /*width: 100%;*/
@@ -60,12 +62,16 @@
             }
             hakuneko-mangas {
                 box-sizing: border-box;
+                resize: horizontal;
+                overflow: auto;
             }
             hakuneko-chapters {
                 box-sizing: border-box;
+                flex-grow: 1;
             }
             hakuneko-jobs {
                 flex: 0;
+                width: -webkit-fill-available;
             }
             hakuneko-pages {
                 flex: 100;
--- a/src/web/lib/hakuneko/frontend@classic-dark/app.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/app.html
@@ -71,7 +71,7 @@
             }
             hakuneko-jobs {
                 flex: 0;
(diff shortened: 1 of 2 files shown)


src/web/lib/hakuneko/frontend@classic-dark/app.html
```diff
@@ -60,12 +62,16 @@
...
+                resize: horizontal;
+                overflow: auto;
             }
             hakuneko-chapters {
                 box-sizing: border-box;
+                flex-grow: 1;
             }
             hakuneko-jobs {
                 flex: 0;
+                width: -webkit-fill-available;
```
Sheepux: do not use browser specific css prefix, prefer standard fill-available

src/web/lib/hakuneko/frontend@classic-light/app.html
```diff
@@ -61,12 +63,16 @@
...
+                resize: horizontal;
+                overflow: auto;
             }
             hakuneko-chapters {
                 box-sizing: border-box;
+                flex-grow: 1;
             }
             hakuneko-jobs {
                 flex: 0;
+                width: -webkit-fill-available;
```
Sheepux: do not use browser specific css prefix, prefer standard fill-available

Robonau: invalid

Sheepux:
Disabling the reader will ignore de default sizing and therefore creating a blank spacing.
The width should be probably reset tto it's default state in the setting event.
Check in app.html
`             onSettingsSaved(event) {
                this.readerEnabled = Engine.Settings.readerEnabled.value;
            }`
https://github.com/manga-download/hakuneko/blob/master/src/web/lib/hakuneko/frontend%40classic-dark/app.html#L204

Sheepux:
Also, possibly future request: sizing is probably lost on HN restart. Saving it in a local session could allow to restore it on startup.
Also having a "reset to default" option could help.

Robonau:
im not sure how i would save the setting (or even read when it changes) so imma leave it for now

Sheepux:
Finally took time testing it.
Disabling the reader still show the resize option in the manga list, but doesn't work.

Sheepux (changes requested):
As mentioned 1 use case to solve

27136773:
No activity for some weeks => closed
Feel free to re-open in case of continuing the work

PR #3155, opened 2021-03-03
use mobile site for kuaikanmanhua for larger images.

adelie:
**tl;dr.** mobile site gives bigger images. bigger images makes me happy.

**context.** a friend of mine recently tipped me off that kuaikanmanhua serves two different image sizes - 750px wide for the desktop site and 1280px wide for the mobile site. from a scanlation perspective, i prefer high-quality images whenever they're available.

i've tested these changes locally and i believe these are the only changes that need to be made. faking a mobile user-agent will redirect to the mobile site, and then the data structure is slightly different (`comicImages` vs. `comic_images`).

### DIFF
--- a/src/web/mjs/connectors/kuaikanmanhua.mjs
+++ b/src/web/mjs/connectors/kuaikanmanhua.mjs
@@ -78,8 +78,6 @@ export default class Kuaikanmanhua extends Connector {
     }
 
     async _getPageList(manga, chapter, callback) {
-        // clone header, then set mobile user-agent.
-        // mobile site gives larger (1280px wide) images, while desktop site returns smaller (750px) images.
         let newRequestOptions = Object.assign({}, this.requestOptions);
         newRequestOptions.headers.set(
             'x-user-agent',
--- a/src/web/mjs/connectors/kuaikanmanhua.mjs
+++ b/src/web/mjs/connectors/kuaikanmanhua.mjs
@@ -1,7 +1,7 @@
 import Connector from '../engine/Connector.mjs';
 import Manga from '../engine/Manga.mjs';
 
-export default class Kauikanmanhua extends Connector {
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/kuaikanmanhua.mjs
```diff
@@ -1,7 +1,7 @@
 import Connector from '../engine/Connector.mjs';
 import Manga from '../engine/Manga.mjs';
 
-export default class Kauikanmanhua extends Connector {
+export default class Kuaikanmanhua extends Connector {
```
27136773, line 4: Rename file to reflect classname

src/web/mjs/connectors/kuaikanmanhua.mjs
```diff
@@ -78,13 +78,21 @@ export default class Kauikanmanhua extends Connector {
     }
 
     async _getPageList(manga, chapter, callback) {
+        // clone header, then set mobile user-agent.
```
27136773: Redundant comment, this is a well known approach

src/web/mjs/connectors/kuaikanmanhua.mjs
```diff
@@ -78,13 +78,21 @@ export default class Kauikanmanhua extends Connector {
...
+        // mobile site gives larger (1280px wide) images, while desktop site returns smaller (750px) images.
```
27136773: Redundant comment, commit messages should be enough to trace the reason for change

27136773 (approved):
Looks good, minor improvments possible

src/web/mjs/connectors/kuaikanmanhua.mjs
adelie: file is already named correctly - i just fixed the classname to reflect that!

Issue #3212, opened 2021-03-19
[https://mangaspark.com/] Connector not working


satyogi:
https://mangaspark.com/
stop working plz fix

27136773:
Tried to fix in #3213, but due to the CloudFlare bug in electron the URL https://mangaspark.com/manga/_/_/?style=list ends up in the infinite redirect loop...

Can only be solved by #2933

PR #3213, opened 2021-03-19
CloudFlare adjustments

### LINKED ISSUE #3214: [https://mangastarz.com/] Connector not working
https://mangastarz.com/ 
stop working to 
plz fix

27136773:
More and more websites are using a pattern matching against certain search parameters to detect and block HakuNeko requests with CloudFlare firewall. This PR removes all the `rd` and `ts` parameters attached by HakuNeko originally intended to bypass the browser cache. Furthermore the same seems to apply for the `style=list` parameter for WordpressMadara websites, therefore the request will be retry without this parameter (in case of failure).

----

Since this is a major change that affects hundreds of websites, their is a risk that some of them will cease to work...


27136773:
References: #3212

Sheepux (approved):
/LGTM
How about adding the original cache workaround as a third fallback ?

27136773:
Good idea, but to lazy... May consider this when there are a lot of error reports with the current solution

PR #3231, opened 2021-03-23
Changed the Loading Animation at startup

ef1500:
Updater


27136773:
Avoid adding user specific configuration files in `.vs/` to the repository

package.json
```diff
@@ -15,6 +15,7 @@
     "@hakuneko/kindlegen-binaries": "latest",
     "@logtrine/logtrine": "latest",
     "asar": "latest",
+    "babel-core": "^6.26.3",
```
27136773: Do not add unused dependency

package.json
```diff
@@ -26,6 +27,8 @@
     "polymer-build": "latest",
     "puppeteer-core": "latest",
     "rcedit": "latest",
+    "react": "^17.0.1",
```
27136773: Do not add unused dependency

package.json
```diff
@@ -26,6 +27,8 @@
...
+    "react-dom": "^17.0.1",
```
27136773: Do not add unused dependency

src/app/CommandlineArgumentExtractor.js
```diff
@@ -50,6 +50,8 @@ module.exports = class CommandlineArgumentExtractor {
      */
     get options() {
         return {
+            //Add the Cube loader (yet again)
+            applicationCubeURL:           this._get('--cubeloader-url'),
```
27136773: The command line flags are more targeting developers, i don't think it is necessary to make the loading page configurable

src/app/App.js
```diff
@@ -70,7 +63,7 @@ module.exports = class App {
                 delete process.env.HAKUNEKO_PORTABLE;
             }
             await this._electron.launch();
-            await this._electron.loadHTML(loadingPage);
+            await this._electron.loadURL(this._configuration.applicationCubeURL);
```
27136773: I suggest to always load `hakuneko://cache/loader.html` instead of making it configurable

src/app/ElectronBootstrap.js
```diff
@@ -248,12 +248,12 @@ module.exports = class ElectronBootstrap {
         }
 
         this._window = new electron.BrowserWindow({
```
27136773: Keep the default width/height/title until there is a request by users to change them

src/app/ElectronBootstrap.js
```diff
@@ -248,12 +248,12 @@ module.exports = class ElectronBootstrap {
...
-            width: 1120,
-            height: 680,
-            title: 'HakuNeko',
+            width: 800, //1120
+            height: 600, //680
+            titlebarstyle: 'hidden', //oooo borderless window!
```
27136773: The window is already set to borderless in a different place.

src/app/ElectronBootstrap.js
```diff
@@ -248,12 +248,12 @@ module.exports = class ElectronBootstrap {
...
             icon: this._appIcon,
-            show: false,
-            backgroundColor: '#f8f8f8',
+            show: false, //No Idea what the hell this does, I'll leave it alone
+            backgroundColor: '#202225', //The Color change was killing me going into dark mode
```
27136773: If the background color should be changed i suggest a more neutral grey shade, that is independent of any user selected theme

src/app/ElectronBootstrap.js
```diff
@@ -248,12 +248,12 @@ module.exports = class ElectronBootstrap {
...
-            width: 1120,
-            height: 680,
-            title: 'HakuNeko',
+            width: 800, //1120
+            height: 600, //680
+            titlebarstyle: 'hidden', //oooo borderless window!
             icon: this._appIcon,
-            show: false,
-            backgroundColor: '#f8f8f8',
+            show: false, //No Idea what the hell this does, I'll leave it alone
```
27136773: Do not add personal comments

src/web/lib/hakuneko/frontend@classic-dark/cubeloader.html
```diff
@@ -0,0 +1,14 @@
...
+            <div class="top"></div>
+            <div class="right"></div>
+            <div class="bottom"></div>
+            <div class="left"></div>
+            <div class="front"></div>
+            <div class="back"></div>
+        </div>
+    </div>
+    <div class="text">Loading Hakuneko</div>
+</div>
```
27136773: This file seems not to be used, delete unused file

27136773 (changes requested):

src/app/ElectronBootstrap.js
ef1500: Sorry about that, won't happen again.

package.json
```diff
@@ -16,7 +16,7 @@
     "@logtrine/logtrine": "latest",
     "asar": "latest",
     "discord-rpc": "latest",
-    "electron": "8.3.4",
+    "electron": "^7.3.3",
```
27136773: Should keep `8.3.4` if there is no compatibility issue with the new loading screen

src/app/Configuration.js
```diff
@@ -25,6 +25,8 @@ module.exports = class Configuration {
         let applicationExecutableDirectory = path.dirname(electron.app.getPath('exe'));
```
27136773: Can revert changes in this file since the `cubeloader.html` is now statically applied in `App.js`

27136773 (changes requested):

27136773 (commented):
Ready for manual verification by reviewer

27136773:
I tested the changes locally and it worked so far.
However, the loading screen flashes in an instant (when no updates are need to be installed) and therefore no animation could be seen at all. I'm afraid the user might be confused about the short flashing image without enough time to see what it was.

Also the color change from dark grey to the light theme (yes i'm guilty of being a light theme user) is very steep and uncomfortable.

Furthermore, the change requires an updated electron client (nightly build) to be fully working. It is very unlikely that any new clients based on electron will be released in the future.

https://user-images.githubusercontent.com/27136773/112731397-421b7100-8f37-11eb-852f-0661a9943941.mp4

27136773:
I recommend to take a look into the [HaruNeko Prototype Project](https://github.com/manga-download/haruneko/blob/master/src/splash.html) which will probably be the successor of HakuNeko and contribute the splash screen solution for this project.

This also has the advantage for being more experimental, since this is in development and has no active users :)

Sheepux:
Note: also in haruneko it's more than just an updater it's also waiting for the UI to be ready (when UI is sending a ready signal) to display the main window and therefore you won't be getting the second "waiting time" you can see in the video

27136773:
Going to close this for 6.1.7 / Nightly since chances are slim that it would make its way to the users...

PR #3232, opened 2021-03-24
Adding number of retries setting to retry on fetchDOM fail

unluckythoughts:
**Issue**: When getting mangas from large sites like manganel, mangakakalot and others the fetch sometime fails after fetching like 100 or so pages. 
To fix the above issue, adding `Retries` setting to retry on fetch fails in dom. Had to add try/catch as the fetch sometimes fails with null excetion which is not captured by `await`

### DIFF
--- a/src/web/mjs/engine/Connector.mjs
+++ b/src/web/mjs/engine/Connector.mjs
@@ -445,14 +445,15 @@ export default class Connector {
         if((response == null || response.status >= 500) && retries > 0) {
             await this.wait(2500);
             return this.fetchDOM(request, selector, retries - 1);
+        } else if (response != null ) {
+            const content = response.headers.get('content-type');
+            if(response.status === 200 || content.includes('text/html')) {
+                const data = await response.arrayBuffer();
+                const dom = this.createDOM(new TextDecoder(encoding || 'utf8').decode(data));
+                return Promise.resolve(!selector ? dom : [...dom.querySelectorAll(selector)]);
+            }
         }
 
-        const content = response.headers.get('content-type');
-        if(response.status === 200 || content.includes('text/html')) {
-            const data = await response.arrayBuffer();
-            const dom = this.createDOM(new TextDecoder(encoding || 'utf8').decode(data));
-            return Promise.resolve(!selector ? dom : [...dom.querySelectorAll(selector)]);
-        }
         throw new Error(`Failed to receive content from "${request.url}" (type: ${content}, status: ${response.status}) - ${response.statusText}`);
     }
 
--- a/src/web/mjs/engine/Connector.mjs
+++ b/src/web/mjs/engine/Connector.mjs
@@ -445,15 +445,16 @@ export default class Connector {
         if((response == null || response.status >= 500) && retries > 0) {
             await this.wait(2500);
             return this.fetchDOM(request, selector, retries - 1);
-        } else if (response != null ) {
-            const content = response.headers.get('content-type');
-            if(response.status === 200 || content.includes('text/html')) {
-                const data = await response.arrayBuffer();
-                const dom = this.createDOM(new TextDecoder(encoding || 'utf8').decode(data));
-                return Promise.resolve(!selector ? dom : [...dom.querySelectorAll(selector)]);
-            }
+        } else if (response == null ) {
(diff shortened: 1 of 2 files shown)


unluckythoughts:
@ronny1982 the linux connectore2e tests keep failing. Can you review this.

src/web/mjs/engine/Connector.mjs
```diff
@@ -55,6 +57,7 @@ export default class Connector {
      */
     async _initializeConnector() {
         let uri = new URL(this.url);
+        this.retries = parseInt(Engine.Settings.retries.value) || this.retries;
```
Robonau, line 60:
```js
parseInt(Engine.Settings.retries.value)
```
should really just be in the fetches (fetchDOM, fetchJSON, etc)
as it is right now you can't change the setting after the connector initialize
(you would replace `this.retries` with `parseInt(Engine.Settings.retries.value)`)

src/web/mjs/engine/Connector.mjs
```diff
@@ -423,19 +426,29 @@ export default class Connector {
      * and get all elements matching the given CSS selector.
      */
     async fetchDOM(request, selector, retries, encoding) {
-        retries = retries || 0;
+        retries = retries || this.retries;
```
Robonau, line 429: https://github.com/manga-download/hakuneko/pull/3232#discussion_r600717609
here

src/web/mjs/engine/Connector.mjs
```diff
@@ -449,7 +462,7 @@ export default class Connector {
      *
      */
     fetchJSON( request, retries ) {
-        retries = retries || 0;
+        retries = retries || this.retries;
```
Robonau, line 465: https://github.com/manga-download/hakuneko/pull/3232#discussion_r600717609
and here

src/web/mjs/engine/Settings.mjs
```diff
@@ -93,6 +93,13 @@ export default class Settings extends EventTarget {
             value: false
         };
 
+        this.retries = {
+            label: 'Retries',
+            description: 'No of times to retry if the site fetch failes',
```
Robonau, line 98: *fails

Robonau: and since the user will be reading this setting it might be worth to not use `no` as shorthand for `number` (non english users ...)

src/web/mjs/engine/Connector.mjs
```diff
@@ -423,19 +426,29 @@ export default class Connector {
...
         }
         // TODO: check if this will affect (replace) the input parameter?
         if(request instanceof URL) {
             request = new Request(request.href, this.requestOptions);
         }
-        const response = await fetch(request.clone());
-        if(response.status >= 500 && retries > 0) {
+
+        let response = null;
+        try {
```
27136773, line 439: Do not use `try ... catch ...`, fail fast.
There are hundreds of system errors were a retry makes no sense (`OUT_OF_MEM`, `E_CONN_RESET`, `E_DISCONNECTED`, ...)

src/web/mjs/engine/Connector.mjs
27136773: I second this, retrying should be the exception and not the rule. I suggest to point out what connectors have problems, analyze the cause and just add the retry to where it is acceptable.

src/web/mjs/engine/Settings.mjs
```diff
@@ -93,6 +93,13 @@ export default class Settings extends EventTarget {
...
+            input: types.numeric,
```
27136773, line 99: Restrict range, the ability to set `-312434143` or `3421446354` may confuse the user

27136773 (commented):
While the idea of configurable retry is interesting, i think it should be left as is right now and focus should be to investigate the websites that have such errors. Maybe add throttling and retries only for the affected website...

27136773:
> @ronny1982 the linux connectore2e tests keep failing. Can you review this.

These are probably the UI integration tests with puppeteer. These are not easy to fix since the reason for failure is not quite clear :(

27136773:
I will reject this PR for now, if retry became a serious issue, we may discuss this on discord first on how to handle this

Issue #3208, opened 2021-03-17
[Bug] Kakao page


Ever4engel:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
YES

**Describe the bug**
can't get list of available chapters for some webtoons (i found one). and, can't login in hakuneko in kakao's connector.

**To Reproduce**
1. copy this link : https://page.kakao.com/home/%EC%B2%9C%EC%9E%AC-%EC%9D%98%EC%82%AC-%EC%9D%B4%EB%AC%B4%EC%A7%84/56329995
2. use clipboard 
3. click on the webtoon to get chapterlist
4. no chapters (in the browser there is 28 chapters)

**Expected behavior**
i get chapterlist for the first bug. second bug i can't login.

**Context (please complete the following information):**
 - OS: windows 10 20H2
 - HakuNeko Version 8.3.4/6.1.7

ap-sensei:
I am having the same problem too
I can see the webpage in my browser, however, in HakuNeko, it says
"Connection Timed Out"

XenonR:
Closed by merging #3239

Ever4engel:
@XenonR you done an error about this ticket. he's related about other bugs (the login bug)

MikeZeDev:
Tested, fixed

PR #3239, opened 2021-03-26
kakaopage fix

x4nk1t:


src/web/mjs/connectors/kakaopage.mjs
```diff
@@ -19,6 +19,7 @@ export default class kakaopage extends Connector {
         let id = uri.searchParams.get('seriesId');
```
27136773: May use short circuiting
```javascript
let id = uri.searchParams.get('seriesId') || uri.pathname.split('/').filter(part => part).pop();
```

27136773 (changes requested):
Fix lint errors before merge.
Optional: Minor improvements possible.

27136773 (approved):
Looks good

x4nk1t:
Closes #3208

PR #3254, opened 2021-03-31
Improve Discord Presence

### LINKED ISSUE #2022: [Bug] Silent error in discord presence
No

**Describe the bug**
Discord presence throws an uncaught error when transitioning from `start` to `stop`

**To Reproduce**
1. Start HakuNeko (nightly), press press F12 to show error console tab
2. Open settings and change discord presence by applying changes in the following order:
`No presence` > `All but hentai` > `No presence`
3. See error when transitioning from enabled to disabled

**Expected behavior**
No error should be shown

 - OS: any
 - HakuNeko Nightly or Development (master)

**Screenshots**
image

XenonR:
This tries to alleviate the bug #2022
Unfortunately I'm unable to catch the last edge case. But even considering that, it is better than before.
Closes #2022


src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -101,14 +101,16 @@ export default class DiscordPresence {
         }
     }
 
-    stopDiscordPresence() {
+    stopDiscordPresence(reset = false) {
```
27136773: May split into two methods instead of introducing flag parameters, e.g. `stopDiscordPresenceGracefully()` and `terminateDiscordPresence()` or similar

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -101,14 +101,16 @@ export default class DiscordPresence {
...
-        if (this.rpc) {
+        if (this.rpc && !reset) {
             this.rpc.clearActivity();
             this.rpc.destroy();
-            this.rpc = null;
         }
+        this.rpc = null;
     }
 
     async startDiscordPresence() {
```
27136773, line 114: Method complexity is growing, consider outsourcing conditional blocks to separate methods

src/web/mjs/engine/DiscordPresence.mjs
```diff
@@ -124,20 +126,48 @@ export default class DiscordPresence {
...
+                        // Re-evaluate if enabled
+                        this._onSettingsChanged();
+                    }, 120000);
+
+                    return;
                 }
-                this.rpc = null;
-            });
+
+                throw new Error(error); // Unknown error
```
27136773: Since `error` is expected to be of type `Error` it can be re-thrown directly instead of wrapping into another error, e.g. `throw error`

27136773 (approved):

src/web/mjs/engine/DiscordPresence.mjs
XenonR: will keep it mind for the future. for now: won't change

27136773 (approved):
Looks good

PR #3264, opened 2021-04-02
Added kumamanga support

### LINKED ISSUE #3259: [Site Request] kumamanga
**Name of the website**
kumamanga

- Site: https://kumamanga.my.id/
- Manga List: https://kumamanga.my.id/project-kuma/
- Manga example: https://kumamanga.my.id/series/made-in-abyss/
- Chapter online viewer example: https://kumamanga.my.id/made-in-abyss-chapter-50-bahasa-indonesia/

**Languages**
Indonesia

sisco0:
KumaManga icon and connector were added. This pull request closes https://github.com/manga-download/hakuneko/issues/3259


src/web/mjs/connectors/KumaManga.mjs
```diff
@@ -0,0 +1,39 @@
...
+        super.id = 'kumamanga';
+        super.label = 'KumaManga';
+        this.tags = [ 'manga', 'indonesian' ];
+        this.url = 'https://kumamanga.my.id';
+        this.path = '/manga-list/?list';
+
+        this.querMangaTitleFromURI = 'div.series-title h2';
+        this.queryMangas = 'div.mangalist-blc ul li a';
+        this.queryChapters = 'div.series-chapter ul.series-chapterlist li div.flexch-infoz a';
+        this.queryChaptersTitle = 'span';
```
27136773, line 16: How about setting `this.queryChaptersTitleBloat = 'span > span.date'` to remove the date from the chapter title?
This may avoid re-implementation of the `_getChapters()` method.

See also: [MangaKane](https://github.com/manga-download/hakuneko/blob/f31ea8f2c95bc373b28503549c0e12d3ee15bd46/src/web/mjs/connectors/MangaKane.mjs#L17)

src/web/mjs/connectors/KumaManga.mjs
```diff
@@ -0,0 +1,39 @@
...
+        this.queryPages = 'div.reader-area img[src]:not([src=""])';
+    }
+
+    async _getChapters(manga) {
```
27136773: Method may be obsolete when utilizing `queryChaptersTitleBloat `

27136773 (commented):
Looks good.
Maybe it is possible to simplify the implementation (need to be verified by contributor)

sisco0:
Hello, @ronny1982 I would take care of the feedback and would update the pull request by today. I think I need to add a test case as well.
I have seen we have the `assertConnector` which could be used for getting the project more consistent between versions and detecting updated pages as well.
Hopefully, something could be done on this in a separate issue, so we could automatically generate tests for the whole `mjs` folder files.

src/web/mjs/connectors/KumaManga.mjs
sisco0: I modified that code in this new commit for accomplishing the tasks of getting rid of the date field more shortly, as advised.

27136773:
There are problems with Puppeteer in GitHub Actions, so browser based integration tests are not running at all => Not worth the effort to write tests :(

You can directly reply and resolve the [reviewers comment](https://github.com/manga-download/hakuneko/pull/3264#discussion_r606157465) after pushing updates or explain and rejecting the proposal. The reviewer will be notified automatically.

----

image

src/web/mjs/connectors/KumaManga.mjs
sisco0: Updated with the simplified solution proposed above, so no method is overridden anymore.

sisco0:
I could continue contributing to this interesting project. I have used `puppeteer` in the past under _Github Actions_ without inconvenience. Hopefully, that could be the case for this project, so I would review the current _Jest_ infrastructure and see what could be done.

27136773 (approved):
Looks good

PR #3313, opened 2021-04-10
[hentaimoney] added website support

27136773:
References: dazedcat19/FMD2#491


Sheepux (approved):
I was so surprised with the getChapters code (chapter=manga) that i had to check the website itself to understand wtf was happening :D
/LGTM

27136773:
Yeah, many of the hentai websites don't use the concept of manga/chapter

PR #3328, opened 2021-04-11
[xcalibrscans] added website support

27136773:
References: dazedcat19/FMD2#496


src/web/mjs/connectors/XCaliBRScans.mjs
```diff
@@ -0,0 +1,24 @@
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
+
+export default class XCaliBRScans extends WordPressMangastream {
+
+    constructor() {
+        super();
+        super.id = 'xcalibrscans';
+        super.label = 'xCaliBR Scans';
```
Sheepux, line 8: label capitalization difference with classname or something intended ?

Sheepux (approved):
/LGTM 
minor capitalization double check

src/web/mjs/connectors/XCaliBRScans.mjs
27136773: Intentional.
Class names are supposed to be PascalCase, but the label is spelled the same way as the scanlation group does on their website.

Issue #3330, opened 2021-04-12
[Sekaikomik] Connector not working


manhwalover:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes/

**Describe the bug**
Chapter list is empty. Because Cloudflare not success clear captcha

**To Reproduce**
1. Name of connector: Sekaikomik 
2. Name of manga: All
3. Name of chapter: All

**Screenshots**
![image](https://user-images.githubusercontent.com/79884368/114360772-97928780-9b9f-11eb-8919-d86b34ecd17f.png)

**HakuNeko (please complete the following information):**
 - Version 6.17 stable

27136773:
Handled by #2933

Issue #2933, opened 2021-01-16
[Bug] CloudFlare Challenge get stuck in Infinite Loop


27136773:
## Describe the bug
When opening any website that is protected through a CloudFlare JavaScript challenge, the website is stuck in an infinite reload loop instead of solving the challenge and redirecting to the content of the website.

This affects many websites right now that can no longer be used in HakuNeko due to this issue!

### With HakuNeko
1. Start HakuNeko
2. Open the Website choosing List
3. Search for `raws.mangazuki.co`
4. Open the manual website interaction link on the result
  => Observe the website re-loads forever
### With Electron
1. Start electron `./node_modules/.bin/electron`
2. Open the developer tools (e.g. from electron menu)
3. Switch to the console tab and type `window.location = 'https://cloudscraper.cf/cloudflare/challenge.html'`
  => Observe the website re-loads forever

**References**
- #2885

27136773:
Looks like CloudFlare revised their challenge and HakuNeko now ends up with infinite redirects...

----

```html
<!DOCTYPE HTML>
<html lang="en-US">
<head>
  <meta charset="UTF-8" />
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
  <meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1" />
  <meta name="robots" content="noindex, nofollow" />
  <meta name="viewport" content="width=device-width,initial-scale=1" />
  <title>Just a moment...</title>
  <style type="text/css">
    html, body {width: 100%; height: 100%; margin: 0; padding: 0;}
    body {background-color: #ffffff; color: #000000; font-family:-apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, "Helvetica Neue",Arial, sans-serif; font-size: 16px; line-height: 1.7em;-webkit-font-smoothing: antialiased;}
    h1 { text-align: center; font-weight:700; margin: 16px 0; font-size: 32px; color:#000000; line-height: 1.25;}
    p {font-size: 20px; font-weight: 400; margin: 8px 0;}
    p, .attribution, {text-align: center;}
    #spinner {margin: 0 auto 30px auto; display: block;}
    .attribution {margin-top: 32px;}
    @keyframes fader     { 0% {opacity: 0.2;} 50% {opacity: 1.0;} 100% {opacity: 0.2;} }
[...]

PR #3331, opened 2021-04-12
[sekaikomik] initialize all CloudFlare protected paths

27136773:
References: #3330 
This PR will only add CloudFlare initialization for the `/manga/` URL pattern, loading chapters will still fail due to #2933


XenonR (approved):
All good.

PR #3474, opened 2021-05-20
Support for purchased Volumes on Viz

playmer:
I'm sure there's some additional error handling that needs to be done/moved around, but I think by and large the functionality is there. The approach is likely a bit naïve however, I'm not remotely fluent in JS, so let me know any suggestions to clean it up.

### DIFF
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -161,7 +161,7 @@ export default class VizShonenJump extends Connector {
         let mangaID = 0;
 
         if (chapter.id.startsWith("read/manga")) {
-            pageCount = parseInt([...responseData.matchAll(/<div class="mar-b-md"><strong>Length<\/strong> (\d+) pages<\/div>/g)][0][1]);
+            pageCount = parseInt([...responseData.matchAll(/<div\s+class="mar-b-md">\s*<strong>\s*Length\s*<\/strong>\s+(\d+)\s+pages\s*<\/div>/g)][0][1]);
             mangaID = parseInt([...responseData.matchAll(/var\s+mangaCommonId\s*=\s*(\d+)/g)][0][1]);
         } else if (chapter.id.startsWith("/shonenjump")) {
             pageCount = parseInt([...responseData.matchAll(/var\s+pages\s*=\s*(\d+)/g)][0][1]);
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -177,13 +177,10 @@ export default class VizShonenJump extends Connector {
         let mangaID = 0;
 
         if (chapter.id.startsWith("read/manga")) {
-            const pageCountData = await this.requestRegex(responseData, /<div class="mar-b-md"><strong>Length<\/strong> (\d+) pages<\/div>/);
-            pageCount = parseInt(pageCountData[0]);
-            const mangaIdData = await this.requestRegex(responseData, /var\s+mangaCommonId\s*=\s*(\d+)/);
-            mangaID = parseInt(mangaIdData[0]);
+            pageCount = parseInt([...responseData.matchAll(/<div class="mar-b-md"><strong>Length<\/strong> (\d+) pages<\/div>/g)][0][1]);
+            mangaID = parseInt([...responseData.matchAll(/var\s+mangaCommonId\s*=\s*(\d+)/g)][0][1]);
         } else if (chapter.id.startsWith("/shonenjump")) {
-            const data = await this.requestRegex(responseData, /var\s+pages\s*=\s*(\d+)/);
-            pageCount = parseInt(data[0]);
+            pageCount = parseInt([...responseData.matchAll(/var\s+pages\s*=\s*(\d+)/g)][0][1]);
             mangaID = chapter.id.match(/chapter\/(\d+)/)[1];
         }
 
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -57,8 +57,6 @@ export default class VizShonenJump extends Connector {
             return;
         }
 
-        data = [...data.querySelectorAll('a.o_chapters-link')];
-
         data = [...data.querySelectorAll('div.o_sort_container div.o_sortable a.o_chapters-link')];
 
         return data.map(manga => {
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -158,6 +158,8 @@ export default class VizShonenJump extends Connector {
         if (manga.id.startsWith("/library")) {
             return await this._getMangaVolumes(manga);
         }
+        
+        throw new Error(`Failed to get chapters/volumes for manga ${manga.id}, because the manga type is not supported!`);
     }
 
     async requestRegex(responseData, regex) {
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -68,13 +68,7 @@ export default class VizShonenJump extends Connector {
     }
 
     async _getMangas() {
-        const mangaAvailibleByVolumes = await this._getMangasAvailibleByVolumes();
-        const mangaAvailibleByChapters = await this._getMangasAvailibleByChapters();
-        let both = mangaAvailibleByVolumes.concat(mangaAvailibleByChapters);
-
-        both.sort((a, b) => a.title > b.title && a.id > b.id );
-
-        return both;
+        return [ ...await this._getMangasAvailibleByVolumes(), ...await this._getMangasAvailibleByChapters() ];
     }
 
     async _getMangaChapters(manga) {
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -34,7 +34,21 @@ export default class VizShonenJump extends Connector {
         };
     }
 
-    async _getMangas() {
+    async _getMangasAvailibleByVolumes() {
+        const request = new Request(new URL('/library', this.url), this.requestOptions);
+        let data = await this.fetchDOM(request);
+
+        data = [...data.querySelectorAll('table.purchase-table a')];
+
+        return data.map(manga => {
+            return {
+                id: manga.pathname,
+                title: manga.innerText.trim()
+            };
+        });
+    }
+
+    async _getMangasAvailibleByChapters() {
         const request = new Request(new URL('/shonenjump', this.url), this.requestOptions);
         let data = await this.fetchDOM(request);
 
@@ -43,17 +57,29 @@ export default class VizShonenJump extends Connector {
             return;
         }
 
+        data = [...data.querySelectorAll('a.o_chapters-link')];
+
         data = [...data.querySelectorAll('div.o_sort_container div.o_sortable a.o_chapters-link')];
 
         return data.map(manga => {
             return {
                 id: manga.pathname,
-                title: manga.querySelector(':nth-child(2)').innerText.trim()
+                title: manga.innerText.trim()
             };
         });
     }
 
-    async _getChapters(manga) {
+    async _getMangas() {
+        const mangaAvailibleByVolumes = await this._getMangasAvailibleByVolumes();
+        const mangaAvailibleByChapters = await this._getMangasAvailibleByChapters();
+        let both = mangaAvailibleByVolumes.concat(mangaAvailibleByChapters);
+
+        both.sort((a, b) => a.title > b.title && a.id > b.id );
+
+        return both;
+    }
+
+    async _getMangaChapters(manga) {
         let auth = await this._getUserInfo();
         const request = new Request(new URL(manga.id, this.url), this.requestOptions);
         let data = await this.fetchDOM(request);
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
             });
     }
 
+    async _getMangaVolumes(manga) {
+        let auth = await this._getUserInfo();
+        const request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request);
+
+        return [...data.querySelectorAll('#o_products tr td:last-of-type a')]
+            .filter(element => {
+                // login required (e.g. age verification) => javascript:void('login to read');
+                if(/(?=javascript:)(?=.*login)/i.test(element.href)) {
+                    return auth.isLoggedIn && auth.isAdult;
+                }
+                // subscription required => javascript:void('join to read');
+                if(/(?=javascript:)(?=.*join)/i.test(element.href)) {
+                    return auth.isMember;
+                }
+                // free
+                return true;
+            })
+            .map(chapter => {
+                return {
+                    id: chapter.href.replace("hakuneko://cache/", ""),
+                    title: 'Vol. ' + chapter.href.match(/-volume-([-_0-9]+)/)[1]
+                };
+            });
+    }
+
+    async _getChapters(manga) {
+        if (manga.id.startsWith("/shonenjump/chapters")) {
+            return await this._getMangaChapters(manga);
+        } else if (manga.id.startsWith("/library")) {
+            return await this._getMangaVolumes(manga);
+        }
+    }
+
+    async requestRegex(responseData, regex) {
+        let regex2 = new RegExp(regex.source, regex.flags + "g");
+        let result = [];
+        let match = undefined;
+        // eslint-disable-next-line no-cond-assign
+        while(match = regex2.exec(responseData)) {
+            result.push(match[1]);
+        }
+
+        return result;
+    }
+
     async _getPages(chapter) {
         const request = new Request(new URL(chapter.id, this.url));
-        const data = await this.fetchRegex(request, /var\s+pages\s*=\s*(\d+)/g);
-        const pageCount = parseInt(data[0]);
-        const chapterID = chapter.id.match(/chapter\/(\d+)/)[1];
+        let response = await fetch(request);
+        let responseData = await response.text();
+
+        let pageCount = 0;
+        let mangaID = 0;
+
+        if (chapter.id.startsWith("read/manga")) {
+            const pageCountData = await this.requestRegex(responseData, /<div class="mar-b-md"><strong>Length<\/strong> (\d+) pages<\/div>/);
+            pageCount = parseInt(pageCountData[0]);
+            const mangaIdData = await this.requestRegex(responseData, /var\s+mangaCommonId\s*=\s*(\d+)/);
+            mangaID = parseInt(mangaIdData[0]);
+        } else if (chapter.id.startsWith("/shonenjump")) {
+            const data = await this.requestRegex(responseData, /var\s+pages\s*=\s*(\d+)/);
+            pageCount = parseInt(data[0]);
+            mangaID = chapter.id.match(/chapter\/(\d+)/)[1];
+        }
 
         return Array(pageCount+1).fill().map((_, index) => {
             let page = new URL('/manga/get_manga_url', this.url);
             page.searchParams.set('device_id', 3);
-            page.searchParams.set('manga_id', chapterID);
+            page.searchParams.set('manga_id', mangaID);
             page.searchParams.set('page', index);
             return this.createConnectorURI({ id: page.href, referer: chapter.id });
         });
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -84,11 +84,11 @@ export default class VizShonenJump extends Connector {
         return [...data.querySelectorAll('div > a.o_chapter-container[data-target-url], tr.o_chapter td.ch-num-list-spacing a.o_chapter-container[data-target-url]')]
             .filter(element => {
                 // login required (e.g. age verification) => javascript:void('login to read');
-                if(/(?=javascript:)(?=.*login)/i.test(element.href)) {
+                if(/javascript:.*login/i.test(element.href)) {
                     return auth.isLoggedIn && auth.isAdult;
                 }
                 // subscription required => javascript:void('join to read');
-                if(/(?=javascript:)(?=.*join)/i.test(element.href)) {
+                if(/javascript:.*join/i.test(element.href)) {
                     return auth.isMember;
                 }
                 // free
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -120,9 +120,8 @@ export default class VizShonenJump extends Connector {
     async _getMangaVolumes(manga) {
         let auth = await this._getUserInfo();
         const request = new Request(new URL(manga.id, this.url), this.requestOptions);
-        let data = await this.fetchDOM(request);
 
-        return [...data.querySelectorAll('#o_products tr td:last-of-type a')]
+        return [...await this.fetchDOM(request, '#o_products tr td:last-of-type a')]
             .filter(element => {
                 // login required (e.g. age verification) => javascript:void('login to read');
                 if(/(?=javascript:)(?=.*login)/i.test(element.href)) {
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -154,7 +154,8 @@ export default class VizShonenJump extends Connector {
     async _getChapters(manga) {
         if (manga.id.startsWith("/shonenjump/chapters")) {
             return await this._getMangaChapters(manga);
-        } else if (manga.id.startsWith("/library")) {
+        } 
+        if (manga.id.startsWith("/library")) {
             return await this._getMangaVolumes(manga);
         }
     }
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -37,7 +37,7 @@ export default class VizShonenJump extends Connector {
     async _getMangasAvailibleByVolumes() {
         const request = new Request(new URL('/library', this.url), this.requestOptions);
         let data = await this.fetchDOM(request);
-        
+
         if ( data.innerText.includes('Log in to view your library') ) { // User isn't logged in, so there's no availible volumes
             return [];
         }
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -34,26 +36,87 @@ export default class VizShonenJump extends Connector {
...
         });
     }
 
-    async _getChapters(manga) {
+    async _getMangas() {
+        const mangaAvailibleByVolumes = await this._getMangasAvailibleByVolumes();
+        var mangaAvailibleByChapters = await this._getMangasAvailibleByChapters();
+
+        mangaAvailibleByVolumes.forEach(mangaAvailibleByVolume => {
+                const index = mangaAvailibleByChapters.findIndex(element => element.title === mangaAvailibleByVolume.title);
```
Robonau: can just use find and not findindex
```js
            let element = mangaAvailibleByChapters.find(element => element.title === mangaAvailibleByVolume.title);
            // Found it.
            if (element != undefined) {
                var complexManga = JSON.parse(element.id);

                element = {
                    ...
                };
            }
```

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -34,26 +36,87 @@ export default class VizShonenJump extends Connector {
...
+                }),
+                title: manga.innerText.trim(),
             };
         });
     }
 
-    async _getChapters(manga) {
+    async _getMangas() {
+        const mangaAvailibleByVolumes = await this._getMangasAvailibleByVolumes();
+        var mangaAvailibleByChapters = await this._getMangasAvailibleByChapters();
```
Robonau: `let` is better than `var` (this goes for every time you used `var`)

Robonau:
lint (can wait till its done) (i suggest vscode + eslint plugin, if you are gonna keep working on HN (or with js in general))

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -10,6 +11,7 @@ export default class VizShonenJump extends Connector {
         this.tags = [ 'manga', 'english' ];
         this.url = 'https://www.viz.com';
         this.requestOptions.headers.set( 'x-referer', this.url );
+        this.EntryTypeEnum = Object.freeze({"volume":1, "chapter":2});
```
Robonau: is there a reason for the `Object.freeze`, i don't see it, nor can i really test it

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -34,26 +36,87 @@ export default class VizShonenJump extends Connector {
...
+                        title: mangaAvailibleByVolume.title,
+                    });
+                }
+            })
+
+        return mangaAvailibleByChapters;
+    }
+
+    async _getMangaChapters(manga) {
+        if ((manga.entry_type & this.EntryTypeEnum.chapter) != this.EntryTypeEnum.chapter) {
```
Robonau: `if (manga.entry_type  != this.EntryTypeEnum.chapter)`
why use bitwise operators when you don't need to?

actually just use Switch case Statement in _getChapters

actually aren't you trying get from both volumes and chapters?
as the `return chapters.concat(volumes);` in _getChapters would suggest
you currently just get from one or the other

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -34,26 +36,87 @@ export default class VizShonenJump extends Connector {
...
+            console.log(`Exiting from _getMangaChapters ${JSON.stringify(manga)}`);
```
Robonau: _getMangaVolumes?
don't think you need to stringify either

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +164,79 @@ export default class VizShonenJump extends Connector {
             });
     }
 
+    async _getMangaVolumes(manga) {
+        if ((manga.entry_type & this.EntryTypeEnum.volume) != this.EntryTypeEnum.volume) {
```
Robonau, line 168: `if (manga.volume != this.EntryTypeEnum.volume)`

Robonau:
all of what i said could be total rubbish (other than the lint comment), gotta wait for ronny to say really
ronny will definitely say more than i did

src/web/mjs/connectors/VizShonenJump.mjs
playmer: This is used as a Flag enum, so it's not one or the other, it could be both. A user might have manga chapters available via SJ, as well as volumes they've purchased from Viz. So a switch case doesn't work here, as if both flags were set, the enums value would be `3` and neither condition would be hit.

Thus we union on the values returned from both functions, but sometimes one or the other might be empty.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: I'm not sure, I just looked up how to do enums in JS and this was recommended. From the docs, it seems like this carries a level of immutability with it, so EntryTypeEnum can't be accidentally modified.: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

src/web/mjs/connectors/VizShonenJump.mjs
playmer: Thanks!

src/web/mjs/connectors/VizShonenJump.mjs
playmer: I'll give this a try. I wasn't sure of the semantics allowing me to modify the object inside the array with find, appreciate the tip!

src/web/mjs/connectors/VizShonenJump.mjs
playmer: This was just for testing purposes, I'm just going to remove it. The stringification was from trying to get something more than like `[Object object]` to print out.

src/web/mjs/connectors/VizShonenJump.mjs
Robonau: yup this was just be being kinda dumb
`[1,3].includes(this.EntryTypeEnum.chapter)` (or [2,3] )
is probably more readable, using bitwise is rather uncommon in js

or just `this.EntryTypeEnum.chapter != 2` (or 1 depending on vol vs chapters)

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -34,26 +36,86 @@ export default class VizShonenJump extends Connector {
...
+                // Found it.
+                if (element != undefined) {
+                    let complexManga = JSON.parse(element.id);
+
+                    element = {
+                        id: JSON.stringify({
+                            id: complexManga.id,
+                            volume_id: mangaAvailibleByVolume.volume_id,
+                            title: mangaAvailibleByVolume.title,
+                            entry_type: this.EntryTypeEnum.volume | this.EntryTypeEnum.chapter
```
Robonau: why | and not + ?
(this is what threw me off on the bitwize comment)
`this.EntryTypeEnum.volume + this.EntryTypeEnum.chapter` (or just `3`)
again, bitwise is not common in js

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -119,6 +244,7 @@ export default class VizShonenJump extends Connector {
     async _handleConnectorURI(payload) {
         try {
             // API request
+            const testOrigin = new URL(this.url).origin
```
Robonau: why?

27136773:
After some investigation here are my thought.

Personally i would go for a different approach and keep the manga IDs as simple as possible:
- The chapter based mangas from the [/shonenjump](https://www.viz.com/shonenjump) endpoint have relative URLs similar to `/shonenjump/chapters/dr-stone`
- The volume based mangas from the [/library](https://www.viz.com/library) endpoint have relative URLs similar to `/read/manga/aion-volume-1/product/4681/digital?action=read`

This makes them perfectly distinguishable just by the first two segments in the path (`/shonenjump/chapters` vs `/read/manga`).

This will lead to manga entries in the list to e.g.
```javascript
const mangaList = [
    {
        id: '/shonenjump/chapters/dr-stone',
        title = 'Dr. Stone'
    },
    {
        id: '/read/manga/dr-stone-volume-1/product/0000/digital?action=read',
        title = 'Dr. Stone, Vol. 1'
    },
    {
        id: '/read/manga/dr-stone-volume-7/product/0000/digital?action=read',
        title = 'Dr. Stone, Vol. 7'
    }
];
```

In `_getChapters(manga)` the mangas can be handled differently depending if they are chapter based (id starting with `/shonenjump/chapters`), or if they are volume based (id starting with `/read/manga`).

For volume based mangas the volume is already the chapter, so it can simply be passed through with a new [assigned object](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/NHentai.mjs#L30-L32) or a [spread object](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/HentaiFox.mjs#L28-L30). This check could be added [right at the start](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/VizShonenJump.mjs#L56) of the existing chapter method.

For chapter based mangas the [existing method](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/VizShonenJump.mjs#L56-L102) can be used to get the chapter list. 

----

I agree that this is not optimal from a usability point of view, but it makes the implementation very easy and since the readers are probably already used to the difference between the free serialized chapters and the purchased volumes from the website, i think they will understand the difference.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: Superfluous, again for testing, will remove, thanks!

playmer:
Hmm, that makes sense, so you're saying that, if you owned volumes in a manga, as well as had access to chapters of that manga, you'd add two entries to the Manga list? One would be something like `${MangaName} Chapters` and the other `${MangaName} Volumes`?

27136773:
> Hmm, that makes sense, so you're saying that, if you owned volumes in a manga, as well as had access to chapters of that manga, you'd add two entries to the Manga list? One would be something like `${MangaName} Chapters` and the other `${MangaName} Volumes`?

Yes, the manga titles shown in the `/library` and in `/shonenjump` are already fine.
The user can choose to download volumes, single chapters or both.

playmer:
Gotcha gotcha, I'll work on that later today/tomorrow then! That obviates the need of the bitwise stuff too, so less confusion will happen thankfully. Thanks for the feedback!

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
-        const pageCount = parseInt(data[0]);
-        const chapterID = chapter.id.match(/chapter\/(\d+)/)[1];
+        let response = await fetch(request);
+        let responseData = await response.text();
+
+        let pageCount = 0;
+        let mangaID = 0;
+
+        if (chapter.id.startsWith("read/manga")) {
+            const pageCountData = await this.requestRegex(responseData, /<div class="mar-b-md"><strong>Length<\/strong> (\d+) pages<\/div>/);
```
27136773: The [slightest change with whitespaces](https://regex101.com/r/YflMBl/1/) (linebreak, space) will cause this to fail, isn't there a more reliable way to get the page count?

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+            pageCount = parseInt(pageCountData[0]);
+            const mangaIdData = await this.requestRegex(responseData, /var\s+mangaCommonId\s*=\s*(\d+)/);
```
27136773: May be simplified
```javascript
responseData.match(/var\s+mangaCommonId\s*=\s*(\d+)/)[1]
```

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+
+    async _getChapters(manga) {
+        if (manga.id.startsWith("/shonenjump/chapters")) {
+            return await this._getMangaChapters(manga);
+        } else if (manga.id.startsWith("/library")) {
+            return await this._getMangaVolumes(manga);
+        }
+    }
+
+    async requestRegex(responseData, regex) {
```
27136773: The `async` is obsolete, since there is no `await` anywhere in this method.
Method name contains the phrase `request`, which may be misleading since there is no request performed.
Method seems redundant in general, it can probably be removed in favor of `match`.

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -34,7 +34,21 @@ export default class VizShonenJump extends Connector {
         };
     }
 
-    async _getMangas() {
+    async _getMangasAvailibleByVolumes() {
+        const request = new Request(new URL('/library', this.url), this.requestOptions);
+        let data = await this.fetchDOM(request);
```
27136773, line 39: Consider using the CSS query directly `this.fetchDOM(request, 'table.purchase-table a')`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -34,7 +34,21 @@ export default class VizShonenJump extends Connector {
...
+            return {
+                id: manga.pathname,
+                title: manga.innerText.trim()
+            };
+        });
+    }
+
+    async _getMangasAvailibleByChapters() {
         const request = new Request(new URL('/shonenjump', this.url), this.requestOptions);
         let data = await this.fetchDOM(request);
```
27136773, line 57: Consider using the CSS query directly `this.fetchDOM(request, 'div.o_sort_container div.o_sortable a.o_chapters-link')`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -43,17 +57,29 @@ export default class VizShonenJump extends Connector {
             return;
         }
 
+        data = [...data.querySelectorAll('a.o_chapters-link')];
+
         data = [...data.querySelectorAll('div.o_sort_container div.o_sortable a.o_chapters-link')];
```
27136773: The method `querySelectorAll` should not be used on an array (`data` is an array at this point, see previous assignment)

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -43,17 +57,29 @@ export default class VizShonenJump extends Connector {
...
+                title: manga.innerText.trim()
             };
         });
     }
 
-    async _getChapters(manga) {
+    async _getMangas() {
+        const mangaAvailibleByVolumes = await this._getMangasAvailibleByVolumes();
+        const mangaAvailibleByChapters = await this._getMangasAvailibleByChapters();
+        let both = mangaAvailibleByVolumes.concat(mangaAvailibleByChapters);
```
27136773: May be simplified with `spread` operator.
Also the sort by title is not required if i remember correctly as this is already done during further processing of the manga list.
```javascript
return [ ...mangaAvailibleByVolumes, ...mangaAvailibleByChapters ];
```

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
             });
     }
 
+    async _getMangaVolumes(manga) {
+        let auth = await this._getUserInfo();
+        const request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request);
```
27136773, line 128: Consider using the CSS query directly `this.fetchDOM(request, '#o_products tr td:last-of-type a')`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+
+        return [...data.querySelectorAll('#o_products tr td:last-of-type a')]
+            .filter(element => {
+                // login required (e.g. age verification) => javascript:void('login to read');
+                if(/(?=javascript:)(?=.*login)/i.test(element.href)) {
```
27136773: Remove unused non-capturing groups from regex `/javascript:.*login/i` should be sufficient

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+                    return auth.isLoggedIn && auth.isAdult;
+                }
+                // subscription required => javascript:void('join to read');
+                if(/(?=javascript:)(?=.*join)/i.test(element.href)) {
```
27136773: Remove unused non-capturing groups from regex `/javascript:.*join/i` should be sufficient

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+    async _getMangaVolumes(manga) {
+        let auth = await this._getUserInfo();
+        const request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request);
+
+        return [...data.querySelectorAll('#o_products tr td:last-of-type a')]
+            .filter(element => {
+                // login required (e.g. age verification) => javascript:void('login to read');
+                if(/(?=javascript:)(?=.*login)/i.test(element.href)) {
+                    return auth.isLoggedIn && auth.isAdult;
```
27136773: There is some kind of contradiction, how can there be a login link (`javascript:.*login`), when the user is already logged in (`auth.isLoggedIn === true`)? Is there a use case where this check makes sense?

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+                // subscription required => javascript:void('join to read');
+                if(/(?=javascript:)(?=.*join)/i.test(element.href)) {
+                    return auth.isMember;
+                }
+                // free
+                return true;
+            })
+            .map(chapter => {
+                return {
+                    id: chapter.href.replace("hakuneko://cache/", ""),
```
27136773: Use `this.getRootRelativeOrAbsoluteURI(chapter, this.url)` instead

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+                    title: 'Vol. ' + chapter.href.match(/-volume-([-_0-9]+)/)[1]
```
27136773: Is `volume` always lower case or should `i` modifier be added to the regex?

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+                };
+            });
+    }
+
+    async _getChapters(manga) {
+        if (manga.id.startsWith("/shonenjump/chapters")) {
+            return await this._getMangaChapters(manga);
+        } else if (manga.id.startsWith("/library")) {
```
27136773: May use `if` instead of `else if` since the method is "short-circuiting" with return.

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+            return await this._getMangaVolumes(manga);
+        }
```
27136773, line 149: Missing return (method will await forever), consider to throw if non of the conditions above is fulfilled, e.g.
```javascript
throw new Error(`Failed to get chapters/volumes for manga ${manga.id}, because the manga type is not supported!`);
```

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +127,75 @@ export default class VizShonenJump extends Connector {
...
+        let pageCount = 0;
+        let mangaID = 0;
+
+        if (chapter.id.startsWith("read/manga")) {
+            const pageCountData = await this.requestRegex(responseData, /<div class="mar-b-md"><strong>Length<\/strong> (\d+) pages<\/div>/);
+            pageCount = parseInt(pageCountData[0]);
+            const mangaIdData = await this.requestRegex(responseData, /var\s+mangaCommonId\s*=\s*(\d+)/);
+            mangaID = parseInt(mangaIdData[0]);
+        } else if (chapter.id.startsWith("/shonenjump")) {
+            const data = await this.requestRegex(responseData, /var\s+pages\s*=\s*(\d+)/);
```
27136773: Consider using `match`

27136773 (changes requested):
Approach in general seems good.
Some optional changes possible for code improvement.
Some mandatory changes since they may lead to errors.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: Yeah I think that was a bad rebase on my part, thanks!

27136773:
Is this PR still worked on or dropped?

playmer:
Sorry, I didn't drop it. Some real life stuff, as well as some additional need for me on another project has delayed me. I'll work on this tonight and see if I can get it up to snuff.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: Switching to `matchAll`

src/web/mjs/connectors/VizShonenJump.mjs
playmer: Removing method in favor of `matchAll`, match doesn't seem to return quite what I need.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: Wouldn't that kill the geolock check?

src/web/mjs/connectors/VizShonenJump.mjs
playmer: I think it's always lowercase, but no harm in adding the `i` modifier I think.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: I figured this stuff was off. I spent some time and adjusted the volume paths for login handling that's unique to how they work. Had to revert a couple minor changes with the `fetchDom` calls with queries for the checks, but I think this is generally better.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: Unfortunately not as far as I can tell, I can send an example page to you if you'd like to take a look and see if a better regex could be constructed for this. I'm not much of a wiz at these as you can tell, mostly just built off what I saw already here.

playmer:
Okay I think I got most of it, a couple comments I wasn't sure about so I've left those open. But feel free to get back to me when you can/reopen stuff or whatever. Sorry for the long delay.

src/web/mjs/connectors/VizShonenJump.mjs
27136773: Filling in whitespace may be sufficient
`/<div\s+class="mar-b-md">\s*<strong>\s*Length\s*<\/strong>\s+(\d+)\s+pages\s*<\/div>/`
https://regex101.com/r/UjHC6C/1

src/web/mjs/connectors/VizShonenJump.mjs
27136773: With the current implementation `data` is the body element right now, i think a CSS query is missing to find the chapters `this.fetchDOM(request, 'div.mangas')`

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -101,16 +123,55 @@ export default class VizShonenJump extends Connector {
             });
     }
 
+    async _getMangaVolumes(manga) {
+        const request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request);
+
+        if ( data.innerText.includes('Log in to view your library') ) {
+            alert('Lost your login cookie between fetching the manga list and now (fetching the volume list).', this.label, 'info');
```
27136773: Just throw and let HakuNeko do the error handling `throw new Error('Message whats wrong and what a user needs to do ...')`

src/web/mjs/connectors/anikiga.mjs
```diff
@@ -1,12 +0,0 @@
-import WordPressMadara from './templates/WordPressMadara.mjs';
-
-export default class Anikiga extends WordPressMadara {
```
27136773: Website is still working for me, don't remove

src/web/mjs/connectors/mundowuxia.mjs
```diff
@@ -1,12 +0,0 @@
-import WordPressMadara from './templates/WordPressMadara.mjs';
-
-export default class MundoWuxia extends WordPressMadara {
```
27136773: Consider to remove website in a separate pull request (also remove the corresponding icon)

src/web/mjs/connectors/anikiga.mjs
playmer: Oh I thought these didn't stick when I squashed. I'll remove this. 

That being said, I think these two files are a mistake? They're just dupes of two other files with slightly different casing. On a case insensitive file system like Windows you can't have both and so git obviously sees one or the other deleted.

27136773: Renaming them on your end to fit case on `master` branch may solve the conflict

playmer: I can't rename to match because both files exist in 'master'. So I'm always either removing one or the other. Like I said I can get this uncommitted. But I don't think it's intentional that these dupe files are here.

src/web/mjs/connectors/VizShonenJump.mjs
playmer: There is a query to find the chapters: https://github.com/manga-download/hakuneko/blob/d7c1f52adf137762f25a7fed41931fe90731bae3/src/web/mjs/connectors/VizShonenJump.mjs#L55-L72

If I were to implement the fetchDOM with the query, the geolock check wouldn't pass I think right? I'm assuming it'd not contain that text anymore. But I can't test it.

27136773: Yes you are right, for some reason i didn't see line 59 - 64 during my review ...

27136773 (approved):
Ready to be merged

27136773 (changes requested):
Check failed, need to fix the `lint` error

playmer:
Hey was there anything else needed for this to be merged? I think the lint got solved from that last push.

27136773:
Oh didn't got a notification mail ...

27136773 (approved):
Looks good now

PR #3493, opened 2021-05-22
[TsundokuTraduções] Add website support #2914

M4lho:
add #2914


src/web/mjs/connectors/TsundokuTraducoes.mjs
```diff
@@ -0,0 +1,14 @@
...
+
+    constructor() {
+        super();
+        super.id = 'tsundokutraducoes';
+        super.label = 'Tsundoku Traduções';
+        this.tags = [ 'webtoon', 'novel', 'portuguese', 'scanlation' ];
+        this.url = 'https://tsundokutraducoes.com.br';
+
+        this.novelContentQuery = 'div.reading-content div[class^="text-"';
+    }
```
27136773, line 12: What about [obstacles removal query](https://github.com/manga-download/hakuneko/blob/c81e82ae03f94e0cd64e760ee3ffcda3bb519cd8/src/web/mjs/connectors/WBNovel.mjs#L12)?
Are there no ads or go-to-top overlays in the novels when reading with HakuNeko that needs to be removed?

M4lho: Updated, see again pls

27136773 (approved):
Looks good

PR #3516, opened 2021-05-27
coreview fix

Ever4engel:
use a more stable mobile user agent.


27136773 (approved):
Don't see any advantage, but a more modern signature couldn't hurt

PR #3517, opened 2021-05-27
Added websites support

M4lho:
Added BL Manhwa Club;
Added Winter Scan;
Added Lemon Juice Scan;
Added Sweet Time Scan;


src/web/mjs/connectors/SweetTimeScan.mjs
```diff
@@ -0,0 +1,12 @@
...
+
+export default class SweetTimeScan extends WordPressMadaraNovel {
+
+    constructor() {
+        super();
+        super.id = 'sweettimescan';
+        super.label = 'Sweet Time Scan';
+        this.tags = [ 'manga', 'webtoon', 'portuguese' ];
+        this.url = 'https://sweetscan.net';
+    }
```
27136773, line 11: Is there no need to define `this.novelObstaclesQuery` to remove potential overlays from novels?

27136773 (approved):
Looks good

src/web/mjs/connectors/SweetTimeScan.mjs
M4lho: > Is there no need to define `this.novelObstaclesQuery` to remove potential overlays from novels?

No, dont have AD

PR #3538, opened 2021-06-01
Website Maintenance

27136773:
Update and add various websites


src/web/mjs/connectors/ComicFX.mjs
```diff
@@ -50,6 +51,21 @@ export default class ComicFX extends Connector {
...
+                }
+            } catch(error) {/**/}
+            url = url
+                // HACK: bypass 'i0.wp.com' image CDN to ensure original images are loaded directly from host
+                .replace(/\/i\d+\.wp\.com/, '')
+                // HACK: bypass 'cdn.statically.io' image CDN to ensure original images are loaded directly from host
+                .replace(/cdn\.statically\.io\/img\//, '')
+                .replace(/\/(w=\d+|h=\d+|q=\d+|f=auto)(,(w=\d+|h=\d+|q=\d+|f=auto))*\//, '/');
+            return url.includes('img.comicfx.net') ? this.createConnectorURI(url) : url;
+        }).filter(url => !url.endsWith('/4WWje.jpg'));
```
Sheepux: Add comment on this filter for future reference "removed because ad" ?

Sheepux (approved):
/LGTM

src/web/mjs/connectors/ComicFX.mjs
27136773: Thats a good and valid point that i recently overlooked too often ;)
Added constant with expressive name instead of using a magic string.

PR #3542, opened 2021-06-02
[WarQueenScan] Change domain

M4lho:
wqscan.com to wqscan.com.br


M4lho:
Opss, I accidentally committed NinjaScan again

PR #3555, opened 2021-06-08
Add more filter combinations to MangaDex connector in order to get entire list

viliml:
I removed the asc/desc combinations everywhere I could to avoid duplicates.
The old workaround fetched a total of around 40 thousand manga, now it fetches around 60 thousand. Everything is definitely complete, except the completed/safe/no demographic combination, although I doubt it has over 20 thousand manga in itself.
Some of them are still dangerously close to the 10 thousand limit and could exceed it in a few months or years, so it would be good if someone worked on a complete solution.
The best solution would be to not store the list locally at all, instead fetching it along with the search query every time, but that would require a large overhaul.
A more robust workaround would be to programmatically generate all 80 possible combinations, automatically deciding whether to request the reverse list based on if the returned count is capped at ten thousand.
Or possibly filtering by date created, restricting to a single month for example and looping over all months.

### DIFF
--- a/src/web/mjs/connectors/MangaDex.mjs
+++ b/src/web/mjs/connectors/MangaDex.mjs
@@ -57,14 +57,28 @@ export default class MangaDex extends Connector {
         let mangaList = [];
         // NOTE: Due to a limit of 10k results, it is necessary to permutate some filters in order to get as many results as possible
         const queries = [
-            { status: 'ongoing', order: 'asc' },
-            { status: 'ongoing', order: 'desc' },
-            { status: 'completed', order: 'asc' },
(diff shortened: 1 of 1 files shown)


27136773 (approved):
Probably took some time to find some good combinations, looks good
I hope the `year` filter will be fixed, so we can just go through all years (i doubt there is a year with more then 10k releases)

27136773 (changes requested):
Ups approval was to fast, please see lint errors

27136773:
Is this PR still worked on or dropped?

27136773:
Seems no further interest to fix `lint` errors => closed

PR #3566, opened 2021-06-10
[MiniTwoScan] Added Website Support

M4lho:


src/web/mjs/connectors/MiniTwoScan.mjs
```diff
@@ -0,0 +1,12 @@
+import WordPressMadara from './templates/WordPressMadara.mjs';
+
+export default class minitwoscan extends WordPressMadara {
```
27136773: Use pascal casing for class name (same as filename)

27136773:
Is the website not providing an icon?
Check if they have facebook/twitter and get icon from there.
Or use a copy of the default item (e.g. from antisensescan)

src/web/mjs/connectors/MiniTwoScan.mjs
M4lho: Okay, updated.

M4lho:
> Is the website not providing an icon?
> Check if they have facebook/twitter and get icon from there.
> Or use a copy of the default item (e.g. from antisensescan)

they have a favicon but consider it to be too small and of poor quality.
minitwo

Their Facebook page image is a generic from some manga.

M4lho:
Added ICO.

27136773 (approved):
Looks good

PR #3586, opened 2021-06-15
[Katakomik] Added Website Support

Kappamalone:


27136773:
Is the website providing an icon that can be added as well?
If not, check if they have facebook/twitter and get icon from there.
Or use a copy of the default item (e.g. from `antisensescan`)

PR #3637, opened 2021-06-28
Remove duplicate connectors Anikiga and Mundo Wuxia

playmer:
So these connectors are duplicates (even diff-wise) that have their filenames fully lowercased. Their casings were corrected in the following PR commits:
https://github.com/playmer/hakuneko/commit/87121f34549484860a254e5d2846aa8b377ac529
https://github.com/playmer/hakuneko/commit/e9fa36b0b7103b811ec436e6e35e3876a1955b37

However their original lower case files were not removed. So this PR is to remove them. They cause issues in case insensitive file systems such as Windows, as both files can't exist at the same time. Git will constantly recognize them (either these or their camel cased counterparts) as being deleted, and they cannot be restored. I got into a state where

You can see that the connectors still exist in this branch here, just in their camel cased versions:
https://github.com/playmer/hakuneko/blob/RemoteDupeConnectors/src/web/mjs/connectors/Anikiga.mjs
https://github.com/playmer/hakuneko/blob/RemoteDupeConnectors/src/web/mjs/connectors/MundoWuxia.mjs


27136773 (approved):
Looks good

Issue #3362, opened 2021-04-19
[MangaHost] Connector not working


Slowftw:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes

**Describe the bug**
Only works if I use the clipboard with the "this.url" (https://mangahostz.com/)
https://github.com/manga-download/hakuneko/blob/da8856ac6b76fd93ffc44762065c9c798fea9bd7/src/web/mjs/connectors/MangaHost.mjs#L52

**Screenshots**
![image](https://user-images.githubusercontent.com/46169656/115293191-65e46800-a12d-11eb-82bd-3cfe2c62527c.png)

**HakuNeko (please complete the following information):**
 - Version Nightly

dodieboy:
Cause by outdated domain

dodieboy:
BTW MangaHost is also affected by cloudflare too
Handled by #2933

PR #3691, opened 2021-07-11
Websites support & Fixes

### LINKED ISSUE #3704: [NameOfConnector] Connector not working
Yes

Yes

**Describe the bug**
Site changed url from .net to .com. Manga list appears, but not the chapters.

**To Reproduce**
1. Neox Scanlator
2. any manga
3. chapter no appears
4. Failed to load resource: the server responded with a status of 403 ()

**Screenshots**

 - Version [e.g. 6.17]

Slowftw:


src/web/mjs/connectors/MangaHost.mjs
```diff
@@ -49,7 +49,11 @@ export default class MangaHost extends Connector {
     }
 
     _getChapterList(manga, callback) {
-        this.fetchDOM(this.url + manga.id, 'div.chapters div.card.pop div.pop-title span')
+        this.fetchDOM(
+            // https://regex101.com/r/2gZjlK/1
+            manga.id.match(/(https?:)?(\/\/|\/\\\\)/) ?
```
27136773: Looks complicated, can `const url = new URL(manga.id, this.url).href` not be used instead?

Slowftw: ya but according to dodieboy's comment https://github.com/manga-download/hakuneko/issues/3362#issuecomment-823093119
if the domain isn't the same, the URL constructor will return an exception...
I didn't just change the domain because the domain changes constantly

The only problem is that the clipboard will only work with the domain of this.url

27136773: The URL constructor will automatically take care of it by providing a base URL
```javascript
new URL('/manga/one-piece', 'https://connector.domain').href;
// => 'https://connector.domain/manga/one-piece'
new URL('https://foreign.domain/manga/one-piece', 'https://connector.domain').href;
// => 'https://foreign.domain/manga/one-piece'
```

27136773: For regularly changing domains, simply apply the new domain during initialization, e.g. as in [HeavenManga](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/HeavenManga.mjs#L20-L21)

27136773: To support multi-domain copy & paste, overwrite the [canHandleURI()](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/HeavenManga.mjs#L13-L15) method with a corresponding pattern matching

Slowftw: o✌️ 
I didn't know about multi-domain 👀

27136773 (approved):
Looks good

PR #3736, opened 2021-07-18
Update NeoxScan.mjs

M4lho:
.com to .net


src/web/mjs/connectors/NeoxScan.mjs
```diff
@@ -7,6 +7,6 @@ export default class NeoxScan extends WordPressMadara {
         super.id = 'neoxscan';
         super.label = 'Neox Scanlator';
         this.tags = [ 'manga', 'webtoon', 'portuguese', 'scanlation' ];
-        this.url = 'https://neoxscans.com';
+        this.url = 'https://neoxscans.net';
```
27136773, line 10: The `.net` is redirect to the `.com` domain for me.
Could it be possible they are swapping domains in regular intervals?

PR #3770, opened 2021-08-01
New Connector for "onepiece-tube.com"

ProfEibe:
Simple connector for the german manga-site onepiece-tube


ProfEibe:
I haven't changed anything on this part of the code, so I don't get it, why it fails...

src/web/mjs/connectors/OnePieceTube.mjs
```diff
@@ -0,0 +1,48 @@
+import Connector from "../engine/Connector.mjs";
+
+export default class OnePieceTube extends Connector {
+    constructor() {
+        super();
+        super.id = 'onepiecetube';
+        super.label = 'One Piece Tube';
+        this.tags = ['manga', 'german'];
+        this.url = 'http://onepiece-tube.com/';
```
27136773, line 9: May remove trailing `/` slash

src/web/mjs/connectors/OnePieceTube.mjs
```diff
@@ -0,0 +1,48 @@
...
+                title: `${number} - ${title}`
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(`${this.url}${chapter.id}/1`, this.requestOptions);
+        let data = await this.fetchDOM(request, 'td#tablecontrols a');
+
+        data = await Promise.all(data.map(d => {
```
27136773, line 40: Do not start scraping all pages, this will slow down HakuNeko. Use the protocol mechanism(`createConnectorURI`) by returning the HTML links and extract the images in `_handleConnectorURI` like done in e.g. [TAAD](https://github.com/manga-download/hakuneko/blob/c81e82ae03f94e0cd64e760ee3ffcda3bb519cd8/src/web/mjs/connectors/TAADD.mjs#L75-L85)

27136773:
No further response within 2 month => assumed to be abandoned
Feel free to re-open the PR in case the comments are going to be addressed.

src/web/mjs/connectors/OnePieceTube.mjs
ProfEibe: think i fixed it

PR #3784, opened 2021-08-07
Add generation of filter combinations for queries to Mangadex connector in order to get entire list of mangas

Druzai:
I tried to work on a complete solution mentioned in #3555 that generates array of queries with necessary combinations of filters and doesn't download duplicates when there are 2 queries which only differ by order filter.

This code finds and downloads entire list of mangas. I compared sum of all mangas fetched via mangadex api with number of mangas from hakuneko after updating local list, and they were equal.

Generating all queries (9 at this moment) takes about 2 seconds on my machine.
```
0: {publicationDemographic: "none", order: "asc", rating: "safe"}
1: {publicationDemographic: "none", order: "desc", rating: "safe", mangasLimit: 6266}
2: {publicationDemographic: "none", order: "asc", rating: "suggestive"}
3: {publicationDemographic: "none", order: "asc", rating: "erotica"}
4: {publicationDemographic: "none", order: "asc", rating: "pornographic"}
5: {publicationDemographic: "josei", order: "asc"}
6: {publicationDemographic: "seinen", order: "asc"}
7: {publicationDemographic: "shoujo", order: "asc"}
8: {publicationDemographic: "shounen", order: "asc"}
```
Currently search runs with these filters: `publicationDemographic[]`, `order[createdAt]`, `contentRating[]`, `status[]`.
From the [search](https://api.mangadex.org/manga?limit=0&offset=0&status[]=completed&publicationDemographic[]=none&order[createdAt]=asc&contentRating[]=safe) with all 4 parameters that gives the largest number of "total" parameter I get 11181 items, so I think it will be fine for several months or maybe years. :)

I'm not sure about naming of some variables, so if something wrong I'm open to suggestions.

### DIFF
--- a/src/web/mjs/connectors/MangaDex.mjs
+++ b/src/web/mjs/connectors/MangaDex.mjs
@@ -12,6 +12,7 @@ export default class MangaDex extends Connector {
         this.api = 'https://api.mangadex.org';
         this.requestOptions.headers.set('x-referer', this.url);
         this.throttleGlobal = 100;
+        this.resultLimit = 10000;
         this.licensedChapterGroups = [
             '4f1de6a2-f0c5-4ac5-bce5-02c7dbb67deb' // MangaPlus
         ];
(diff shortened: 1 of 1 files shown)


27136773:
To me the solution looks quite smart, but the code is complicated and has a high complexity (3rd level nested `for` and `if` blocks).
Any opinions if it is worth to save a few dozen requests at the cost of code readability?

Druzai:
Okay, I will try to simplify my code by splitting it into 3-4 functions.

Druzai:
Tried to remove 3rd level nested `for` (max nested level is 2nd now), but I don't think generating queries can be simplified more than this because there are more than 1 filter.
If you think that this code still not readable, then maybe it is better to use static queries and just extend them with including `rating` or `status` parameter when `publicationDemographic` is none.

Druzai:
@ronny1982 I answered you but without mention in last comment. Sorry if you didn't saw it.

> If you think that this code still not readable, then maybe it is better to use static queries and just extend them with including `rating` or `status` parameter when `publicationDemographic` is none.

I need your opinion because I accidentally deleted my fork of this repository. (stupid, I know)
If you do not reject these changes, I will create a new PR with these commits (I have them locally). Otherwise, it's better to close this PR.

27136773:
I saw it, but I'm still not sure if the overhead and complexity is worth it, so i left it open for later to decide or to wait for comments from other contributors/users ...

Druzai:
Okay, got it

27136773:
Closed due to #4167

PR #3798, opened 2021-08-08
Update ManhwaTaro.mjs

dengole899:
Update site Please


Sheepux:
We should just delete the connector if it's now Manhwaland (a connector that already exists)

Relates #3795

PR #3807, opened 2021-08-11
Fix/kakaopage

Pierre-Monier:
I haven't found a way to bypass the auth cookie :/, it's way harder than I thought. So for now, the kakaopage connector can only download "free" content which doesn't require authentication (a very small amount of content)

### DIFF
--- a/src/web/mjs/connectors/kakaopage.mjs
+++ b/src/web/mjs/connectors/kakaopage.mjs
@@ -74,6 +74,6 @@ export default class kakaopage extends Connector {
             return data.downloadData.members.files.map(element => data.downloadData.members.sAtsServerUrl + element.secureUrl);
         }
 
-        throw new Error('Logged in / Purchased ...');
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/kakaopage.mjs
```diff
@@ -70,6 +70,10 @@ export default class kakaopage extends Connector {
         this.requestOptions.method = 'GET';
         delete this.requestOptions.body;
         let data = await this.fetchJSON(request);
-        return data.downloadData.members.files.map(element => data.downloadData.members.sAtsServerUrl + element.secureUrl);
+        if (data.downloadData && data.downloadData.members) {
+            return data.downloadData.members.files.map(element => data.downloadData.members.sAtsServerUrl + element.secureUrl);
+        }
+
+        throw new Error('Can\'t fetch this ressource because it\'s protected');
```
27136773: May use backticks to avoid escaping `` `Can't` ``

27136773 (approved):
Looks good

PR #3808, opened 2021-08-12
add support for 520漫画网 (kokomh).

adelie:
**tl;dr.** new connector for [520漫画网](https://www.kokomh.com/).

this connector is heavily based on the connector for `bx117`, which shares the same pages query script. i've just changed it around to account for the different DOM elements and converted it to using the `SinMH` format.


src/web/mjs/connectors/KokoMH.mjs
```diff
@@ -0,0 +1,34 @@
+import SinMH from './templates/SinMH.mjs';
+
+export default class KokoMH extends SinMH {
+    
+    constructor() {
+        super();
+        super.id = 'kokomh';
+        super.label = '520漫画网 (KokoMH)';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.kokomh.com/';
```
27136773: May remove trailing`/` slash

src/web/mjs/connectors/KokoMH.mjs
```diff
@@ -0,0 +1,34 @@
...
+        this.path = '/all/%PAGE%.html'
+        this.queryManga = '.mh-date-info-name h4 a';
+        this.queryMangasPageCount = 'div.NewPages ul li:last-child a';
+        this.queryMangasPageCountMatch = /(\d+)(\/|\.html)?$/;
+        this.queryMangas = 'div.mh-works-title h4 a';
+        this.queryChaptersScript = `
+            new Promise(resolve => {
+                let chapters = [...document.querySelectorAll('ul#mh-chapter-list-ol-0 li a')].map(element => {
+                    return {
+                        id: new URL(element.href, window.location).pathname,
```
27136773: I think could be simplified to `id: element.pathname,`

27136773:
See `lint` errors marked by Continuous Integration run, otherwise looks good

27136773 (approved):
Looks good

PR #3900, opened 2021-08-28
GekkouScans - CMS to Madara

M4lho:


27136773:
Does the icon also need to be updated?

PR #3928, opened 2021-09-04
[SoftKomik] updates API (#3879)

### LINKED ISSUE #3879: [SoftKomik] Connector not working
Yes

Yes

**Describe the bug**
cannot load chapter list

**To Reproduce**
1. Name of connector

**Screenshots**

 - Version [e.g. 6.17]

dodieboy:
fix #3879

### DIFF
--- a/src/web/mjs/connectors/SoftKomik.mjs
+++ b/src/web/mjs/connectors/SoftKomik.mjs
@@ -27,13 +27,11 @@ export default class SoftKomik extends Connector {
         let mangaList = [];
         for (let page = 1, run = true; run; page++) {
             const mangas = await this._getMangasFromPage(page);
-            if(mangas.length == 0){
+            if(mangas.length == 0) {
                 run = false;
-            }
-            else if (mangaList.length == 0 || mangas[mangas.length - 1].id !== mangaList[mangaList.length - 1].id){
+            } else if (mangaList.length == 0 || mangas[mangas.length - 1].id !== mangaList[mangaList.length - 1].id) {
                 mangaList.push(...mangas);
-            }
-            else{
+            } else {
                 run = false;
             }
         }
--- a/src/web/mjs/connectors/SoftKomik.mjs
+++ b/src/web/mjs/connectors/SoftKomik.mjs
@@ -9,7 +9,7 @@ export default class SoftKomik extends Connector {
         super.label = 'Softkomik';
         this.tags = [ 'manga', 'indonesian' ];
         this.url = 'https://softkomik.site';
-        this.api = 'https://img.softkomik.online/api/';
+        this.api = 'https://mongo.animeyusha.com/';
         this.cdn = 'https://cdn.statically.io/img/softkomik.online/';
         this.links = {
             login: this.url + '/akun'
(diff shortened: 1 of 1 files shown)


black0888 (approved):

src/web/mjs/connectors/SoftKomik.mjs
```diff
@@ -27,16 +27,16 @@ export default class SoftKomik extends Connector {
         let mangaList = [];
         for (let page = 1, run = true; run; page++) {
             const mangas = await this._getMangasFromPage(page);
-            mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+            mangas.length > 27 ? mangaList.push(...mangas) : run = false;
```
Sheepux: That means that if there are 26 titles on the page they won't get retrieved. I think there's an issue here :)

dodieboy: My bad, fix the issue alr

dodieboy: My bad, fix the issue alr

Sheepux: why did you remove the ternary operator (if/else) ?
If the list length is 0 (when reaching the last page and the preview page was full with 26 items) it would push the undefined result into the array.
+
lint error ;)

dodieboy: The api will keep giving the same result when it reach the last page. For example the last page have 26 data, last page +1 will also have 26 and not 0

dodieboy: Will this be a better way to write the code? Not very good in ternary operator, that why i change it to if else operator

```
            mangas.length > 27 ? mangaList.push(...mangas) : run = false;
}
mangaList.push(..mangas)
```

Sheepux: I think we need to compare the last results to find if they're equals and then get out of the loop.

Sheepux: I think we can leave it like that (your last code in comments). If someone requests a few missing titles we'll take a look for an alternative as we're not able to unit test  of what happens when the last page has 28 titles.
Edit: ah, it could loop infinitely in such case :\ So comparing the last results would be the actual solution i guess

dodieboy: I feel that this code will be better, as the previous code in the comment will cause a loop if the last page have 27 data.
```    async _getMangas() {
        let mangaList = [];
            if(mangas.length == 0){
                run = false;
            }
            else if (mangaList.length == 0 || mangas[mangas.length - 1].id !== mangaList[mangaList.length - 1].id){
                mangaList.push(...mangas);
            }
            else{
                run = false;
            }
        }
        return mangaList;
    }
```

27136773:
There are still some `lint` errors that needs to be resolved

dodieboy:
> There are still some `lint` errors that needs to be resolved

image
Hey, will like to check why im have this lint error 
Closing curly brace does not appear on the same line as the subsequent block  brace-style

27136773:
`} else {`

src/web/mjs/connectors/SoftKomik.mjs
```diff
@@ -27,16 +27,22 @@ export default class SoftKomik extends Connector {
         let mangaList = [];
         for (let page = 1, run = true; run; page++) {
             const mangas = await this._getMangasFromPage(page);
-            mangas.length > 0 ? mangaList.push(...mangas) : run = false;
+            if(mangas.length == 0) {
```
27136773, line 30: Is it possible to include this check into `_getMangasFromPage` and return an empty manga list instead a duplicate of the last page list (e.g. by checking the page exceeds the max. number of pages or something)

dodieboy: Dont think so, when the api reach the last page, it will repeat the last page. Even the website itself do not have a max page number and will repeat the last managa list as seen in the img below. The only way is to check for duplicate.
image

27136773 (approved):
Looks good now

PR #3935, opened 2021-09-05
Fixed Webnovel comics - API Update

amit34521:
WebNovel Comics API Updated. Added Fix for that


Sheepux:
I don't see any difference, did it fix anything ? The current connector is working fine here.

amit34521:
Its not working. the API path has changed hence i updated the same in connector.

image

Sheepux (approved):
My Bad, it was indeed not working on my side !
/LGTM

PR #3942, opened 2021-09-08
fix(connector/9anime): change the _getPages script to avoid clicking

### LINKED ISSUE #3960: [9Anime] Connector not working
Yes

Yes

**Describe the bug**

**To Reproduce**
1. Name of connector (9Anime)
2. Name of manga (if connector update is working) Naruto

**Screenshots**

 - Version [e.g. 6.17]

Pierre-Monier:
I spend a lot of time figuring out why the `button.click()` doesn't do anything. I'm still not sure why, but I change the script and it's working again.

Maybe we can decrease the `setInterval` delay to go faster :thinking: ?

### DIFF
--- a/src/web/mjs/connectors/NineAnime.mjs
+++ b/src/web/mjs/connectors/NineAnime.mjs
@@ -139,23 +139,15 @@ export default class NineAnime extends Connector {
             new Promise((resolve, reject) => {
                 const timer = setInterval(() => {
                     try {
-                        if(button = document.querySelector('ul.episodes li a.active')) {
+                        const iframe = document.querySelector('div#player iframe')
+
+                        if (iframe) {
                             clearInterval(timer);
                             localStorage.setItem('player_autoplay', 0);
                             localStorage.setItem('_lastServerId', ${chapterID.server});
-                            new MutationObserver(mutations => {
-                                for(const mutation of mutations) {
-                                    for(const node of mutation.addedNodes) {
-                                        if(/iframe/i.test(node.tagName)) {
-                                            node.parentNode.removeChild(node);
-                                            resolve(node.src);
-                                        }
-                                    }
-                                }
-                            }).observe(document.querySelector('div#player'), {
-                                childList: true
-                            });
-                            button.click();
+                            resolve(iframe.src)
+                        } else {
+                            window.location.reload()
                         }
                     } catch(error) {
                         reject(error);


src/web/mjs/connectors/NineAnime.mjs
```diff
@@ -139,23 +139,15 @@ export default class NineAnime extends Connector {
...
-                                        }
-                                    }
-                                }
-                            }).observe(document.querySelector('div#player'), {
-                                childList: true
-                            });
-                            button.click();
+                            resolve(iframe.src)
+                        } else {
+                            window.location.reload()
```
Sheepux, line 150: I think that with the browser reload proposal  you're relying on the browser cache "fast load" but it would behave differently on potato computers and it would be a race condition.
The mutation observer needs to be fixed or altered.

Pierre-Monier: I didn't really have potato computers in mind when developing that. 

The problem didn't come from the MutationObserver. It's working fine, the issue was the `button.click()` which doesn't trigger anything for a reason I don't understand. If you click on the button manually, the MutationObserver do his job.

27136773: 9anime may have overwritten the original click event, or are checking if the event arguments are from a valid mouse click...
There are other ways to raise the event programatically with corresponding mouse arguments as well

Pierre-Monier: We can make the click works with something like that : 

```javascript
  const clickEvent = new MouseEvent("click", {
      pointerX: getOffset(button).left,
      pointerY: getOffset(button).top,
      button: 0,
      ctrlKey: false,
      altKey: false,
      shiftKey: false,
      metaKey: false,
      bubbles: true,
      cancelable: false,
      relatedTarget: button
  });
  button.dispatchEvent(clickEvent)
```

The issue is that a "real" click return an object with a key `isTrusted` which got the value `true`. I still can't figure out how to mock this, I tried many things. When the click event return `{isTrusted: false, ...}` it reloads the page for some reason, so the `MutationObserver` is useless.

27136773: In that case we could try to go full out and attach the chrome debugger to fire the mouse click (like done in JapScan)
https://stackoverflow.com/questions/34853588/how-to-trigger-an-istrusted-true-click-event-using-javascript

27136773:
Superseded by #3980, moved to encrypted modules to make it harder for 9anime counter-measures ;)
=> closed

PR #3958, opened 2021-09-12
Fix Mangadex connector: change key from 'results' to 'data'

### LINKED ISSUE #3959: [Mangadex] manga can't be fetched from connector
Yes

Yes

**Describe the bug**
Cannot fetch any manga from mangadex. I can see the request in the developper tools but nothings is adding up in my list. I tried with kissmanga, and it works.

**To Reproduce**
1. MangaDex
4. There is no error in the console

**Screenshots**

 - Version [e.g. 6.17]

Druzai:
It seems that in provided json when requesting mangas, chapters and scanlation groups they changed key 'results' to 'data'.

### DIFF
--- a/src/web/mjs/connectors/MangaDex.mjs
+++ b/src/web/mjs/connectors/MangaDex.mjs
@@ -94,11 +94,11 @@ export default class MangaDex extends Connector {
         uri.searchParams.append('contentRating[]', 'pornographic');
         const request = new Request(uri, this.requestOptions);
         const data = await this.fetchJSON(request, 3);
-        return !data.results ? [] : data.results.map(result => {
+        return !data.data ? [] : data.data.map(result => {
             const title = document.createElement('div');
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -94,11 +94,11 @@ export default class MangaDex extends Connector {
         uri.searchParams.append('contentRating[]', 'pornographic');
         const request = new Request(uri, this.requestOptions);
         const data = await this.fetchJSON(request, 3);
```
Pierre-Monier: You can deconstruct the result as follows `const {data} = await this.fetchJSON(request, 3);`, it's better, but it looks like it's working :)

Druzai: Thanks for the info! I didn't know about this. Then I'll rewrite as you suggested.

27136773 (approved):
Haven't tested it, but technically looks good

PR #3973, opened 2021-09-17
[MangaHost] add throttling

TrickOC:
Adding throttling to get manga from source, site fail when too many requests at the same time during manga list update.


27136773:
There is a `lint` error that needs to be fixed

27136773:
Right now throttling will only be applied when downloading images.
If it should be used when updating the manga list, its value needs to be processed/awaited in `_getMangaListFromPages`

PR #4009, opened 2021-09-30
merge from master

27136773:


27136773:
> Some commits has already been commited to 6.1.7 ...
> Self approved

Cherry picking was done through `GitHub Desktop`, don't know why the already integrated commits from October 28th were added again (despite not being selected for cherry-picking...)
However, there is no harm in picking the same commits twice as they cause no changes or conflicts anyway

PR #4038, opened 2021-10-09
[ganma] manga and chapter list improvments

### LINKED ISSUE #4033: [Ganma] Connector not working
Yes

Yes

**Describe the bug**
Some mangas i need to download (双無しの人形 and others) are not listed in the ganma connector, and copying the link (https://ganma.jp/theoriginaldoll) also doesn't work.
Other problems: the last chapters that you can read ahead of time with the subscription aren't listed either, and when i download a chapter it doesn't download the afterword image, which thends to be a part of the chapter

**To Reproduce**
1. Ganma
2. Open page and log in
3. 双無しの人形

 - Version 6.1.7

27136773:


Sheepux (approved):

src/web/mjs/connectors/Ganma.mjs
```diff
@@ -9,44 +9,56 @@ export default class Ganma extends Connector {
...
-                callback( null, mangaList );
-            } )
-            .catch( error => {
-                console.error( error, this );
-                callback( error, undefined );
-            } );
+    async _getMangas() {
+        return [
+            ...await this._getMangasTop(),
+            ...await this._getMangasCompleted()
```
Sheepux, line 26: could use `await Promise.all` to get it faster ( would charge the website and create issues if throttling)
Edit: hmmm, or maybe it does trigger both at the same time ... good question ...

27136773: Both requests are performed sequentially.
I leave it as is due to better code readability. Performance shouldn't be a problem, it is still considered very fast, even with the list being fetched in maybe four seconds instead of two.

PR #4063, opened 2021-10-14
get links fixed

Lyem:


src/web/mjs/connectors/Leitor.mjs
```diff
@@ -96,6 +95,10 @@ export default class Leitor extends Connector {
         const request = new Request(uri, this.requestOptions);
         request.headers.set('X-Requested-With', 'XMLHttpRequest');
         const data = await this.fetchJSON(request);
-        return data.images;
```
27136773, line 99: May use `return data.images.map(image => image.legacy);` instead of `for` loop

PR #4088, opened 2021-10-26
Viz redundant 18+ check

amit34521:
Download works without login so there's no need to check that.


src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -88,9 +88,9 @@ export default class VizShonenJump extends Connector {
         return [...data.querySelectorAll('div > a.o_chapter-container[data-target-url], tr.o_chapter td.ch-num-list-spacing a.o_chapter-container[data-target-url]')]
             .filter(element => {
                 // login required (e.g. age verification) => javascript:void('login to read');
-                if(/javascript:.*login/i.test(element.href)) {
+                /*if(/javascript:.*login/i.test(element.href)) {
```
27136773: Remove commented code

27136773 (approved):
looks good

Issue #4090, opened 2021-10-27
[Pied Piper] Website Change


ghost:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes

**Describe the bug**
Chapter list doesn't show because of website change.
New site address is https://piedpiperfansub.com/
The old was https://piedpiperfb.com/

**HakuNeko (please complete the following information):**
 - Version 6.17@a8e80e

PR #4094, opened 2021-10-28
Update PiedPiper.mjs

abhakta-47:
#4090 bug ::
Chapter list doesn't show because of website change.
New site address is https://piedpiperfansub.com/
The old was https://piedpiperfb.com/

### DIFF
--- a/src/web/mjs/connectors/PiedPiper.mjs
+++ b/src/web/mjs/connectors/PiedPiper.mjs
@@ -4,9 +4,9 @@ export default class PiedPiper extends WordPressMadara {
 
     constructor() {
         super();
-        super.id = 'piedpiperfb';
+        super.id = 'piedpiperfansub';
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/PiedPiper.mjs
```diff
@@ -4,9 +4,9 @@ export default class PiedPiper extends WordPressMadara {
 
     constructor() {
         super();
-        super.id = 'piedpiperfb';
+        super.id = 'piedpiperfansub';
```
27136773: `ID` should not be changed, bookmarks of the users will become invalid and need to be re-bookmarked.
Also if this is really necessary, the icon must be renamed to match the new `ID` as well.

abhakta-47: OK, I would leave the Id unchanged then

PR #4101, opened 2021-10-29
Fix volume library path, as well as manga with subseries

playmer:
The path to a users library is slightly different now, the old path still works for fetching volume based manga on the account, but it cannot be used to get the volumes  within a given manga. This has been addressed.

As well, there are some manga which contain subseries. These manga would otherwise give titles that are simply the `Vol. ${Number}`, despite there being possibly 2 or even far more duplicates of the same volume number given how many sub series they might have. So we try to detect this situation by checking all the volumes start with the same name before the Volume number, and if they don't we assume this manga has subseries and so give each volume the full name to differentiate.

### DIFF
--- a/src/web/mjs/connectors/VizShonenJump.mjs
+++ b/src/web/mjs/connectors/VizShonenJump.mjs
@@ -127,13 +127,33 @@ export default class VizShonenJump extends Connector {
             throw new Error('Lost your login cookie between fetching the manga list and now (fetching the volume list), please log back in.', this.label, 'info');
         }
 
-        return [...data.querySelectorAll('#o_products tr td:last-of-type a')]
+        let volumeNames = [...data.querySelectorAll('#o_products tr td:nth-child(3)')]
+            .map(volume => {
+                return volume.innerText;
+            });
+
+        let firstVolumeName = volumeNames[0].substring(0, volumeNames[0].indexOf("Vol."));
+        let allNamesMatch = volumeNames.every(volume => {
+            return volume.startsWith(firstVolumeName);
+        });
+
+        let volumes = [...data.querySelectorAll('#o_products tr td:last-of-type a')]
             .map(volume => {
                 return {
                     id: this.getRootRelativeOrAbsoluteLink(volume, this.url).substring(1),
                     title: 'Vol. ' + volume.href.match(/-volume-([-_0-9]+)/i)[1]
                 };
             });
+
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -35,7 +35,7 @@ export default class VizShonenJump extends Connector {
     }
 
     async _getMangasAvailibleByVolumes() {
-        const request = new Request(new URL('/library', this.url), this.requestOptions);
+        const request = new Request(new URL('/account/library', this.url), this.requestOptions);
```
27136773, line 38: Is the query `table.purchase-table a` to extract volumes still working?
Couldn't find it in the page source code...

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -127,20 +127,40 @@ export default class VizShonenJump extends Connector {
             throw new Error('Lost your login cookie between fetching the manga list and now (fetching the volume list), please log back in.', this.label, 'info');
         }
 
-        return [...data.querySelectorAll('#o_products tr td:last-of-type a')]
+        let volumeNames = [...data.querySelectorAll('#o_products tr td:nth-child(3)')]
+            .map(volume => {
+                return volume.innerText;
+            });
+
+        let firstVolumeName = volumeNames[0].substring(0, volumeNames[0].indexOf("Vol."));
```
27136773, line 135: May use `.split('Vol.').shift();`

27136773 (approved):
Looks good

Some optional suggestions...

It seems volume titles and volume links are handled separately and then zipped together.
Maybe it would be easier to map the full table row and then extract/combine the title and link within a single `map`?

```javascript
return [...data.querySelectorAll('#o_products tr')].map(element => {
    const title = element.querySelector('td:nth-child(3)').innerText.trim();
    const link = element.querySelector('td:last-of-type a');
    // some conditional composition of link / title in case of sub-series ...
    return {
        id: this.getRootRelativeOrAbsoluteLink(link, this.url),
        title: 'Vol. ' + link.href.match(/-volume-([-_0-9]+)/i)[1]
    };
});
```

playmer:
Sure, I'll take a pass on these in the next day or so, been busy since I submitted. Just need some time at my PC after work. Thanks for the review!

27136773:
I don't see a blocker, so i'll merge the PR.
You may provide additional improvements in a separate PR.

PR #4142, opened 2021-11-11
[comikey] fixed image extraction

### LINKED ISSUE #4136: [Comikey] Connector not working
Yes

Yes

**Describe the bug**
Downloading any free chapter doesnt work. Keeps throwing error.

**Screenshots**
image

 - Version [e.g. 6.17]

27136773:
Review:
- manga-download/hakuneko-plugins@41cebed
- manga-download/hakuneko-plugins@a40ca95
- manga-download/hakuneko-plugins@6c5aa32


amit34521:
The highest quality returned happens to be webp. however, can you make it to consider the alternate array only when the default readingorder.href is missing as in the image below?

image

27136773:
> The highest quality returned happens to be webp. however, can you make it to consider the alternate array only when the default readingorder.href is missing as in the image below?
> 
> image

What manga/chapter is this?
From my findings the default image is always part of the alternate links as well, in which case only the sorting needs to be changed (consider `.jpg` < `.webp` besides the highest resolution)

amit34521:
I checked both "battle in 5 seconds" ch 80 & 81 and "happy harem making" ch 60 manga. In both of them, I noticed that the highest jpeg was not the part of the alternate array. 
The image is from happy harem making ch 61.

Looking at it, they are providing the jpeg in the root and rest options in the alternate array.

Sheepux (approved):
/LGTM

A quite fun way to handle the case. (such "wars" are actually the fun ones)

PR #4167, opened 2021-11-19
mangadex improve getManga

### LINKED ISSUE #4161: [MangaDex] Connector not working
Yes

Yes

**Describe the bug**
When pressing the refresh button for website MangaDex, I get the error "Failed to upload manga list for MangaDex. Failed to receive content from ...://api.mangadex.org/manga?limit=100&offset=2900&publicationDemographic%5B%5D=... (Status: 404) - 

**To Reproduce**
1. Name of connector - MangaDex
2. Name of manga (if connector update is working) Update not working

 - Version 6.1.7@9436d8

Robonau:
this is just #3784 but better

this should be as good as it gets
all manga
minimum requests
(totalManga + floor(totalManga/10000))/100 requests

it would be nice to get rid of the floor(totalManga/10000) (this is currently 5) but dealing with them is easier than possibly missing one by changing the times

it works now fixed the captcha issue (and 503)
resolves #4161

### DIFF
--- a/src/web/mjs/connectors/MangaDex.mjs
+++ b/src/web/mjs/connectors/MangaDex.mjs
@@ -11,7 +11,7 @@ export default class MangaDex extends Connector {
         this.url = 'https://mangadex.org';
         this.api = 'https://api.mangadex.org';
         this.requestOptions.headers.set('x-referer', this.url);
-        this.throttleGlobal = 100;
+        this.throttleGlobal = 200;
         this.licensedChapterGroups = [
             '4f1de6a2-f0c5-4ac5-bce5-02c7dbb67deb', // MangaPlus
             '8d8ecf83-8d42-4f8c-add8-60963f9f28d9' // Comikey
@@ -227,7 +227,7 @@ export default class MangaDex extends Connector {
         retries = retries || 0;
         let response = await fetch( request );
         if( (response.status >= 500 || /captcha-v3$/.test(response.url)) && retries > 0 ) {
-            await this.wait( 500 );
+            await this.wait( 5000 );
             return await this.fetchJSON( request, retries - 1 );
         }
         if( response.status === 200 ) {
--- a/src/web/mjs/connectors/MangaDex.mjs
+++ b/src/web/mjs/connectors/MangaDex.mjs
@@ -56,7 +56,8 @@ export default class MangaDex extends Connector {
         // NOTE: The MangaDex website is still down, but there are some provisional frontends which can be used for search, copy & paste
         const regexGUID = /[a-fA-F0-9]{8}-([a-fA-F0-9]{4}-){3}[a-fA-F0-9]{12}/;
         const id = (uri.pathname.match(regexGUID) || uri.hash.match(regexGUID))[0].toLowerCase();
-        const {data} = await this.fetchJSON(new URL('/manga/' + id, this.api));
+        const request = new Request(new URL('/manga/' + id, this.api), this.requestOptions);
+        const {data} = await this.fetchJSON(request);
         return new Manga(this, id, data.attributes.title.en || Object.values(data.attributes.title).shift());
     }
 
@@ -97,7 +98,8 @@ export default class MangaDex extends Connector {
             uri.searchParams.append('contentRating[]', 'erotica');
             uri.searchParams.append('contentRating[]', 'pornographic');
             if (nextAt) uri.searchParams.set('createdAtSince', nextAt);
-            data100 = await this.fetchJSON(uri, 3);
+            const request = new Request(uri, this.requestOptions);
+            data100 = await this.fetchJSON(request, 3);
             await this.wait(this.config.throttle.value);
             tmp = [...tmp, ...data100.data];
         }
@@ -127,8 +129,9 @@ export default class MangaDex extends Connector {
         uri.searchParams.append('contentRating[]', 'erotica');
         uri.searchParams.append('contentRating[]', 'pornographic');
         uri.searchParams.set('manga', manga.id);
-        const {data} = await this.fetchJSON(uri, 3);
-        const groupMap = await this._getScanlationGroups(data);
+        const request = new Request(uri, this.requestOptions);
+        const {data} = await this.fetchJSON(request, 3);
+         const groupMap = await this._getScanlationGroups(data);
         return !data ? [] : data.map(result => {
             let title = '';
             if(result.attributes.volume) {
@@ -162,7 +165,8 @@ export default class MangaDex extends Connector {
 
     async _getPages(chapter) {
         const uri = new URL('/chapter/' + chapter.id, this.api);
-        const {data} = await this.fetchJSON(uri, 3);
+        const request = new Request(uri, this.requestOptions);
+        const {data} = await this.fetchJSON(request, 3);
         const server = await this._getServerNode(chapter);
         return data.attributes.data.map(file => this.createConnectorURI({
             networkNode: server, // e.g. 'https://foo.bar.mangadex.network:44300/token/data/'
@@ -202,16 +206,18 @@ export default class MangaDex extends Connector {
         if (groupIDs.length > 0) {
             await this.wait(this.config.throttle.value);
             const uri = new URL('/group', this.api);
-            uri.search = new URLSearchParams([ ['limit', 100 ], ...groupIDs.map(id => [ 'ids[]', id ]) ]).toString();
-            const {data} = await this.fetchJSON(uri, 3);
-            data.forEach(result => groupList[result.id] = result.attributes.name || 'unknown');
+            uri.search = new URLSearchParams([ [ 'limit', 100 ], ...groupIDs.map(id => [ 'ids[]', id ]) ]).toString();
+            const request = new Request(uri, this.requestOptions);
+            const {data} = await this.fetchJSON(request, 3);
+             data.forEach(result => groupList[result.id] = result.attributes.name || 'unknown');
         }
         return groupList;
     }
 
     async _getServerNode(chapter) {
         const uri = new URL('/at-home/server/' + chapter.id, this.api);
-        const data = await this.fetchJSON(uri, 3);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchJSON(request, 3);
         return data.baseUrl + '/data/';
     }
 
--- a/src/web/mjs/connectors/MangaDex.mjs
+++ b/src/web/mjs/connectors/MangaDex.mjs
@@ -52,26 +52,26 @@ export default class MangaDex extends Connector {
     }
 
     async _getMangas() {
-        let mangaList = []
-        let data1 = await this._getMangasFromPages(0, 99)
-        mangaList = [...mangaList, ...data1.data]
-        nextAt = data1.nextAt
+        let mangaList = [];
+        let data1 = await this._getMangasFromPages(0, 99);
+        mangaList = [...mangaList, ...data1.data];
+        let nextAt = data1.nextAt;
 
         for (let i = 1; i <= data1.total / 10000; i += 1) {
-            let data2 = await this._getMangasFromPages(1, 1, nextAt)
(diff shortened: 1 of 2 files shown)


src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -47,66 +47,59 @@ export default class MangaDex extends Connector {
...
-        await this.wait(this.throttleGlobal);
-        const uri = new URL('/manga', this.api);
-        uri.searchParams.set('limit', 100);
-        uri.searchParams.set('offset', 100 * page);
-        uri.searchParams.set('publicationDemographic[]', query.publicationDemographic);
-        uri.searchParams.set('order[createdAt]', query.order);
-        uri.searchParams.append('contentRating[]', 'safe');
-        uri.searchParams.append('contentRating[]', 'suggestive');
-        uri.searchParams.append('contentRating[]', 'erotica');
-        uri.searchParams.append('contentRating[]', 'pornographic');
```
27136773, line 94: Removing this content rating will lead to missing more then 9000 mangas in the list
https://api.mangadex.org/manga?contentRating%5B%5D=safe&contentRating%5B%5D=suggestive&contentRating%5B%5D=erotica&contentRating%5B%5D=pornographic

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -47,66 +47,59 @@ export default class MangaDex extends Connector {
         const regexGUID = /[a-fA-F0-9]{8}-([a-fA-F0-9]{4}-){3}[a-fA-F0-9]{12}/;
         const id = (uri.pathname.match(regexGUID) || uri.hash.match(regexGUID))[0].toLowerCase();
         const request = new Request(new URL('/manga/' + id, this.api), this.requestOptions);
-        const {data} = await this.fetchJSON(request);
+        const { data } = await this.fetchJSON(request);
```
27136773: Use more expressive variable names, or re-use (re-assign) the same variable to prevent confusion within the same scope.
Names such as `data`, `data1`, `data2`, `data3` may be confusing for other contributors.

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -47,66 +47,59 @@ export default class MangaDex extends Connector {
...
+            const uri = new URL('/manga', this.api);
+            uri.searchParams.set('limit', 100);
+            uri.searchParams.set('offset', 100 * page);
+            uri.searchParams.set('order[createdAt]', 'asc');
+            if (nextAt) uri.searchParams.set('createdAtSince', nextAt);
+            const request = new Request(uri);
+            let data3 = await this.fetchJSON(request, 3);
+            await this.wait(this.throttleGlobal);
+            tmp = [...tmp, ...data3.data];
+            if (page == pages) {
```
27136773: Condition is redundant, it is expected that `page == pages` is always the case after the `for` loop

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -47,66 +47,59 @@ export default class MangaDex extends Connector {
...
+                return {
+                    data: tmp,
+                    nextAt: data3.data[data3.data.length - 1].attributes.createdAt.match(/^\d{4}-[0-1]\d-([0-2]\d|3[0-1])T([0-1]\d|2[0-3]):[0-5]\d:[0-5]\d/)[0],
```
27136773: Is the regex match required, or isn't `createdAt` always following this exact format?

27136773: Term `data3.data[data3.data.length - 1].attributes` may be simplified to `data3.data.pop().attributes`

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -231,4 +224,26 @@ export default class MangaDex extends Connector {
...
+    fetchJSON( request, retries ) {
+        retries = retries || 0;
+        if( typeof request === 'string' ) {
+            request = new Request( request, this.requestOptions );
+        }
+        // TODO: check if this will affect (replace) the input parameter?
+        if( request instanceof URL ) {
+            request = new Request( request.href, this.requestOptions );
+        }
+        return fetch( request )
```
27136773: Use async/await pattern instead

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -47,66 +47,59 @@ export default class MangaDex extends Connector {
...
+                    total: data3.total
+                };
+            }
+        }
+    }
+
     async _getChapters(manga) {
         let chapterList = [];
-        for(let page = 0, run = true; run; page++) {
+        for (let page = 0, run = true; run; page++) {
```
27136773: There are many auto-format changes probably done by the editing software which are not related to this PR, i suggest to make format changes in a separate PR if really necessary

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -231,4 +224,26 @@ export default class MangaDex extends Connector {
         });
         return range.join('-');
     }
+
+    fetchJSON( request, retries ) {
```
27136773: Maybe use `fetchJSON(uri, retries)` instead, and always composing the `Request` object from the given `uri` parameter within the function itself, this saves many `new Request()` instantiations before calling `fetchJSON()`

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -231,4 +224,26 @@ export default class MangaDex extends Connector {
...
+        if( typeof request === 'string' ) {
+            request = new Request( request, this.requestOptions );
+        }
+        // TODO: check if this will affect (replace) the input parameter?
+        if( request instanceof URL ) {
+            request = new Request( request.href, this.requestOptions );
+        }
+        return fetch( request )
+            .then( response => {
+                if( (response.status >= 500 || response.status == 404) && retries > 0 ) {
```
27136773: From a technical point of view retrying on `404 - not found` makes no sense (the page does not exist, so why retry), maybe check something like `/captcha.js$/.test(response.url)` instead?

src/web/mjs/connectors/MangaDex.mjs
Robonau: createdAt has some extra stuff at the end, removing it might be possible without regex (or a much simpler regex)

src/web/mjs/connectors/MangaDex.mjs
Robonau: they aren't necessary, my editor does it
undoing all of the changes is quite a bit of work

Robonau:
the mix of 503 errors and captcha-v3 is stopping me from completing a test

Robonau:
i know its an insane amount of retries, but i error-ed out with 20
ideally the 503 and captcha-v3 404 errors should be fixed removing the need for the excess retries
they are not present in browser but after poking around i couldn't find the cause

Sheepux:
Possibly trotting detection.

Robonau:
> Possibly trotting detection.

its not a throttling issue, as i said it is not present in browser (with the same delays)

Sheepux:
Well then ... have fun investigating ... 
but as soon as i set a 5s throttling i get ZERO issues while i'm able to reproduce the issue quite easily with the 200ms throttling.

Robonau:
> Well then ... have fun investigating ... but as soon as i set a 5s throttling i get ZERO issues while i'm able to reproduce the issue quite easily with the 200ms throttling.

it just takes a while for the issue to come up when every request takes 25x longer
note:neither of these issues exist in browser (errors start happening below 200ms)
and the captcha-v3 issue shouldn't exist at all, mangadex(a dev) has said so themselves

the current PR works, and is significantly faster than setting throttle to 5 seconds
im not looking in to how to stop the captcha and 503 issues because i think they are electron related

Sheepux:
Connector cannot be left/merged "as it" as bruteforcing is not acceptable.

This rate limiting is called "throttling". No it doesn't just take longuer, i'm able to get much more list before getting 404.
From what i could test they probably have a throttling/min and possibly a smaller span, or multiple rules as 404 and 503 are probably 2 different DDOS mechanism.

Further testing seems to show 30/min. So global throttling of 2100 was OK  (no 404 or 503 in 10 mins)
This value should be tested further.

+ better management of retries:
```javascript
    async fetchJSON(uri, retries) {
        let response = await fetch(request);
        for (retries = retries || 0; retries >= 0; retries--) {
            if (response.status === 404) await this.wait(15000);
            if (response.status >= 500 || /captcha-v3$/.test(response.url)) await this.wait(15000);
            if (response.status === 200) return response.json();
        }
        throw new Error(`Failed to receive content from "${request.url}" (status: ${response.status}) - ${response.statusText}`);
    }
```

> note: using recursive function is increasing memory for no value in this case. Use loops (while/for) instead

now we can reduce the amount of retries to 2 or 3 ```data100 = await this.fetchJSON(uri, 2);```
as we get a big "pause" (should set 1 min?) if we're reaching the DDOS trigger 

Seeing Ronny's comment: yes 404 is an actual throttling error. We're reducing the strain by now limiting the retries to 2 (a real 404 will persists)

27136773:
I suggest to add a user configurable throttling for scraping the manga list, with a safe default value.

Robonau:
i cant test it, but if what you say is true then this should work

Robonau:
though as i have said a throttle this large should not be necessary 
as https://api.mangadex.org/docs.html#section/Rate-limits says
the endpoint we are fetching has a 5 a second (200ms) rate limit, nothing else
this is what is seen on browser, a delay of 200 ms works fine, not a single issue at all
there is no "30/min" present in browser

```js
while (true){
    fetch("https://api.mangadex.org/manga?limit=100&offset=1200&order%5BcreatedAt%5D=asc&contentRating%5B%5D=safe&contentRating%5B%5D=suggestive&contentRating%5B%5D=erotica&contentRating%5B%5D=pornographic&createdAtSince=2019-04-02T22%3A24%3A22");
    await new Promise(r => setTimeout(r, 200));
}
```  
will run indefinatly

Robonau:
the issue is now actually fixed, at 200 ms throttle

Robonau:
ping me on discord when review done

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -11,7 +11,17 @@ export default class MangaDex extends Connector {
         this.url = 'https://mangadex.org';
         this.api = 'https://api.mangadex.org';
         this.requestOptions.headers.set('x-referer', this.url);
-        this.throttleGlobal = 100;
+        this.requestOptions.headers.set('x-sec-ch-ua', '" Not A;Brand";v="99", "Chromium";v="96", "Google Chrome";v="96"');
```
27136773, line 14: What is the purpose of this header? It is very unlikely that any `sec-*` headers have an impact on requests.

src/web/mjs/connectors/MangaDex.mjs
```diff
@@ -11,7 +11,17 @@ export default class MangaDex extends Connector {
...
+        this.config = {
+            throttle: {
```
27136773: `throttle` is a magic key which will be used when as delay when downloading images. Should be renamed to something else.

src/web/mjs/connectors/MangaDex.mjs
Robonau: that is the header that stops the captcha-v3 404 error and the 503 errors

i went through every header different from a chrome request and HN request this was the one that bade the differance

src/web/mjs/connectors/MangaDex.mjs
Robonau: i mean this should apply to downloads no?

Robonau: nvm downloading dosnt actually do any requests to the api

Robonau:
can you ping me in discord when review done

27136773 (approved):
Looks good

PR #4179, opened 2021-11-27
Reworked Connector for "onepiece-tube.com"

ProfEibe:
Added feedback from last pullRequest and here is the new version


27136773:
References: #3770

27136773 (commented):
Technically good, but not sure about adding a website that only contains a single manga as such websites were rejected before...

27136773:
If the PR is rejected, you may consider converting to a local plugin:
https://hakuneko.download/docs/install/#user-plugins

Sheepux:
I'd say we have to state : 
* what's the puprpose of this small connector ? 
* Does it provide something not available everywhere else ?

Asking the PR Author's about this.

ProfEibe:
I think I can answer both questions together.
The connected website is the only source I found where there are old and new onepiece chapters translated into german.
Somewhere else there are only the last 5 or something chapters online. So it's purpose is to provide all german translated One Piece chapters.

Sheepux:
Well. let's say ok for once.

27136773 (approved):
Looks good, Accepted

PR #4183, opened 2021-11-28
[comikey] bypass updated protection and added traps

27136773:
Review: manga-download/hakuneko-plugins@ba8310c

Added some anti-debugging traps when investigating `Comikey.mjs` in HakuNeko's through developer tools:
- collect and track user + system data (e.g. IP address)
- do some other magic stuff with the local system and network ;)


Sheepux (approved):
Looks good to me.
A small documentation comment on the commit.

PR #4220, opened 2021-12-16
bilibili quality

Robonau:
the description of the setting could probably be better, but idk what to put

### DIFF
--- a/src/web/mjs/connectors/BilibiliManhua.mjs
+++ b/src/web/mjs/connectors/BilibiliManhua.mjs
@@ -12,8 +12,8 @@ export default class BilibiliManhua extends Connector {
 
         this.config = {
             quality:  {
-                label: 'Preferred quality',
-                description: 'quality of images\nwebp is what the site displays as "raw"(lowest quality)\njpg is in the middle\npng is higher quality(and significantly larger file size)',
+                label: 'Preferred format',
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/BilibiliManhua.mjs
```diff
@@ -9,6 +9,20 @@ export default class BilibiliManhua extends Connector {
         super.label = '哔哩哔哩 漫画 (Bilibili Manhua)';
         this.tags = [ 'manga', 'webtoon', 'chinese' ];
         this.url = 'https://manga.bilibili.com';
+
+        this.config = {
+            quality:  {
+                label: 'Preferred quality',
+                description: 'quality of images\nwebp is what the site displays as "raw"(lowest quality)\njpg is in the middle\npng is higher quality(and significantly larger file size)',
```
27136773: Description might be misleading. The website will show optimized `webp` or `jpeg` (depending on browser support). I don't think such an extensive explanation is required, as most users probably know about these formats. Personally i would prefer the term `format` instead of `quality`.

27136773 (approved):
Looks good

Robonau:
simplified description but kept some notation for ppl that dont know what the formats mean

Robonau:
please review #4167 before this if possible
it is wanted by more people and has waited 2 weeks since the last change with no comments

PR #4221, opened 2021-12-16
Update GeassScan

Lyem:


27136773:
Is there no icon for `Ssgt Scan`?

Lyem:
there is not

27136773:
> there is not

You can copy and rename the default icon from e.g. `AntisenseScan`

27136773 (approved):
Looks good

PR #4223, opened 2021-12-19
CrunchyManga: Added Clipboard Support 

amit34521:
Adds clipboard copy support to plugin as Mangalist api is broken

### DIFF
--- a/src/web/mjs/connectors/CrunchyManga.mjs
+++ b/src/web/mjs/connectors/CrunchyManga.mjs
@@ -104,11 +104,11 @@ export default class CrunchyManga extends Crunchyroll {
 
         let title;
         let seriesId;
-        let hasSeriesId = /series_id:(\d+)/g.exec(responseDataText)
+        let hasSeriesId = /series_id:(\d+)/g.exec(responseDataText);
 
         if (hasSeriesId) {
-            title = response.querySelector('div.queue-wrapper > a').title.trim()
-            seriesId = hasSeriesId[1]
+            title = response.querySelector('div.queue-wrapper > a').title.trim();
+            seriesId = hasSeriesId[1];
         } else {
             let chapterLink = uri.href.replace('comics/manga', 'manga').replace('volumes', 'read/1');
             const chapterUrl = new Request(new URL(chapterLink, this.url), this.requestOptions);
--- a/src/web/mjs/connectors/CrunchyManga.mjs
+++ b/src/web/mjs/connectors/CrunchyManga.mjs
@@ -100,24 +100,23 @@ export default class CrunchyManga extends Crunchyroll {
         await this._login();
         const request = new Request(uri, this.requestOptions);
         let response = await this.fetchDOM(request);
-        let responseData = await fetch(request);
-        let responseDataText = await responseData.text();
+        let responseDataText = response.innerHTML;
 
         let title;
         let seriesId;
         let seriesIdList = [...responseDataText.matchAll(/series_id:(\d+)/g)];
-
-        if (seriesIdList.length > 0) {
-            title = response.querySelector('title').innerText.replace('Crunchyroll - ', '');
-            seriesId = parseInt([...responseDataText.matchAll(/series_id:(\d+)/g)][0][1]);
+        if(seriesIdList.length > 0)
+        {
+            title = response.querySelector('div.queue-wrapper > a').title.trim();
+            seriesId = parseInt(seriesIdList[0][1]);
         }
-        else {
+        else
+        {
             let chapterLink = uri.href.replace('comics/manga', 'manga').replace('volumes', 'read/1');
             const chapterUrl = new Request(new URL(chapterLink, this.url), this.requestOptions);
             let chapterResponse = await this.fetchDOM(chapterUrl);
-            let chapterResponseData = await fetch(chapterUrl);
-            let chapterResponseDataText = await chapterResponseData.text();
-            title = chapterResponse.querySelector('#template_body > section > header > h1 > a').innerText;
+            let chapterResponseDataText = chapterResponse.innerHTML;
+            title = chapterResponse.querySelector('section.manga-reader-page a').innerText.trim();
             seriesId = parseInt([...chapterResponseDataText.matchAll(/seriesId:\s"(\d+)/g)][0][1]);
         }
         return new Manga(this, seriesId, title);
--- a/src/web/mjs/connectors/CrunchyManga.mjs
+++ b/src/web/mjs/connectors/CrunchyManga.mjs
@@ -100,23 +100,10 @@ export default class CrunchyManga extends Crunchyroll {
         await this._login();
         const request = new Request(uri, this.requestOptions);
         let response = await this.fetchDOM(request);
-        let responseDataText = response.innerHTML;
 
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/CrunchyManga.mjs
```diff
@@ -1,12 +1,14 @@
 import Crunchyroll from './templates/Crunchyroll.mjs';
+import Manga from '../engine/Manga.mjs';
 
 export default class CrunchyManga extends Crunchyroll {
 
     constructor() {
         super();
+        this.url = "https://www.crunchyroll.com";
```
Robonau, line 8: thsi is already set in Crunchyroll.mjs that this extends from so shouldn't be necessary

Robonau:
lint

src/web/mjs/connectors/CrunchyManga.mjs
```diff
@@ -88,9 +90,36 @@ export default class CrunchyManga extends Crunchyroll {
...
             decrypted[index] ^= key;
         }
         return decrypted;
     }
+
+    async _getMangaFromURI(uri) {
+        await this._login();
+        const request = new Request(uri, this.requestOptions);
+        let response = await this.fetchDOM(request);
+        let responseData = await fetch(request);
```
Robonau: you should only need to fetch once, since you can get the text out of the DOM
example: 
```js
let responseDataText = response.innerHTML
```

src/web/mjs/connectors/CrunchyManga.mjs
```diff
@@ -88,9 +90,36 @@ export default class CrunchyManga extends Crunchyroll {
...
+        let responseDataText = await responseData.text();
+
+        let title;
+        let seriesId;
+        let seriesIdList = [...responseDataText.matchAll(/series_id:(\d+)/g)];
+
+        if (seriesIdList.length > 0) {
+            title = response.querySelector('title').innerText.replace('Crunchyroll - ', '');
```
Robonau:
```js
response.querySelector('div.queue-wrapper > a').title.trim()
```  
is probably better than a query and a replace
using a  .trim() is best practice

src/web/mjs/connectors/CrunchyManga.mjs
```diff
@@ -88,9 +90,36 @@ export default class CrunchyManga extends Crunchyroll {
...
+            seriesId = parseInt([...responseDataText.matchAll(/series_id:(\d+)/g)][0][1]);
```
Robonau: seriesId = parseInt(seriesIdList[0][1]);

src/web/mjs/connectors/CrunchyManga.mjs
```diff
@@ -88,9 +90,36 @@ export default class CrunchyManga extends Crunchyroll {
...
+    async _getMangaFromURI(uri) {
+        await this._login();
+        const request = new Request(uri, this.requestOptions);
+        let response = await this.fetchDOM(request);
+        let responseData = await fetch(request);
+        let responseDataText = await responseData.text();
+
+        let title;
+        let seriesId;
+        let seriesIdList = [...responseDataText.matchAll(/series_id:(\d+)/g)];
```
Robonau:
```js
/series_id:(\d+)/g.exec(responseDataText)
```
seems more reasonable since you only ever use the first match anyways
(the variable name would then be changed since its no longer a list)

(the if would have to changed to `if (seriesIdList) {`)

Robonau:
i cant merge anyways but this stuff would probably be picked up by sheep or ronny

amit34521:
> i cant merge anyways but this stuff would probably be picked up by sheep or ronny

I'll fix the comments. Anyways I just copied most part from viz connector as I don't know how to write the code.

src/web/mjs/connectors/CrunchyManga.mjs
amit34521: Couldn't figure this out so left it as is

Robonau:
```js
        let hasSeriesId = /series_id:(\d+)/g.exec(responseDataText)

        if (hasSeriesId) {
            title = response.querySelector('div.queue-wrapper > a').title.trim()
            seriesId = hasSeriesId[1]
        }
```

Robonau:
you shouldn't mark it as resolved if you cant figure it out,
i was more than happy to explain further 
```js
        let hasSeriesId = /series_id:(\d+)/g.exec(responseDataText)

        if (hasSeriesId) {
            title = response.querySelector('div.queue-wrapper > a').title.trim()
            seriesId = hasSeriesId[1]
        }
```

src/web/mjs/connectors/CrunchyManga.mjs
```diff
@@ -88,9 +90,32 @@ export default class CrunchyManga extends Crunchyroll {
...
+        if (seriesIdList.length > 0) {
+            title = response.querySelector('div.queue-wrapper > a').title.trim();
+            seriesId = parseInt(seriesIdList[0][1]);
+        } else {
+            let chapterLink = uri.href.replace('comics/manga', 'manga').replace('volumes', 'read/1');
+            const chapterUrl = new Request(new URL(chapterLink, this.url), this.requestOptions);
+            let chapterResponse = await this.fetchDOM(chapterUrl);
+            let chapterResponseDataText = chapterResponse.innerHTML;
+            title = chapterResponse.querySelector('section.manga-reader-page a').innerText.trim();
+            seriesId = parseInt([...chapterResponseDataText.matchAll(/seriesId:\s"(\d+)/g)][0][1]);
```
Robonau: you should be able to use exec here too
```js
seriesId = /seriesId:\s"(\d+)/g.exec(chapterResponseDataText)[1]
```
the parseInt shouldn't be necessary

amit34521:
also any way to fix login, everytime it starts session, does login and i am getting blocked by continuous repetition of this step.

Robonau:
> also any way to fix login, everytime it starts session, does login and i am getting blocked by continuous repetition of this step.

because of how this._login() works it shouldn't do anything if it is already logged in (has a _sessionID and _authToken) so there shouldn't be any "repetition" that would get HN blocked

src/web/mjs/connectors/CrunchyManga.mjs
```diff
@@ -88,9 +90,33 @@ export default class CrunchyManga extends Crunchyroll {
...
+        for (let index in decrypted) {
             decrypted[index] ^= key;
         }
         return decrypted;
     }
+
+    async _getMangaFromURI(uri) {
+        await this._login();
+        const request = new Request(uri, this.requestOptions);
+        let response = await this.fetchDOM(request);
```
27136773, line 102: How about extracting from `fetchDOM(request, 'span[group_id]')`?
```html
<span id="sharing_add_queue_button" group_id="527"></span>
```

27136773:
`lint` error is [blocking](https://github.com/manga-download/hakuneko/pull/4223/commits/fae5e9d5ef379993631f4952c73ee207aca5e601#diff-7b73c940387779623e52470e7ad557a522ecb954c26a3784307ca6296a41fab6R106) merge

27136773 (approved):
Looks good

PR #4256, opened 2021-12-28
baozimh added

gammesB:


src/web/mjs/connectors/Baozimh.mjs
```diff
@@ -0,0 +1,47 @@
...
+        super();
+        super.id = 'baozimh';
+        super.label = '包子漫書 (baozimh)';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.baozimh.com/';
+    }
+	
+    async _getMangaFromURI(uri) {
+        const request = new Request(uri, this.requestOptions);
+		const data = await this.fetchDOM(request, '.comics-detail__title'); //title object
```
27136773: Prevent mixing spaces with tabs for indentation

src/web/mjs/connectors/Baozimh.mjs
```diff
@@ -0,0 +1,47 @@
...
+import Manga from '../engine/Manga.mjs';
+
+export default class Baozimh extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'baozimh';
+        super.label = '包子漫書 (baozimh)';
+        this.tags = [ 'manga', 'webtoon', 'chinese' ];
+        this.url = 'https://www.baozimh.com/';
```
27136773: May remove trailing `/` slash

src/web/mjs/connectors/Baozimh.mjs
```diff
@@ -0,0 +1,47 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class Baozimh extends Connector {
```
27136773, line 4: Is this website not related to the `SinMH` template which is used for other websites?

src/web/mjs/connectors/Baozimh.mjs
```diff
@@ -0,0 +1,47 @@
...
+
+    async _getMangas() {
+        let msg = 'This website does not provide a manga list, please copy and paste the URL containing the images directly from your browser into HakuNeko.';
+        throw new Error(msg);
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(new URL(manga.id, this.url), this.requestOptions);
+		//where can it find the chapters
+		const dom = await this.fetchDOM(request);
```
27136773: May use query directly as second parameter `this.fetchDOM(request, '.l-box #chapter-items > div > a, .l-box #chapters_other_list > div > a')`

src/web/mjs/connectors/Baozimh.mjs
```diff
@@ -0,0 +1,47 @@
...
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.text.trim()
+            };
+        });
+    }
+	async _getPages(chapter) {
+        let pageLinks = [];
+        let request = new Request(new URL(chapter.id, this.url), this.requestOptions);
+	    let data = await this.fetchDOM(request, '.comic-contain amp-img noscript');
+		data = data.map(element => element.innerText.replace(/.*src="([^"]*)".*/, '$1'));
```
27136773: no need to re-assign, return directly `return data.map(...);`

27136773:
Many lint errors that needs to be addressed.

src/web/mjs/connectors/Baozimh.mjs
```diff
@@ -0,0 +1,47 @@
...
+	
+    async _getMangaFromURI(uri) {
+        const request = new Request(uri, this.requestOptions);
+		const data = await this.fetchDOM(request, '.comics-detail__title'); //title object
+        const id = uri.pathname;
+		let title = data[0].textContent.trim()
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
```
27136773, line 21: Why not providing the list which is directly shown on the website https://www.baozimh.com?

src/web/mjs/connectors/Baozimh.mjs
gammesB: O how can you tell? I don't know how to see that
I tried changing it to SinMH but then I still need the custom getChapters/getPages/getMangaFormURI

src/web/mjs/connectors/Baozimh.mjs
gammesB: I didn't add it as I never used that feature in hakuneko and saw that a lot of other connectors also didn't have it.
I will try to add it when I fix the other things from your review

gammesB: I checked it out. they work with pagination 
https://www.baozimh.com/classify?page=1
is the url.
the problem is that they are a aggrigator and even when i was at page 41 in my browser it still went on, it has even more then 700 pages with 36 manhua's each. making 700 requests would not be good for the user experience even if the website doesn't block making so much request to the website.

Sheepux:
Make sure to close the "Resolve conversation" if you've solved the issue and also click the re-request review to let us know the PR is read for review again :)

27136773 (approved):

27136773 (approved):

27136773:
Conflicts from a previous PR needs to be resolve before, please sync your branch with the `master` of the upstream repository.

gammesB:
Sorry, @ronny1982 fixed. Seems the other PR I made for Comicrise made the conflicte.

PR #4295, opened 2022-01-09
[new connector] MangaEffect 

Albirew:
just a copy of another WordPressMadara mjs file witch i edited accordingly


27136773 (approved):
Looks good

PR #4314, opened 2022-01-17
Update chapter regex

Zarh:
Add "Page", "N°" and "épisode" flag to detect chapter number used in mangaplus.shueisha.co.jp


27136773 (approved):
Looks good

PR #4336, opened 2022-01-23
[MangaReaderTo] add website support

### LINKED ISSUE #3979: [Site Request] MangaReader
**Name of the website**
MangaReader.io

- Site: https://mangareader.to
- Manga List: https://mangareader.to/az-list
- Manga example: https://mangareader.to/vagabond-4
- Chapter online viewer example: https://mangareader.to/read/vagabond-4/en/chapter-1

**Languages**
List of languages supported by the website: English

**Website relationship*
Its related to zoro.to which is an anime streaming site.

**Additional details**
This website's having two options - we can either read in the format of volumes or chapters.

manh21:
- Resolved #3979


Robonau:
i believe this is the issue?
https://discord.com/channels/581629183159959582/581629183159959584/934702834715336724
might want to explain

manh21:
Stuck at deobfuscating the raw [image](https://c-1.mreadercdn.ru/_v2/1/0dcb8f9eaacfd940603bd75c7c152919c72e45517dcfb1087df215e3be94206cfdf45f64815888ea0749af4c0ae5636fabea0abab8c2e938ab3ad7367e9bfa52/02/38/0238578217af660c5c0da114f876d39d/0238578217af660c5c0da114f876d39d_1024.jpeg?t=4a42d27bfa0548549655a3f49f0d2e41&ttl=1642963098).

image

Ever4engel:
some of the content isn't obfuscated

manh21:
Note: Bad idea

`await Engine.Request.fetchUI(request, script, 60000, true)` inside `_handleConnectorURI(payload)` resulted in degraded performance and cause software hang. The hang were caused by multiple spawn of background browser.

Tested System Info:
OS: Windows 10
CPU: AMD Ryzen 5
RAM: 16 GB
GPU: GTX 1050 ti

```js
        payload = payload.split('@%%%@');
        const image_url = payload[0];
        const chapter_id = payload[1];
        const uri = new URL(chapter_id, this.url);
        const script = `
                (async () => {
                    try {
                        const data = ${JSON.stringify(image_url)};
                        const canvas = await imgReverser(data);
                        const uri = canvas.toDataURL('image/jpeg')
                        resolve(uri);
                    } catch(error) {
                        resolve(error);
                    }
                })()
            });
        `;
        const images = await Engine.Request.fetchUI(request, script, 60000, true);
        const data = images.replace(/^data:image\/\w+;base64,/, "");
        const buf = Buffer.from(data, "base64");
        return {
            mimeType: 'image/jpeg',
            data: buf
        };
    }
```

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+
+export default class MangaReaderTo extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'mangareaderto';
+        super.label = 'MangaReader.to';
+        this.tags = ['manga', 'english'];
+        this.url = 'https://mangareader.to';
+        this.path = '/az-list?page=%PAGE%';
```
27136773: May use `path = '/az-list?page=';`

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+        const mangaList = [];
+        for(let page = 1, run = true; run; page++) {
+            const mangas = await this._getMangasFromPage(page);
+            mangas.length ? mangaList.push(...mangas) : run = false;
+        }
+        return mangaList;
+    }
+
+    async _getMangasFromPage(page) {
+        const uri = new URL(this.path.replace('%PAGE%', page), this.url);
```
27136773: May just concat to simplified path `new URL(this.path + page, this.url);`

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+        });
+    }
+
+    async _getChapters(manga) {
+        let uri = new URL(manga.id, this.url);
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, this.queryChapters);
+        return data.map(element => {
+            const link = this.getRootRelativeOrAbsoluteLink(element, this.url);
+            const title = element.title.replace(/:(.*)/gi, '');
```
27136773, line 57: Should the regex really be greedy in case of multiple `:` within the title?

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+                language
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        const uri = new URL(chapter.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, this.queryPages);
+        const readingId = data[0].getAttribute('data-reading-id');
```
27136773: Mat use dataset `data[0].dataset.readingId`

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+
+        const script = await this._getImages(readingId);
+        const images = await Engine.Request.fetchUI(request, script, 60000, true);
+
+        return images.map(page => this.createConnectorURI(page));
+    }
+
+    async _getImages(readingId) {
```
27136773: Method name is not intuitive, as a developer i would expect it to return a list of images or image links based on its name, but it returns a javascript.

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+
+        const dom = this.createDOM(data.html);
+        const images = Array.from(dom.querySelectorAll('.iv-card'));
+
+        // Example: https://c-1.mreadercdn.ru/_v2/1/0dcb8f9eaacfd940603bd75c7c152919c72e45517dcfb1087df215e3be94206cfdf45f64815888ea0749af4c0ae5636fabea0abab8c2e938ab3ad7367e9bfa52/9d/32/9d32bd84883afc41e54348e396c2f99a/9d32bd84883afc41e54348e396c2f99a_1200.jpeg?t=4b419e2c814268686ff05d2c25965be9&amp;ttl=1642926021
+        const imageData = JSON.stringify(images.map(image => image.getAttribute('data-url')));
+
+        return `
+            new Promise((resolve, reject) => {
+                (async () => {
```
27136773: Redundant, just make the outer function async `new Promise(async (resolve, reject) => ...`

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+                        let images = [];
+                        const data = ${imageData};
+                        for(const d of data) {
+                            const canvas = await imgReverser(d);
+                            const uri = canvas.toDataURL('image/jpeg')
+                            images.push(uri);
+                        }
+                        resolve(images);
+                    } catch(error) {
+                        resolve(error);
```
27136773: Shouldn't this be `reject(error)`?

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+
+        return `
+            new Promise((resolve, reject) => {
+                (async () => {
+                    try {
+                        let images = [];
+                        const data = ${imageData};
+                        for(const d of data) {
+                            const canvas = await imgReverser(d);
+                            const uri = canvas.toDataURL('image/jpeg')
```
27136773: Aren't data URLs limited to max. 2MB or something like that in Chrome?
Use the descrambling format from settings that were configured by the user:
`Engine.Settings.recompressionFormat.value`
and
`parseFloat(Engine.Settings.recompressionQuality.value)/100`

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    async _getPages(chapter) {
+        const uri = new URL(chapter.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, this.queryPages);
+        const readingId = data[0].getAttribute('data-reading-id');
+
+        const script = await this._getImages(readingId);
+        const images = await Engine.Request.fetchUI(request, script, 60000, true);
```
27136773: Can't this be done in `_handleConnectorURI(payload)` instead by just using a single `script` per image as payload?
Would be more memory efficient (at the cost of performing more `fetchUI` calls) and as far as i can see the lease time for the image links in the scripts are ~2 hours which should be enough for that purpose.

src/web/mjs/connectors/MangaReaderTo.mjs
```diff
@@ -0,0 +1,121 @@
...
+
+        return images.map(page => this.createConnectorURI(page));
```
27136773: If images shall be data URLs, they should be usable directly (no need for `_handleConnectorURI `)

src/web/mjs/connectors/MangaReaderTo.mjs
manh21: `await Engine.Request.fetchUI(request, script, 60000, true)` inside `_handleConnectorURI(payload)` resulted in degraded performance and cause software hang. The hang were caused by multiple spawn of background browser.
 
Tested System Info:
OS: Windows 10
CPU: AMD Ryzen 5
RAM: 16 GB
GPU: GTX 1050 ti

src/web/mjs/connectors/MangaReaderTo.mjs
manh21: Their chapter list follow this format `Chapter 1: blahblah`

image

src/web/mjs/connectors/MangaReaderTo.mjs
manh21: yes, data URL have 2MB limit in chrome

27136773 (approved):

Jaxk69:
Hey, @manh21 i wanna submit some feedback .
So, the manga page for e.g. https://mangareader.to/horimiya-121  has two list one for chapters and one for volumes.
The chapter list can contain fan or unofficial translations but the volume one only contains the official english/Japanese translations. But it only shows the chapter list and not the volume list in Hakuneko. Why is that ?

PR #4343, opened 2022-01-25
Mangago Add website Request 

luluchan93:
Is it possible to add mangago to the web sites list Please ?


Sheepux:
Don't use pull request, use issues.

PR #4358, opened 2022-01-29
Better identify the languages in Manga Plus.

apiweb:
Fix the language identifier because now, MangaPlus publish mangas in several languages, instead of English and Spanish only.

Before
image

After
image

### DIFF
--- a/src/web/mjs/connectors/ShueishaMangaPlus.mjs
+++ b/src/web/mjs/connectors/ShueishaMangaPlus.mjs
@@ -15,22 +15,10 @@ export default class ShueishaMangaPlus extends Publus {
     }
 
     _getLanguage(language) {
-        switch (language) {
-            case 1:
-                return ' [es]';
-            case 2:
-                return ' [fr]';
-            case 3:
-                return ' [id]';
-            case 4:
-                return ' [pt-br]';
-            case 5:
-                return ' [ru]';
-            case 6:
-                return ' [th]';
-            default:
-                return ' [en]';
-        }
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/ShueishaMangaPlus.mjs
```diff
@@ -14,14 +14,33 @@ export default class ShueishaMangaPlus extends Publus {
         this.rootType = 'MangaPlus.Response';
     }
 
+    _getLanguage(language) {
+        switch (language) {
```
27136773: It is recommend to use a `map` instead of `switch ... case` when possible, e.g.
```javascript
function _getLanguages(language) {
    const languages = {
        1: '[es]', 2: '[fr]', // ...
    };
    return languages[language] || '[en]';
}

```

27136773 (approved):
Looks good, improvements to meet clean code recommendations possible

src/web/mjs/connectors/ShueishaMangaPlus.mjs
apiweb: Thanks for the feedback. I changed the code.

Zarh:
it wouldn't be better to use the language filter (combobox) instead of tags in the title ? Like it's done with mangadex for example.

Sheepux:
doesn't work if it's not the same mangaid, for mangadex languages are within the same mangaID. On mangaplus i'm guessing that it's considered as 2 separate mangas

PR #4409, opened 2022-02-19
Fixed #4402

### LINKED ISSUE #4402: [Exhentai] Option to use higher quality images
Example
Gallery: `https://exhentai.org/g/2056537/d12611e8c1/`
Image: `https://exhentai.org/s/750f8a5d05/2056537-1`
At the bottom it says: `Download original 2889 x 4075 6.81 MB source`
The image displayed in web view is: `1280 x 1805 :: 214.9 KB`
The image downloaded through hakuneko is: `1280 x 1805 :: 214.9 KB`
What was expected: `2889 x 4075 :: 6.81 MB`

If it helps, there's Userscript available that download galleries in zip, and it does download the real original size.
https://github.com/ccloli/E-Hentai-Downloader

kusomeru:
So, here's what was the problem.
`let data = await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]');`
`fetchDOM` then does `querySelectorAll` 
Then we take `data[0]`
The problem is `querySelectorAll` returns NodeList not in order we call them, but in order it first appears in DOM itself, which means `data[0]` will ALWAYS be low res `source#img`
AFAIK the only fix is either two calls or sorting them afterwards.

Tested, works fine.

https://github.com/manga-download/hakuneko/issues/4402

### DIFF
--- a/src/web/mjs/connectors/EHentai.mjs
+++ b/src/web/mjs/connectors/EHentai.mjs
@@ -59,9 +59,7 @@ export default class EHentai extends Connector {
 
     async _handleConnectorURI(payload) {
         let request = new Request(payload, this.requestOptions);
-        let dataFull = await this.fetchDOM(request, 'a[href*="fullimg.php"]');
-        let dataImg = await this.fetchDOM(request, 'source#img');
-        let data = [...dataFull, ...dataImg];
+        let data = await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]').reverse();
         let response = await fetch(this.getAbsolutePath(data[0], request.url), this.requestOptions);
         if(!response.headers.get('content-type').startsWith('image/')) {
             response = await fetch(this.getAbsolutePath(data[0], request.url), this.requestOptions);
--- a/src/web/mjs/connectors/EHentai.mjs
+++ b/src/web/mjs/connectors/EHentai.mjs
@@ -59,7 +59,9 @@ export default class EHentai extends Connector {
 
     async _handleConnectorURI(payload) {
         let request = new Request(payload, this.requestOptions);
-        let data = await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]');
+        let dataFull = await this.fetchDOM(request, 'a[href*="fullimg.php"]');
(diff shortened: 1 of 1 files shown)


27136773:
If we know it is always the last image in the website, can't we just use `data.pop()` instead of `data[0]`?

kusomeru:
> If we know it is always the last image in the website, can't we just use `data.pop()` instead of `data[0]`?

1. Sometimes there are no original image available so it will only return web version
2. My goal wasn't to rewrite the original code that much

If you don't like my implementation alternatively it can be done by reversing array
`let data = await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]').reverse();`

27136773 (dismissed):

src/web/mjs/connectors/EHentai.mjs
```diff
@@ -59,7 +59,7 @@ export default class EHentai extends Connector {
 
     async _handleConnectorURI(payload) {
         let request = new Request(payload, this.requestOptions);
-        let data = await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]');
+        let data = await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]').reverse();
```
27136773: Add brackets around `await` before calling `reverse()`
```javascript
const data = (await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]')).reverse()
```

kusomeru:
Yeah, sorry I'm shit at coding.

src/web/mjs/connectors/EHentai.mjs
```diff
@@ -59,7 +59,7 @@ export default class EHentai extends Connector {
 
     async _handleConnectorURI(payload) {
         let request = new Request(payload, this.requestOptions);
-        let data = await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]');
+        let data = (await this.fetchDOM(request, 'source#img, a[href*="fullimg.php"]')).reverse()
```
27136773: Please add semicolon as suggested by automated code inspection

kusomeru:
Added

27136773 (approved):

PR #4412, opened 2022-02-20
merge from master

Sheepux:
6.1.7 Release after commit b89d910  2022-01-14


27136773 (approved):
Commit order seems to be shown wrong in PR, but it looks like all 20 new commits since last merge are correctly applied

PR #4458, opened 2022-03-05
Fix user-agent value for CoreView connector

littmus:
There is weird problem that current CoreView connect cannot load chapters properly for recently updated manga.
I investigated the reason and found that user-agent value of request headers was problem.
Checked with MagComi(https://magcomi.com/), especially this manga (https://magcomi.com/episode/3269754496670883221)

Request with previous user-agent value
img003
img004
Response has only two entries.

Request with new user-agent value (Windows Chrome's user-agent value)
img001
img002
This has 3 entries normally.

### DIFF
--- a/src/web/mjs/connectors/templates/CoreView.mjs
+++ b/src/web/mjs/connectors/templates/CoreView.mjs
@@ -9,9 +9,9 @@ export default class CoreView extends Connector {
         super.label = undefined;
         this.tags = [];
         this.url = undefined;
-        this.requestOptions.headers.set('x-user-agent', 'Mozilla/5.0 (iPod; CPU iPhone OS 14_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/87.0.4280.163 Mobile/15E148 Safari/604.1');
+        this.requestOptions.headers.set('x-user-agent', 'Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.80 Safari/537.36');
         // The URL pathes to parse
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/templates/CoreView.mjs
```diff
@@ -9,9 +9,9 @@ export default class CoreView extends Connector {
         super.label = undefined;
         this.tags = [];
         this.url = undefined;
-        this.requestOptions.headers.set('x-user-agent', 'Mozilla/5.0 (iPod; CPU iPhone OS 14_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/87.0.4280.163 Mobile/15E148 Safari/604.1');
+        this.requestOptions.headers.set('x-user-agent', 'Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.80 Safari/537.36');
```
27136773, line 12: What happens when completely remove forced user agent and use the automatically assigned user agent instead?

27136773:
Does this change work with all `CoreView` based websites?

src/web/mjs/connectors/templates/CoreView.mjs
littmus: The problem cannot be reproduced now because it loads chapters normally over time with previous user agent.
Next manga's update date is 3/10, so I'll check it again then.

littmus:
> 
I'll check if the same problem occurs in other CoreView based websites(ShonenJumpPlus, ShonenMagazinePocket, etc) right after the day changes.

Ever4engel:
This pr is useless, need a deletion i think

27136773:
Closed due to inactivity

PR #4462, opened 2022-03-06
It would be nice to change font size

stadros83:
Not sure if I'm in the right place but it would be nice to be able to change the font size inside the app ?
I've got a lot of mangas and it's kinda difficult to select a single one when using a touchscreen because the names are to small.


Sheepux:
create a feature request, not a merge request.
https://github.com/manga-download/hakuneko/issues/new/choose

Issue #2279, opened 2020-09-13
[Feature] mangadex releases order


Albirew:
**Is your feature request related to a problem? Please describe.**
more a enhancement than a problem, having to choose language then scroll all the way down just to see if there's an update or not is a pain

**Describe the solution you'd like**
having reversed order just for mangadex connector so newest releases are up like every other connectors instead of down would be nice

**Context (please complete the following information):**
nightly master@a5c135

27136773:
Toggle the order button on top of the chapter list

Albirew:
order button does not care of release time order, meaning it will shambles releases then have "vol.xx chapter 120" at end of list and would only show "vol.10 chapter 119" as latest

27136773:
So the request is to add a fourth state, because MangaDex naming convention is inconsistent:
[ original, ascending, descending ] => [ original, **reversed**, ascending, descending ]

Albirew:
I don't know if they became consistent or if the fix have been lost but the sort order got reverted again on mangadex.

PR #4547, opened 2022-04-07
[Mangadex] reverse order to keep consistent

Albirew:
only mangadex listed latest chapters on bottom line.
this fix it. (#2279)


Albirew (commented):
what the!? i did not added this at line 241... github web editor?

27136773 (approved):

Issue #1805, opened 2020-05-29
[Site Request] GuyaMoe 


Ever4engel:
**Name of the website**
Guya Moe

**Website urls (examples below)**
- Site: https://guya.moe
- Manga List: https://guya.moe/read/manga/Kaguya-Wants-To-Be-Confessed-To/
- Manga example: https://guya.moe/read/manga/Kaguya-Wants-To-Be-Confessed-To/
- Chapter online viewer example: https://guya.moe/read/manga/Kaguya-Wants-To-Be-Confessed-To/190/1/

**Languages**
English (gb)

**Website relationship**
////////

**Additional details**
[FMD support](https://github.com/fmd-project-team/FMD/commit/5c7a0c43c6d1394094014a0c278531f45180afdf)

27136773:
I think we should slow down adding websites that only provide a single manga, especially when there are direct download links or when the same images are already available at the corresponding Scanlator websites.

See also: #1799 (which includes Kaguya Sama as well)

Ever4engel:
good idea !

you should add ""low priority"" tag for this type of websites.

Sheepux:
My point of view : It's not even a low priority. It's a reject by hakuneko policy :  there is no ads, no tracers, a download option. There is absolutly no need to implement this connector.

27136773:
OK, lets close it...

PR #4555, opened 2022-04-10
Added guya.moe connector

AntarEspadas:


src/web/mjs/connectors/GuyaMoe.mjs
```diff
@@ -0,0 +1,134 @@
...
+        super.id = 'guya-moe';
+        super.label = 'Guya.moe';
+        this.tags = ['manga', 'english'];
+        this.url = 'https://manga.guya.moe';
+    }
+
+    async _getMangas() {
+        const request = new Request(this.url, this.requestOptions);
+        /** @type {HTMLDivElement[]} */
+        const [data] = await this.fetchDOM(request, '.dropdown-menu');
```
27136773, line 17: May combine both CSS queries into one, e.g. `'.dropdown-menu a'`

src/web/mjs/connectors/GuyaMoe.mjs
```diff
@@ -0,0 +1,134 @@
...
+        const anchors = data.querySelectorAll('a');
+        const result = [];
+        anchors.forEach(a => {
```
27136773, line 20: May use `map` instead of pushing into result, e.g.
`const results = anchors.map(a => { /* return manga result instead of push */ })`

src/web/mjs/connectors/GuyaMoe.mjs
```diff
@@ -0,0 +1,134 @@
...
+        });
+        return result;
+    }
+
+    /**
+     * @param {{id: string, title: string}} manga
+     */
+    async _getChapters(manga) {
+        const request = new Request(`${this.url}/api/series/${manga.id}`, this.requestOptions);
+        const response = await fetch(request);
```
27136773, line 38: May use `this.fetchJSON` for more convenience

src/web/mjs/connectors/GuyaMoe.mjs
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param {Chapter} chapter
+     */
+    async _getPages(chapter) {
+        const request = new Request(`${this.url}/api/series/${chapter.manga.id}`, this.requestOptions);
+        const response = await fetch(request);
+        /**@type {ApiResponse} */
+        const data = await response.json();
+        const [pagesUrls, group] = getPagesInfo(data, chapter);
+        const pages = pagesUrls.map(url => {
```
27136773, line 60: May ommit unused local variable `pages` and return the mapped array directly

src/web/mjs/connectors/GuyaMoe.mjs
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param {URL} uri
+     */
+    async _getMangaFromURI(uri) {
+        const start = '/read/manga/';
+        const index = uri.pathname.indexOf(start);
+        let id = uri.pathname.substring(index + start.length);
+        id = id.substring(0, id.indexOf('/'));
+        const request = new Request(`${this.url}/api/series/${id}`, this.requestOptions);
+        const response = await fetch(request);
```
27136773, line 76: May use `this.fetchJSON`

27136773:
Personally i would remove all the type definition comment blocks. To prevent the discovered type information from being lost, it might be useful to also add this as [legacy connector](https://github.com/manga-download/haruneko/tree/master/src/engine/websites/legacy) in HaruNeko

27136773:
Just found that `gyua.moe` was rejected some time ago... #1805
Maybe consider this to install as [local user plugin](https://hakuneko.download/docs/install/#user-plugins) (slight changes required) and feel free to share it with the community...

AntarEspadas:
Alright. Is there anything like a forum or a repo where I might share the resulting plugin?

27136773:
> Alright. Is there anything like a forum or a repo where I might share the resulting plugin?

Maybe create a new folder in [manga-download/hakuneko-userland](https://github.com/manga-download/hakuneko-userland)

27136773:
Closed for the reason stated in the comments

PR #4556, opened 2022-04-10
fix iqiyi

gammesB:
iqiyi stopped working. first needed to fix title.
after that noticed that getchapters would only give the first and last 20. now been fixed too.
example url: https://www.iqiyi.com/manhua/detail_18yzmcjr51.html


27136773 (approved):
Looks good to me

PR #4564, opened 2022-04-13
Improve parallelization of Hentai2Read connector

Bitverdreher:
Yaaz

### DIFF
--- a/src/web/mjs/connectors/Hentai2Read.mjs
+++ b/src/web/mjs/connectors/Hentai2Read.mjs
@@ -24,8 +24,9 @@ export default class Hentai2Read extends Connector {
         const request = new Request(uri, this.requestOptions);
         const data = await this.fetchDOM(request, 'ul.pagination li:nth-last-child(2) a');
         const pageCount = parseInt(data[0].text.trim());
-        for(let page = 1; page <= pageCount; page++) {
-            let mangas = await this._getMangasFromPage(page);
+        const mangaPromises = Array.from(new Array(pageCount), (_, index) => this._getMangasFromPage(index + 1));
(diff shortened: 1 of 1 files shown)


27136773:
The website has a paginated manga list with `754` pages. Creating so many requests at the same time is a very bad idea for the following reasons:
- HakuNeko may become unresponsive
- Users may hit a server side rate limit with worst case of IP ban

Therefore i would suggest to abandon this change

Bitverdreher:
Technically, chromium should restrict the amount of requests sent to one host at a time - so in practice it would 'only' mean, that the load for the server will only increase by a factor of 6 - that should also be the speedup.

27136773:
I tested it in HakuNeko by injecting
```javascript
(async function() {
	const promises = new Array(750).fill().map((_, index) => fetch(`https://hentai2read.com/hentai-list/all/any/all/name-az/${index + 1}/`));
	await Promise.all(promises);
})();
```

**PRO:**
The saved time is really astonishing, it is like 100 times faster then scraping page py page.

**CON:**
On the other hand chrome blocked any other request while the manga list was syncing, due to chromes internal scheduler (other requests such as browsing chapters in the UI are blocked until the manga list sync was completed).

**NOTE:**
The next version of HakuNeko will have a [TaskPool](https://github.com/manga-download/haruneko/tree/master/src/engine/taskpool) with support for configurable concurrency and rate limiting, which is well suited to solve such problems.

27136773:
Dropped for HakuNeko.
The next iteration has the aforementioned ability of better control for multiple concurrent requests and rate limiting, which can be used to increase performance for manga list updates from such websites...

PR #4581, opened 2022-04-19
ComicBrise fix www urls not being found

gammesB:
ComicBrise now auto redirects users from non www to www urls(301 redirect) making users confused if they copy the urls and it doesn't work.


27136773 (approved):
Looks good

PR #4585, opened 2022-04-20
mangahere CDN Cloudflare protection

### LINKED ISSUE #4584: mangahere cant download
get this error  i can view/read via web but got this error when try to download

image

Sheepux:
Fixes #4584


27136773:
Would be better to use the connector URL mechanism, e.g. like in https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/GSNation.mjs#L14-L28

Sheepux:
managed by #4594

PR #4597, opened 2022-04-23
Corrected ReaperScansFr

onelots:
**ReaperScansFr** moved to a new template / new URL so here is the patch.

### DIFF
--- a/src/web/mjs/connectors/ReaperScansFR.mjs
+++ b/src/web/mjs/connectors/ReaperScansFR.mjs
@@ -14,3 +14,4 @@ export default class ReaperScans extends WordPressMadara {
         this.queryChapters = 'div.chapter-link > a';
         this.queryChaptersTitleBloat = 'span.chapter-release-date';
     }
+}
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/ReaperScansFR.mjs
```diff
@@ -1,13 +1,16 @@
...
-        this.url = 'https://reaperscans.fr';
-        this.path = '/manga/?list';
+        this.url = 'https://new.reaperscans.fr';
+        this.links = {
+            login: 'https://new.reaperscans.fr/login'
+        };
+        this.queryChapters = 'div.chapter-link > a';
+        this.queryChaptersTitleBloat = 'span.chapter-release-date';
     }
-}
```
Sheepux, line 13: Missing closing brace

Sheepux (changes requested):

onelots:
Yep, changed

src/web/mjs/connectors/ReaperScansFR.mjs
```diff
@@ -1,13 +1,17 @@
-import WordPressMangastream from './templates/WordPressMangastream.mjs';
+import WordPressMadara from './templates/WordPressMadara.mjs';
 
-export default class ReaperScansFR extends WordPressMangastream {
+export default class ReaperScans extends WordPressMadara {
```
Sheepux: do not change the class name, should be the same as the id and correlated to the name/label.

onelots:
Yeah, excuse me X) i've made a mistake indeed; but now, it's fine.

Sheepux (approved):
/LGTM
(no problem, that's why review are here for)

PR #4650, opened 2022-05-06
Adding eclypse scans

onelots:


src/web/mjs/connectors/Eclypse.mjs
```diff
@@ -0,0 +1,13 @@
...
+
+    constructor() {
+        super();
+        super.id = 'eclypse';
+        super.label = 'eclypse';
+        this.tags = [ 'manga', 'scanlation', 'webtoons', 'french' ];
+        this.url = 'https://eclypse-scan.com';
+    }
+}
+
```
Sheepux: Lint ;)

Sheepux (changes requested):
Lint 
Connector img name should be lowercase

src/web/mjs/connectors/Eclypse.mjs
onelots: I don't understand aha

Sheepux: check this ;)
https://github.com/manga-download/hakuneko/pull/4650/files

PR #4656, opened 2022-05-07
To corona ex

gammesB:
example https://to-corona-ex.com/comics/20000000051072


src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -0,0 +1,116 @@
...
+            };
+        });
+    }
+    async _getPages(chapter) {
+        let request = new Request(
+            "https://api.to-corona-ex.com/episodes/" + chapter.id + "/begin_reading",
+            this.requestOptions
+        );
+        const chapterinfo = await this.fetchJSON(request);
+        let drmhash = [];
```
27136773: May remove unused variable

src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -0,0 +1,116 @@
...
+        let imagelinks = [];
+        for (let i = 0; i < chapterinfo.pages.length; i++) {
+            drmhash.push(chapterinfo.pages[i].drm_hash);
+            imagelinks.push(this.createConnectorURI(chapterinfo.pages[i].page_image_url));
```
27136773: May use `.map` instead of loop

src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -0,0 +1,116 @@
...
+        }
+        return imagelinks;
+    }
+    _handleConnectorURI(payload) {
+        let promise;
+        switch (true) {
```
27136773: May use guard to fall thorugh
```javascript
if(!payload.includes('https://cdn.to-corona-ex.com/pages/images/')) {
    return super._handleConnectorURI(payload);
}
let data = await this.prepareForDescramble(payload);
data = this._blobToBuffer(data);
this._applyRealMime(data);
return data;
```

27136773:
The icon is to large, it should not exceed 16 KB and ideally hava a resolution of 64x64 px

27136773 (approved):

lasthm:
Is it possible to download monthly paid member chapters？

PR #4690, opened 2022-05-21
Removed Login need from CrunchyManga Clipboard processor

amit34521:
Can create login issues. Got Blocked many times with login. as this is not needed and fetcher still works fine, hence removed


27136773 (approved):
Looks good

PR #4722, opened 2022-06-03
Added language functionality

starring94:
* [connectors.html] Added Support for tags and connectors
* [frontend@classic-dark] Added support for languages
* [frontend@classic-light] Added support for languages
* [i18n] Added support for English (en.json)
* [i18n] Added support for Hungarian (hu.json)
* [jobs.html] Added Support for connectors
* [HakuNeko] Added LanguageManager to accessible objects
* [LanguageManager] Added a class for managing languages (similiar to Connectors)
* [menu.html] Added support for category names, setting labels, descriptions and options
* [Settings] Added selection of language (restart required to update)

### DIFF
--- a/src/web/i18n/en.json
+++ b/src/web/i18n/en.json
@@ -1,0 +1,1162 @@
+{
+    "languageCode": "en",
+    "languageName": "English (English)",
+    "frontend": {
+        "frontend@classic-light": {
+            "bookmarks": {
+                "titleListRemove": "Click to remove the selected manga from the bookmark list",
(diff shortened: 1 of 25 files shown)


27136773:
This is quite a big change, not sure if it should be added to the current version of HakuNeko which is not much maintained in case this PR may break functionality / introduce bugs...

27136773:
Will abandon this PR since risk/return ratio seems to high for me (espacially with the long-term plan to replace this with the next major iteration)

Issue #4778, opened 2022-06-24
[Manga line] Connector not working perfectly


Ever4engel:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes

**Describe the bug**
The connector work well on the mangalist that haku grabs. the problem is it's not the full mangalist.
here a little exemple: https://manga.line.me/product/periodic?id=00009ahv this manga isn't available in the mangalist. I can directly use the clipboard to see it but i can't preview neither download it (it has some free chapters) @Elastic1 and i'm pretty sure lot of mangas has the same issue.

PR #4783, opened 2022-06-25
Fixed Line Manga JP

Elastic1:
https://github.com/manga-download/hakuneko/issues/4778

> the problem is it's not the full mangalist

This site does not have a list of all manga (or they are too large to scrap). Therefore, I am listing only original manga of this site.

Non-original manga of this site seem to be scrambled. (Fixed)

### DIFF
--- a/src/web/mjs/connectors/LineMangaJP.mjs
+++ b/src/web/mjs/connectors/LineMangaJP.mjs
@@ -51,12 +51,370 @@ export default class LineMangaJP extends Connector {
 
     async _getPages(chapter) {
         const request = new Request(`${this.url}/book/viewer?id=${chapter.id}`, this.requestOptions);
-        const imgs = await Engine.Request.fetchUI(request, 'window.imgs || {}');
-        if (Object.keys(imgs).length == 0) {
+        const [images, option] = await Engine.Request.fetchUI(request, '[Object.values(window.imgs || {}), window.OPTION || {}]');
+        if (option.mediado_token) {
+            const mediadoData = await this._getMediadoData(option);
+            return mediadoData.images.map(imageData => {
+                const metaData = mediadoData.metaDataList[imageData.pageNum - 1];
+                return this.createConnectorURI({
+                    imageData,
+                    metaData,
+                });
+            });
+        }
+        if (images.length == 0) {
             throw new Error(`The chapter '${chapter.title}' is neither public, nor purchased!`);
         }
-        return Object.values(imgs)
-            .map(img => img.url)
+        return images
+            .map(image => image.url)
             .filter(url => !url.includes('inline_ads_banner'));
     }
+
+    async _handleConnectorURI(payload) {
+        const canvas = document.createElement('canvas');
+        canvas.width = payload.imageData.width;
+        canvas.height = payload.imageData.height;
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/LineMangaJP.mjs
```diff
@@ -51,12 +51,370 @@ export default class LineMangaJP extends Connector {
...
+        return this._blobToBuffer(blob);
+    }
+
+    async _getImageDataUrl(url) {
+        const uri = new URL(url);
+        const request = new Request(uri);
+        request.headers.set('x-origin', uri.origin);
+        const res = await fetch(request);
+        const blob = await res.blob();
+        return this._blobToDataUrl(blob);
```
27136773, line 93: I think data URLs have a limit of around ~2MB in chrome, so it may not work for larger images.
Never tried, but maybe Object URLs could be used as workaround (but they need to be released at some point to prevent memory leaks).

src/web/mjs/connectors/LineMangaJP.mjs
```diff
@@ -51,12 +51,370 @@ export default class LineMangaJP extends Connector {
...
+    }
+
+    // Copied from Line Manga
+    _getMediadoData(option) {
+        const contentUrl = option.mediado_contents_url;
+        const contentFile = option.mediado_contents_file;
+        const token = option.mediado_token;
+        const mediadoUrl = `${contentUrl}${contentFile}?token=${token}`;
+        NFBR.a8O.a8M();
+        return fetch(mediadoUrl, {
```
27136773: Prefer async/await pattern over (nested) promise chaining

src/web/mjs/connectors/LineMangaJP.mjs
```diff
@@ -51,12 +51,370 @@ export default class LineMangaJP extends Connector {
...
+            canvas,
+            image,
+            DummyWidth,
+            DummyHeight + offset,
+            imageFilePath
+        );
+    }
+
+    _unscrambleMediadoImage(imageData, metaData, canvas) {
+        return new Promise((resolve, reject) => {
```
27136773, line 170: Consider using async/await instead of promise chaining

src/web/mjs/connectors/LineMangaJP.mjs
```diff
@@ -51,12 +51,370 @@ export default class LineMangaJP extends Connector {
...
+            };
+            image.src = imageData.url;
+        }).then((image) => this._drawImage(canvas, image, metaData));
+    }
+}
+
+/**
+ * Publus
+ */
+const NFBR = {
```
27136773, line 187: Not reviewed for correctness since this is extracted from website itself

27136773 (commented):
Looks good, may need to check if DataURL is limited

src/web/mjs/connectors/LineMangaJP.mjs
Elastic1: It seems "Image" object has no limit for dataURL. However, there is a ~2MB limit for iframe tags or opening in a new tab.
(tested [here](https://codepen.io/Elastic1/pen/zYWGKrM))

27136773 (approved):
Ready for merge

Issue #4483, opened 2022-03-13
https://manhwa18.com/


Dan20088:
site changed and not working anymore

MikeZeDev:
Preview and download works fine on nightly. @Sheepux ?

Issue #4320, opened 2022-01-21
[manhwa18.com] Connector not working


xlunarwastakenx:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes

**Describe the bug**
When I go to the connector and click the green button to reload, no mangas popup. Ex:
![image](https://user-images.githubusercontent.com/98142191/150468650-f42b4b4a-b136-42fd-b32f-29df0deaccd3.png)
I have repeatedly clicked the button but no manga show up. 

**To Reproduce**
1.  manhwa 18: the site has visible chapters, but someone reason they don't load.

![image](https://user-images.githubusercontent.com/98142191/150468546-fdf56730-6f23-4865-ae53-966ae34fe570.png)
Thanks so much for you help

MikeZeDev:
Preview and download works fine on nightly. @Sheepux ?

Sheepux:
can confirm

Issue #4279, opened 2022-01-05
https://manhwa18.com/


jimiz88:
https://manhwa18.com/

site got updated it's not working anymore

MikeZeDev:
Preview and download works fine on nightly. @Sheepux ?

Issue #4151, opened 2021-11-14
[manhwa18] Connector not working


zjhl214:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes

**Describe the bug**
Chapter lists all display 0/0

**To Reproduce**
1. manhwa18.com
2. Any
3. Any
4. _getUpdatedChaptersFromWebsite Error: Failed to receive chapter list (status: 404) - at Manhwa18._getChapters (FlatManga.mjs:49) at aysnc Manhwa18._getChapterList (Connector.mjs:595)

**HakuNeko (please complete the following information):**
 - 6.17

27136773:
No longer `FlatManga` template, seems to be completely new layout.

MikeZeDev:
Everything works. Nightly. Can be closed?

MikeZeDev:
Cant repro.

PR #4815, opened 2022-07-09
Manhwa18com connector

jmvrgit:
Website: Manhwa18.com 
Issue: had an update that broke the connector

Resolves the following open issues:
https://github.com/manga-download/hakuneko/issues/4483
https://github.com/manga-download/hakuneko/issues/4320
https://github.com/manga-download/hakuneko/issues/4279
https://github.com/manga-download/hakuneko/issues/4151

Before:
Even after synch, manga list is empty.
image

After:
Everything working as expected.
image


27136773 (approved):
Looks good

Issue #4816, opened 2022-07-09
[manhwa18-int] Connector not working


jmvrgit:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Is the website of the connector working properly / are you able to see the manga within your browser**
No

**Describe the bug**
Website update broke the connector.

**To Reproduce**
1. manhwa18-int / ManhwaEighteen / Manhwa18 (.net)
2. After Selecting Connector, Click on Synch
3. Any
4. N/A

**Screenshots**
![image](https://user-images.githubusercontent.com/9809253/178102425-55c636e5-098f-4c74-9180-f12a21c7276a.png)

**HakuNeko (please complete the following information):**
 - Latest Master

jmvrgit:
Issue Resolved

PR #4817, opened 2022-07-09
Manhwa18net connector

jmvrgit:
https://github.com/manga-download/hakuneko/issues/4816

Solves issue for Manhwa18.net connector not working due to website update.

Be advised that there are pages / manhwa that they are hosting that seems to be broken even during normal browsing. I thought this was due to a header issue prior, but its not. Their site is just broken.
Before:
image
After:
image

### DIFF
--- a/src/web/mjs/connectors/ManhwaEighteen.mjs
+++ b/src/web/mjs/connectors/ManhwaEighteen.mjs
@@ -45,7 +45,7 @@ export default class ManhwaEighteen extends Connector {
         return data.map(element => {
             return {
                 id: this.getRootRelativeOrAbsoluteLink(element, request.url),
-                title: element.text.trim(),
+                title: element.querySelector('div.chapter-name').textContent.trim(),
                 language: ''
             };
         });
--- a/src/web/mjs/connectors/ManhwaEighteen.mjs
+++ b/src/web/mjs/connectors/ManhwaEighteen.mjs
@@ -1,4 +1,5 @@
 import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
 
 export default class ManhwaEighteen extends Connector {
 
@@ -51,5 +52,13 @@ export default class ManhwaEighteen extends Connector {
         let request = new Request(this.url + chapter.id, this.requestOptions);
         let data = await this.fetchDOM(request, 'div.chapter-content source._lazy');
         return data.map(element => this.getAbsolutePath(element.dataset.src || element, request.url));
+    }
+
+    async _getMangaFromURI(uri) {
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, 'nav ol li.active');
+        let id = uri.pathname;
+        let title = (data[0].content || data[0].textContent).trim();
+        return new Manga(this, id, title);
     }
 }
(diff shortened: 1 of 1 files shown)


jmvrgit:
Do not merge yet, we are trying to solve Response 520

jmvrgit:
Solved this issue on latest commit:
Once a manga list is generated via synch, it throws a response 520. Resynching the manga list results in a blank manga list.
Also solved that it takes longer than necessary to generate the mangalist.

jmvrgit:
result:
image

Druzai:
You forgot add `_getMangaFromURI` method. It can be implemented like this:
```js

...

    async _getMangaFromURI(uri) {
        let data = await this.fetchDOM(request, 'nav ol li.active');
        let id = uri.pathname;
        let title = (data[0].content || data[0].textContent).trim();
        return new Manga(this, id, title);
    }
```
Without it the manga name is incorrectly retrieved from the link in clipboard.

src/web/mjs/connectors/ManhwaEighteen.mjs
```diff
@@ -1,15 +1,59 @@
...
+    }
+
+    async _getChapters(manga) {
+
+        let request = new Request(this.url + manga.id, this.requestOptions);
+        let data = await this.fetchDOM(request, 'ul.list-chapters a');
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, request.url),
+                title: element.text.trim(),
```
Druzai: To remove the upload date of chapter, I think it's better to get `div.chapter-name`.
```suggestion
                title: element.querySelector('div.chapter-name').textContent.trim(),
```

src/web/mjs/connectors/ManhwaEighteen.mjs
```diff
@@ -1,15 +1,59 @@
...
+            if (JSON.stringify(mangaList[mangaList.length - 1]) == undefined && mangas[mangas.length - 1] != undefined ) {
+                mangaList.push(...mangas);
+            } else if (JSON.stringify(mangaList[mangaList.length - 1].id) != JSON.stringify(mangas[mangas.length - 1].id)) {
+                if (mangas.length > 0) {
+                    mangaList.push(...mangas);
+                }
+            } else {
+                run = false;
+            }
+        }
```
Druzai: To remove that 502 Error, I think it's better to get total count of pages, because the site can return the same last page with increasing page parameter in the request.
```suggestion
        const uri = new URL('/manga-list.html?listType=pagination&sort=name&sort_type=ASC', this.url);
        const request = new Request(uri, this.requestOptions);
        const data = await this.fetchDOM(request, 'div.pagination-wrap ul.pagination li:nth-last-child(2) a');
        const pageCount = parseInt(data[0].text.trim());
        for (let page = 1; page <= pageCount; page++) {
            const mangas = await this._getMangasFromPage(page);
            mangaList.push(...mangas);
        }
```

jmvrgit: yours is a much cleaner solution. what i did was to compare the contents of the last push to the new ones to be pushed since anything beyond the last page is repeated with each request until like ~160 requests until it throws the 520 error.

jmvrgit:
I've implemented the changes. Thanks for the detailed explanation

27136773 (approved):
Looks good

PR #4841, opened 2022-07-18
manga.ovh add

knst0:
I developed Connector for manga.ovh based on it REST API.


MWGuy:
🤣

src/web/mjs/connectors/MangaOvh.mjs
```diff
@@ -0,0 +1,96 @@
...
+        super.label = 'MangaOvh';
+        this.tags = [ 'manga', 'webtoon', 'russian' ];
+        this.url = 'https://manga.ovh';
+        this.api = 'https://api.manga.ovh';
+    }
+
+    async _getMangasFromPage(page) {
+        const requestUrl = new URL('/book', this.api);
+        requestUrl.searchParams.set('type', 'COMIC');
+        requestUrl.searchParams.set('page', String(page));
```
27136773: MAy remove redundant conversion `String(page)`, `searchParams` are always processed as strings

src/web/mjs/connectors/MangaOvh.mjs
```diff
@@ -0,0 +1,96 @@
...
+        const requestUrl = new URL('/book/' + slug, this.api);
+        requestUrl.searchParams.set('type', 'COMIC');
+
+        const request = new Request(requestUrl, this.requestOptions);
+        const data = await this.fetchJSON(request, 3);
+
+        return new Manga(this, data.id, data.name.en || Object.values(data.name).shift());
+    }
+
+    async _getBranches(manga) {
```
27136773: Method only has 2 lines of code and is only used in a single place, code may be inlined instead of extracted in separate method.
Some other methods as well...

src/web/mjs/connectors/MangaOvh.mjs
```diff
@@ -0,0 +1,96 @@
...
+        const request = new Request(new URL('/book/' + manga.id + '/branches', this.api), this.requestOptions);
+        return this.fetchJSON(request);
+    }
+
+    async _getChaptersFromManga(manga) {
+        const branches = await this._getBranches(manga);
+        let chapters = [];
+
+        for (let i = 0; i < branches.length; i++) {
```
27136773: May be simplified `for (const branch of branches)` to avoid using intermediate index variable

src/web/mjs/connectors/MangaOvh.mjs
```diff
@@ -0,0 +1,96 @@
...
+            const tmp = await this._getChaptersFromBranch(branches[i].id);
+            tmp.map(chapter => chapter.branchName = branches[i].translators.map(translator => {
```
27136773: May use `forEach`, `map` implies other meaning and therefore is a bad choice for the purpose iterating

27136773 (approved):
Functionality looks good, code has some room for optional improvements

knst0:
Yesterday they released a fix against parsers, so the whole code has to be redone...

27136773:
> Yesterday they released a fix against parsers, so the whole code has to be redone...

😢 May abandon this PR and create a new one when done

PR #4851, opened 2022-07-22
Fix Cmoa

Elastic1:
Support for downloading purchased item


src/web/mjs/connectors/Cmoa.mjs
```diff
@@ -35,14 +35,19 @@ export default class Cmoa extends SpeedBinb {
             const pageRequest = new Request(uri, this.requestOptions);
             const data = await this.fetchDOM(pageRequest, '.title_vol_vox_vols .title_vol_vox_vols_i');
             for (const element of data) {
-                const cartButton = element.querySelector('.cart_into_btn');
-                if (cartButton == null) {
+                const anchorElements = Array.from(element.querySelectorAll('a'));
+                const chapterPath = anchorElements.find(anchorElement => {
```
27136773: May be simplified to something like `const chapterPath  = element.querySelector('a[href^="/reader/"]')?.pathname;`

27136773 (approved):

PR #4882, opened 2022-08-03
cubari.moe support

knst0:
Supporting cubari.moe site


27136773:
Cubari is a reader midddleware that doesn't host content on its own website, it just proxies to other websites such as Imgur, MangaDex, MangaSee, ...
All these websites are already supported in HakuNeko.

Therefore I suggest to not add Cubari itself as a separate website.

27136773:
In case you want to keep it for personal use, you may consider to make it as a user plugin:
To do this, [update the imports and place it in the plugin directory](https://hakuneko.download/docs/install/#user-plugins)

knst0:
Thanks, I put connector to separate plugin.

PR #4891, opened 2022-08-11
[Mangapark] Fix getting manga title from clipboard

Druzai:
Mangapark added endpoints with new front-end - `React`, so query selector has to be different for old and new front-end.

### DIFF
--- a/src/web/mjs/connectors/MangaParkEN.mjs
+++ b/src/web/mjs/connectors/MangaParkEN.mjs
@@ -35,13 +35,12 @@ export default class MangaParkEN extends Connector {
     async _getMangasFromPage(page) {
         const uri = new URL('/browse?sort=name&page=' + page, this.url);
         const request = new Request(uri, this.requestOptions);
-        const data = await this.fetchDOM(request, '#subject-list div.item');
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -15,8 +15,11 @@ export default class MangaParkEN extends Connector {
 
     async _getMangaFromURI(uri) {
         const request = new Request(uri, this.requestOptions);
-        const data = await this.fetchDOM(request, 'meta[property="og:title"]');
-        return new Manga(this, uri.pathname.match(/\/(\d+)\/?/)[1], data[0].content.trim());
+        let queryMangaTitleFromURI = 'meta[property="og:title"]';
```
27136773: May use ternary operator
`const queryMangaTitle = /\/title\/\d+/.test(uri.pathname) ? 'main h3' : meta[property="og:title"]';`

27136773 (approved):

src/web/mjs/connectors/MangaParkEN.mjs
Druzai: Fixed

27136773 (approved):

PR #4916, opened 2022-08-24
Futekiya

### LINKED ISSUE #4862: [Site Request] Futekiya (MangaPlanet)
**Name of the website**
Futekiya (MangaPlanet)

- Site: https://read.futekiya.com/
- Manga List: https://read.futekiya.com/browse
- Manga example: https://read.futekiya.com/comic/60d5288c53477
- Chapter online viewer example: https://image.futekiya.com/viewer/60d5289f215bd

**Languages**
English

**Website relationship*
https://read.mangaplanet.com/

**Additional details**
I believe futekiya is under MangaPlanet? MP has decent library. Anyway, I would like to download the content I subscribed for to read offline, but...
Apparently it uses SpeedBinb reader version '1.6600.2001'?
MP has notable titles like AOT, Tokyo Revengers, Your Lie in April, Welcome to the Ballroom... futekiya is um... ahem... a lot of BL, lmao...
idk if this is any help lol: https://gist.github.com/SingularReza/63a3b2a1fee231676260d657bc29b6cb

gammesB:
the _getPageList had to be slichtly altered because without the right referer you will get redirected to the home page/get a 404 page.

added fix for the tocoronaex test whilee I was at it.

example: https://read.futekiya.com/comic/6296b6745ae85

resolves: #4862


src/web/mjs/connectors/Futekiya.mjs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    async _getMangas() {
+        let maxpagecount = 1;
+        let manga = [];
+        for (let i = 1; i <= maxpagecount; i++) {
+            const request = new Request(new URL('/browse?page=' + i, this.url), this.requestOptions);
+            if (maxpagecount == 1) {
+                const newpagecount = await this.fetchDOM(request, ".pagination :nth-last-child(2)");
+                maxpagecount = parseInt(newpagecount[0].innerText.trim());
```
27136773: Can this be done before starting the loop?
This would avoid the conditional check.

src/web/mjs/connectors/Futekiya.mjs
```diff
@@ -0,0 +1,73 @@
...
+        let request = new Request(uri, this.requestOptions);
+        let data = await this.fetchDOM(request, '.card-body.book-detail h3');
+        let id = uri.pathname;
+        let title = data[0].innerText.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let maxpagecount = 1;
+        let manga = [];
```
27136773: Consider using plural naming for array variables.

src/web/mjs/connectors/Futekiya.mjs
```diff
@@ -0,0 +1,73 @@
...
+                    id: url,
+                    title: title + chapter.querySelector('span').innerText.trim()
+                });
+            }
+        }
+
+        return mangadownloads;
+    }
+    _getPageList(manga, chapter, callback) {
+        this.requestOptions.headers.set('x-referer', "https://read.futekiya.com/");
```
27136773: may use `this.url + '/'`

src/web/mjs/connectors/Futekiya.mjs
```diff
@@ -0,0 +1,73 @@
...
+        let data = super._getPageList(manga, chapter, callback);
```
27136773, line 74: The state of `this.requestOptions.headers` may become something unexpected, e.g. when an exception occurs
=> use local variables whenever possible instead of relying on modified fields (in a critical application this would also introduce concurrency problems)

27136773 (approved):

src/web/mjs/connectors/Futekiya.mjs
gammesB: sorry I don't get the problem with this one.
I checked the code on how other connectors do it but they all use this to set the referer.
what local method could I use to set the referer? is this not the normal method for changing requestOptions?
the super uses new Request(uri, this.requestOptions); just like most requests so unless I want to copy the whole _getPageList and downloadfunction I don't see how i can make a local requestOptions

src/web/mjs/connectors/Futekiya.mjs
gammesB: good catch, my intention first was to only use fetchDOM once but as I did use two in the end adding it to the for loop doesn't help anyone.

src/web/mjs/connectors/Futekiya.mjs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    async _getMangas() {
+        let maxPageCount = 1;
+        let mangas = [];
+        const request = new Request(new URL('/browse?page=1', this.url), this.requestOptions);
+        const newpagecount = await this.fetchDOM(request, ".pagination :nth-last-child(2)");
+        maxPageCount = parseInt(newpagecount[0].innerText.trim());
+        for (let i = 1; i <= maxPageCount; i++) {
+            const request = new Request(new URL('/browse?page=' + i, this.url), this.requestOptions);
```
Druzai: Um, this will cause redeclaration error in runtime. Did you forget to define `request` as `let`?
```suggestion
        let request = new Request(new URL('/browse?page=1', this.url), this.requestOptions);
        const newpagecount = await this.fetchDOM(request, ".pagination :nth-last-child(2)");
        maxPageCount = parseInt(newpagecount[0].innerText.trim());
        for (let i = 1; i <= maxPageCount; i++) {
            request = new Request(new URL('/browse?page=' + i, this.url), this.requestOptions);
```

gammesB: I think it will still work and not cause erors because the const is in a forloop making it on a different level. but it's indeed still a good change.

src/web/mjs/connectors/Futekiya.mjs
27136773: The idea is that `this.requestOptions` should be handled as readonly (only updated once in constructor).
If a method needs different `this.requestOptions` it should create a clone and make the changes on the clone, so it won't affect other methods that relies on `this.requestOptions` in its initial state.

27136773 (approved):

PR #4967, opened 2022-09-13
add connector

akn19:
https://www.hinapyon.fun/


src/web/mjs/connectors/DoujinDesu.mjs
```diff
@@ -7,21 +7,21 @@ export default class DoujinDesu extends WordPressMangastream {
...
         this.queryChapters = 'div#chapter_list div.epsleft span.lchx a';
+        this.queryPages = 'div.main div img[src]:not([src=""])';
         this.queryChaptersTitle = undefined;
-        this.queryPages = 'div.reader-area img[src]:not([src=""])';
+        this.querMangaTitleFromURI = 'section.metadata h1.title';
     }
 
-    // NOTE: Fallback to pagination since '/komik-list/?list' empty
+    // NOTE: Fallback to pagination since '/manga/?list' empty
     async _getMangas() {
```
27136773, line 25: Can this method be removed?
Looks to me like it would work fine with `/manga/page/1/'`, so probably no workaround required.

27136773:
The icon for `hinapyon.fun` is to large.
It should have max. size of 16KB, may try to reduce resolution (e.g. 64x64 px) or use other format with better compression (e.g. jpg)

27136773 (changes requested):
Looks good, but the icon should be smaller

PR #4998, opened 2022-09-22
allows downloding subscription chaps from corona.ex and fixed bug with downloading chapters.

gammesB:
- fixed bug with descramble accidentally removing 5 pixels of the right of the image(as that part isn't part ofthe scramble).
- added support to download subscription only chaps instead of only the free chaps.

### DIFF
--- a/src/web/mjs/connectors/To-Corona-Ex.mjs
+++ b/src/web/mjs/connectors/To-Corona-Ex.mjs
@@ -23,11 +23,11 @@ export default class ToCoronaEx extends Connector {
     async _getMangas() {
         let mangaList = [];
         let nextCursor = '/comics?order=asc&sort=title_yomigana';
-        for(let run = true; run;) {
+        for (let run = true; run;) {
             const data = await this._getMangasFromPage(nextCursor);
             nextCursor = data.nextCursor;
             mangaList.push(...data.mangas);
-            if(nextCursor == null) {
+            if (nextCursor == null) {
                 run = false;
             }
         }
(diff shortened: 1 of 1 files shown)


sboudouk (approved):

src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -51,30 +51,89 @@ export default class ToCoronaEx extends Connector {
 
     async _getChapters(manga) {
         let chapterList = [];
+        const getToken = await this.getToken();
```
27136773, line 54: The variable has the same name as the member function `getToken`, i recommend to use a more appropriate name

src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -51,30 +51,89 @@ export default class ToCoronaEx extends Connector {
...
-            if(nextCursor == null) {
+            if (nextCursor == null) {
                 run = false;
             }
         }
         return chapterList.reverse();
     }
+    //get auth token from indexedDB
+    async getToken() {
+        let script = `new Promise(resolve => {
```
27136773, line 85: The cyclomatic complexity is quite high for this script, may try to reduce nested levels
e.g.

```javascript

// nested callbacks
setTimeout(() => {
    const request1 = new Request(0);
    request1.onerror = (err) => { /* Handle Error */ };
    request1.onsuccess = (data1) => {
        const request2 = new Request(data1.foo);
        request2.onerror = (err) => { /* Handle Error */ };
        request2.onsuccess = (data2) => { /* Do something with data2 ... */ };
    };
}, 2500);

// flatten control flow with promises
setTimeout(() => {
    const data1 = await new Promise((resolve, reject) => {
        const request = new Request(0);
        request.onerror = (err) => reject(err);
        request.onsuccess = (data) => resolve(data);
    });
    const data2 = await new Promise((resolve, reject) => {
        const request = new Request(data1.foo);
        request.onerror = (err) => reject(err);
        request.onsuccess = (data) => resolve(data);
    });
}, 2500);
```

src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -51,30 +51,89 @@ export default class ToCoronaEx extends Connector {
...
+            setTimeout(() => {
+                try{
+                    let tRequest = window.indexedDB.open("firebaseLocalStorageDb", 1);
+                    tRequest.onerror = function (event) {
+                        resolve(undefined);
```
27136773, line 90: Consider to reject the promise

src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -51,30 +51,89 @@ export default class ToCoronaEx extends Connector {
...
 
-    async _getChaptersFromPage(manga, nextCursor) {
+                        const db = e.target.result;
+                        const transaction = db.transaction("firebaseLocalStorage", 'readonly');
+                        const objectStore = transaction.objectStore("firebaseLocalStorage");
+
+                        if ('getAll' in objectStore) {
+                            objectStore.getAll().onsuccess = function (event) {
+                                if(event.target.result.length == 0){
+                                    resolve(undefined);
```
27136773, line 101: Consider to reject the promise

src/web/mjs/connectors/To-Corona-Ex.mjs
```diff
@@ -51,30 +51,89 @@ export default class ToCoronaEx extends Connector {
...
+                                }
+                                else{resolve(event.target.result[0].value.stsTokenManager.accessToken);
+                                    }
+                            };
+                            objectStore.getAll().onerror = function (event) {
+                                resolve(undefined);
```
27136773, line 107: Consider to reject the promise

27136773 (approved):
Looks good, some room for possible improvements

PR #5129, opened 2022-11-07
Update MangaShip.mjs

Furkangcmn:
i updated the domain name


Sheepux:
duplicate, merged #5130

Barxaomanga:
the site works but does not show sections in the program

PR #5133, opened 2022-11-09
Fix getting pictures in Mangatown

### LINKED ISSUE #4726: [Bug] Can't Enjoy the Downloaded File
Yes

**Describe the bug**
I cannot seem to download the picture.. well, I can download it, but the file is just broken. When I downloaded a certain chapter (i.e.: Black Clover 331), the output only showed files with **_.bin_** extension instead of **_.jpg_**, **_.jpeg_**, or **_.png_**; the issue occurs on MangaTown & MangaHere, I dunno if the same case exists elsewhere.

**To Reproduce**
1. Select a website (MangaTown or MangaHere)
2. Select any manga & any chapter to download
3. Open your folder of downloaded chapter
4. Witness the majestic bugs 👀

**Expected behavior**
an outcome of solid pictures just like what the usual HakuNeko offers ):

**Screenshots**
HaquNeqo

 - OS: Windows 11 Pro
 - 6.17

**Additional context**
I'm from **Indonesia** & I can access the site just fine, even without a VPN.
I tried downloading 1 random chapter from **Kissmanga & it works just fine**.

MikeZeDev:
Add mangahere referer to picture requests. Fixes https://github.com/manga-download/hakuneko/issues/4726


27136773:
Check and fix `lint` errors.

27136773 (approved):

src/web/mjs/connectors/MangaTown.mjs
```diff
@@ -66,6 +66,7 @@ export default class MangaTown extends Connector {
     }
 
     _handleConnectorURI( payload ) {
+        this.requestOptions.headers.set("x-referer", "mangahere.com");
```
27136773: This will add `MangaHere` as referer to all requests for this connector (manga lists, chapter lists, ...), is this intentional?
If not create a clone of request options and add the header to the clone.

Sheepux: personally i'd rather reject the request and alert("use mangahere directly")

MikeZeDev: Hum, you are right, better not meddle with the entire request options.

MikeZeDev: In fact setting the referer to mangahere is only needed for downloading pictures, which is in fine handled by  Connector.handleconnectorURI. And it uses "this.requestoptions" without referer. No matter what i do in mangatown own handleconnectorURI, its useless .

Is there any reason we are using handleconnectorURI in this connector? How could i set a referer but only for picture requests? I can only think about dirty workarounds.

27136773: Just set the header on the cloned request...
```javascript
let request = new Request(payload, this.requestOptions);
request.headers.set('x-referer', 'https://mangahere.com');
```

MikeZeDev: This request get the page with the pics, and we dont need mangahere ref for that. :)

Downloading pictures need mangahere ref.

Which is done not in Mangatown.mjs but in parent class, where i cant pass a custom referrer .
Wait, can we pass a custom referer in payloads?

MikeZeDev: I managed to fix the code so only picture requests get to use mangahere referer.

src/web/mjs/connectors/MangaTown.mjs
```diff
@@ -56,23 +51,32 @@ export default class MangaTown extends Connector {
...
+        let imgpages = data
             .filter(option => !option.value.endsWith('featured.html'))
-            .map(element => this.createConnectorURI(this.getAbsolutePath(element.value, request.url)));
+            .map(element => this.getAbsolutePath(element.value, request.url));
+        //NOW we have all the pages, get the picture url from each page
+        //and create the payloads with mangahere referer
+        let imglist = [];
+        for (let i = 0; i<imgpages.length; i++ ) {
+            let request = new Request(imgpages[i], this.requestOptions);
+            data = await this.fetchDOM( request, 'source#image' );
```
27136773: I strongly recommend to move the content of this loop as part into `_handleConnectorURI`.
The `_getPages` method is supposed to be fast and should not make too many requests if possible.
e.g.,

```javascript
    const pageData = await this.fetchDOM(new Request(payload, this.requestOptions), 'source#image');
    const imageURL = this.getAbsolutePath(pageData[0].src, request.url).replace('hakuneko://', 'https://');
    const request = new Request(imageURL, this.requestOptions);
    request.headers.set('x-referer', 'mangahere.com');
    const imageData = await response.blob();
    return this._blobToBuffer(imageData);
}
```

MikeZeDev: Done

PR #5146, opened 2022-11-12
Added ynjn (ヤンジャン!)

Elastic1:
#5047


src/web/mjs/connectors/Ynjn.mjs
```diff
@@ -0,0 +1,51 @@
...
+    async _getMangas() {
+        let msg = 'This website does not provide a manga list, please copy and paste the URL containing the chapters directly from your browser into HakuNeko.';
+        throw new Error(msg);
+    }
+
+    async _getChapters(manga) {
+        const chapters = [];
+        let hasNext = true;
+        let page = 1;
+        while (hasNext) {
```
27136773: May use a `for` loop
```javascript
for(let page = 1, let hasNext = true; hasNext; page++)
{
    // ...
}
```

27136773 (approved):

Issue #4644, opened 2022-05-04
[Manga Livre] Can't download mangas from "Manga Livre" gives "TypeError: Cannot read property 'src' of null"


LaBatata101:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Describe the bug**
When trying to download or show the preview any manga from "Manga Livre" this error appears `TypeError: Cannot read property 'src' of null`

**To Reproduce**
1. Go to Connectors
2. Search for "Manga Livre"
3. Synchronize the manga list
4. Click on any manga from the manga list
5. Click on download or show preview button

**Expected behavior**
The manga should be downloaded or previewed without any errors showing up

**Screenshots**
Trying to download
![Screenshot from 2022-05-04 19-47-11](https://user-images.githubusercontent.com/20308796/166837719-904e14c3-a277-4e85-b6b3-f668e8e648b2.png)

**Context (please complete the following information):**
 - OS: Manjaro Linux 
 - HakuNeko Version 6.1.7@cf569b

**Additional context**

adamesalles:
Wow, we have the very same issue, with the very same configs. (Manjaro, MangaLivre, etc.)

Did you figure out how to solve this?

I'm thinking of downloading HakuNeko from other source than AUR.

LaBatata101:
> Wow, we have the very same issue, with the very same configs. (Manjaro, MangaLivre, etc.)
> 
> Did you figure out how to solve this?
> 
> I'm thinking of downloading HakuNeko from other source than AUR.

I downloaded from a different website instead of "Manga Livre". Downloading HakuNeko from a different source other than AUR ins't going to fix the issue, because the problem is likely due to "Manga Livre" changing its HTML source code and HakuNeko ins't able to scrap it.

adamesalles:
Which one are you using? I think I could use YesMangasBR, but not really sure once it's not working now.

LaBatata101:
I don't remember anymore, which one I used.

MikeZeDev:
Geoblocked

MikeZeDev:
Duplicate https://github.com/manga-download/hakuneko/issues/4061

PR #5195, opened 2022-12-01
Manga livre don't work's

thiagovianav:
Hi, thanks for the good work. I'm from Brazil, but in Brazil, mangaLivre is the most used manga services. But in hakuneko this dont's works for downloads, just error. :( (url: https://mangalivre.net)


Sheepux:
Open an issue not a merge request. You are not coding .

Also it's already been reported https://github.com/manga-download/hakuneko/issues/4644

PR #5216, opened 2022-12-09
Add videosteam hoster : Dood.wf

MikeZeDev:
* This is enough to download the video using Hakuneko. 
* Preview doesnt work because connection is refused, i guess its a limitation of electron. Curl got the same problem.
* Kinda dirty method but it works.

### DIFF
--- a/src/web/mjs/videostreams/Dood.mjs
+++ b/src/web/mjs/videostreams/Dood.mjs
@@ -13,16 +13,16 @@ export default class Dood {
         let request = new Request(this._uri);
         const data = await Engine.Request.fetchUI(request, script, 30000);
         //end of final video url
-        const videourlend = data.tok;
+        const videoUrlEnd = data.tok;
         //get the url for requestion the video url beginning
         let dom = (new DOMParser).parseFromString(data.doc, "text/html");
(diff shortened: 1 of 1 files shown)


src/web/mjs/videostreams/Dood.mjs
```diff
@@ -0,0 +1,28 @@
...
+        new Promise(resolve => {
+            resolve({
+                tok : makePlay(), doc : document.documentElement.innerHTML
+            });
+        });
+        `;
+        let request = new Request(this._uri);
+        const data = await Engine.Request.fetchUI(request, script, 30000);
+        //end of final video url
+        const videourlend = data.tok;
```
27136773: Use camelCase for local variable names (other variables as well)

27136773 (commented):
It seems this video provider is not integrated in any of the existing websites, is this by intention or was this forgotten to be added to an existing website?

If this video provider is unused, it should not be added.

MikeZeDev:
Indeed its ready to be included in future connectors i already coded. Idk if its worth, as previewing doesnt work for strange reasons (connection refused, not even http error). That said, downloading works.

27136773 (approved):

PR #5234, opened 2022-12-14
Update NetTruyen domain

Khangry:


MikeZeDev (approved):

MikeZeDev:
I see the url changed but it doesnt work anymore even after url change :/

PR #5236, opened 2022-12-14
ADD - AnimeParadise.moe

MikeZeDev:
Fully working anime website, with subtitles preview and download <3


Sheepux (approved):
LGTM (Ronny is too busy to review)
however i've never worked on an anime connector and therefore cannot asses if everything is correct

PR #5242, opened 2022-12-16
Update Nettruyen domain

Khangry:


MikeZeDev:
First domain still works.  I dont see the need to change.

Sheepux:
Current domain doesn't redirect, how do you know wich one is correct ?

Khangry:
> First domain still works. I dont see the need to change.

it works when you live abroad but in Vietnam it is blocked

Khangry:
> Current domain doesn't redirect, how do you know wich one is correct ?

I don't know, but sometimes the current domain work but sometimes it isn't. However, providers in Vietnam are very capricious :(

MikeZeDev:
If you watch carefully, their manga list is not the same. One more reason not to change. 
image
image

Same page on both website, different content. Seems a clone to me, like we can see for many websites.

Given that this website just use a template without modifications its prob someone who download previous website content and made his own.

Sheepux:
could be added as a connector setting (domain change)

PR #5244, opened 2022-12-17
Update Nhattruyen domain

Khangry:


MikeZeDev:
http://nhattruyenvip.com redirect to http://nhattruyentop.com for me. 

Since the redirection is automatic we can assume  nhattruyentop is the new one and not the one you used. I see no pb approving it if redirection is done, so change the url again pls.

That said i dont know whats going on with viet websites but seing PR about address changes every two days is annoying.
Lets hope they calm down a bit.

Sheepux (changes requested):
same things as mike, the redirect points to https://nhattruyentop.com/
nhattruyentop.com Created on 2022-12-06
nhattruyenmin.com Created on 2022-11-23
I'd say that nhattruyentop.com is the correct domain

If the domain still changes so often, using domain settings in HN will be required

Khangry:
I don't know but with different providers they block different domains, with provider A they block nhattruyentop with provider B they block nhattruyenme. My opinion is to support multiple domains at the same time as an alternative.

MikeZeDev:
For now we can merge, we will put a setting after it if changes too much i guess.

Sheepux (approved):

PR #5246, opened 2022-12-17
Flatmanga : better chapter title handling

MikeZeDev:
If there is a special regex character in the title (like +, [ ] () ) , an exception was caused and no chapter was returned. Better having a long a* title than no chapter at all.

Exemple : KLMANGA > + Tic Nee Sans , [HAISIN CHUU]

### DIFF
--- a/src/web/mjs/connectors/templates/FlatManga.mjs
+++ b/src/web/mjs/connectors/templates/FlatManga.mjs
@@ -58,11 +58,7 @@ export default class FlatManga extends Connector {
             }
             let title = (this.queryChapterTitle ? element.querySelector(this.queryChapterTitle) : element).textContent.replace(manga.title, '');
             let mangaTitle = manga.title.replace(/\s*-\s*RAW$/, '');
-            try {
-                title = title.replace(new RegExp(mangaTitle, 'i'), '');
-            } catch (error) {
-
-            }
+            title = title.replace(new RegExp(mangaTitle.replace(/[#-}]/g, '\\$&')), '');
             title = title.replace(/^\s*-\s*/, '');
             title = title.replace(/-\s*-\s*Read\s*Online\s*$/, '');
             title = title.trim();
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/templates/FlatManga.mjs
```diff
@@ -58,7 +58,11 @@ export default class FlatManga extends Connector {
             }
             let title = (this.queryChapterTitle ? element.querySelector(this.queryChapterTitle) : element).textContent.replace(manga.title, '');
             let mangaTitle = manga.title.replace(/\s*-\s*RAW$/, '');
-            title = title.replace(new RegExp(mangaTitle, 'i'), '');
```
Sheepux, line 61: how about `new RegExp(myregex.replace(/[#-}]/g, '\\$&'))` to sanitize the string ?

Sheepux: did you test it well first ? I'm not sure if it covers all use cases

MikeZeDev: Hum, if i force the manga title to be in chapter name (for testing), and in use the regex fixed, nothing is removed :/
```
            title += ' '+ mangaTitle;
            title = title.replace(new RegExp(mangaTitle.replace(/[#-}]/g, '\\$&')), '');
```
image

Gotta dig a bit

MikeZeDev: OK got a better regex using this as reference https://www.debuggex.com/cheatsheet/regex/javascript

`            title = title.replace(new RegExp(mangaTitle.replace(/[\*\^\$\.\|\?\+\-\(\)\[\]\{\}]/g, '\\$&')), '');`

Seems to work fine

MikeZeDev: Using this as reference : https://exlskills.com/learn-en/courses/javascript-fundamentals-basics_javascript/regular-expressions-cWSVBpppzSat/applications-lPjEnTmnZThz/special-characters-in-regex-xMIVQGkIObNm

`[\*\^\$\.\|\?\+\-\(\)\[\]\{\}]` seems to work fine. We escape only special chars and it works.

MikeZeDev:
image

Works fine.

Sheepux (approved):

PR #5256, opened 2022-12-20
FIX - Readerfront websites fixes

### LINKED ISSUE #4976: [RavensScans] Connector not working
English or spanish, nothing works. Manga list, preview, download. Api changed?

MikeZeDev:
* Fixes https://github.com/manga-download/hakuneko/issues/4976
* Rewrite ScyllaScans so it use ReaderFront template

### DIFF
--- a/src/web/mjs/connectors/RavensScansEN.mjs
+++ b/src/web/mjs/connectors/RavensScansEN.mjs
@@ -11,6 +11,7 @@ export default class RavensScansEN extends ReaderFront {
         this.cdn = 'https://img-cdn1.ravens-scans.com';
         this.apiURL = 'https://api.ravens-scans.com';
         this.language = 'en';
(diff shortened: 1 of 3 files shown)


MikeZeDev:
Just a note : template is wrong.
Graphql underneath expect 'languages" to be an ARRAY of integer, one integer for each language you want ([1,2] for english and spanish), or an  empty array for ALL LANGUAGES. If this.language = '' then the array will be [0]  and the request to get manga list will return no mangas.

Thats easy to fix BUT then, chapter request expect a single language, not an array. We could use the manga language if only the info was retained but its not.

I guess its the reason why there is Raven scans (english) and  Raven scans (Spanish)

MikeZeDev:
Tbh since ReaderFront template is old school code i wonder if i should use ScyllaScans old code as base to rewrite it :D

Sheepux (approved):

PR #5259, opened 2022-12-21
Ichijinplus & to-Corona-ex : Add needed API KEY

MikeZeDev:
WIthout api key we got 403 errors

Fixes https://github.com/manga-download/hakuneko/pull/5259


Sheepux:
is there a way to dynamicaly fetch the api key on connector init ?

MikeZeDev:
Key seems to be fixed, or i am blind. If its not, i have to dig but its not in storage or cookies

MikeZeDev:
Key is hardcoded in app.js
image

I am afraid no, you cant easily fetch it.

Sheepux (approved):

PR #5273, opened 2022-12-24
FIX - Alphapolis Get high resolution pictures

MikeZeDev:


src/web/mjs/connectors/Alphapolis.mjs
```diff
@@ -47,7 +47,7 @@ export default class Alphapolis extends Connector {
         const data = await this.fetchDOM(request, 'viewer-manga-horizontal');
         try {
             const pages = JSON.parse(data[0].getAttribute('v-bind:pages'));
-            return pages.filter(element => typeof element != 'object' && !element.match('white_page'));
+            return pages.filter(element => typeof element != 'object' && !element.match('white_page')).map(element => element.replace(/\/[0-9]+x[0-9]+.([\w]+)/, '/1080x1536.$1'));
```
Sheepux, line 50: how was this value determined ? i don't see how all mangas would have the same resolution

MikeZeDev: It was already here in the previous connector versions, i dont know. 

The thing is : all pages from the same chapters are named after their resolution and not the page number.
I guess its a folder structure like /manga/11111/chapter/22222/page1/ for all page 1 of the chapter. I guess they provide high res for all manga, i saw no complains.

Sheepux: we'll see if anyone complains

Sheepux (approved):

PR #5352, opened 2023-01-10
9anime fixes

MikeZeDev:
Relevant file https://github.com/manga-download/hakuneko-plugins/blob/master/src/NineAnime.js
* Fixes getting animes chapters pages and clipboard
* Support Subbed and Dubbed episodes
* Removed Streamango and Openload (hosters are dead)
* Add a new video hoster that works, FileMoon
* VideoHosters still needs to be fixed


27136773:
Please link the related commit(s) with all changes for the obfuscated file(s)
e.g., https://github.com/manga-download/hakuneko-plugins/commit/0e49950f4b4083480d7cfb78f9bec0a01be03862

27136773 (approved):

PR #5374, opened 2023-01-15
MadTheme : Fix missing pictures

MikeZeDev:
final_images only contains lazy loaded images, missing the first 3.

Affected Websites : MangaTheme websites without custom _getpages


src/web/mjs/connectors/templates/MadTheme.mjs
```diff
@@ -54,7 +54,8 @@ export default class MadTheme extends Connector {
     async _getPages(chapter) {
         let scriptPages = `
         new Promise(resolve => {
-            resolve(final_images);
+        	  let images = window.chapImages.split(',');
```
27136773: Spaces got replaced by tab, was this by intention?

27136773 (approved):

PR #5375, opened 2023-01-15
Tenshi.Moe : rewrite everything

MikeZeDev:
* They changed name and URL to Marin.moe 
* Old code doesn't work at all
* They have an api we can use now

### DIFF
--- a/src/web/mjs/connectors/TenshiMoe.mjs
+++ b/src/web/mjs/connectors/TenshiMoe.mjs
@@ -127,7 +127,7 @@ export default class TenshiMoe extends Connector {
         const data = await response.json();
         return data.props.episode_list.data.map(element => {
             return {
-                id: `/anime/${manga.id}/${+element.slug}`,
+                id: `/anime/${manga.id}/${element.slug}`,
                 title: element.slug + ' : '+ element.title.trim()
             };
         });
--- a/src/web/mjs/connectors/TenshiMoe.mjs
+++ b/src/web/mjs/connectors/TenshiMoe.mjs
@@ -6,9 +6,10 @@ export default class TenshiMoe extends Connector {
     constructor() {
         super();
         super.id = 'tenshimoe';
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/TenshiMoe.mjs
```diff
@@ -61,78 +62,100 @@ export default class TenshiMoe extends Connector {
...
+            }
+        };
+
+        const request = new Request(uri, {
+            method: 'POST',
+            headers: {
+                'Accept': 'text/html, application/xhtml+xml',
+                'Content-Type': 'application/json',
+                'x-origin' : this.url,
+                'x-referer' : uri,
```
27136773: Consider using explicit `uri.href` instead of relying on `.toString()`

src/web/mjs/connectors/TenshiMoe.mjs
```diff
@@ -61,78 +62,100 @@ export default class TenshiMoe extends Connector {
...
+         resolve(document.cookie) 
+          )` );
+
+        token = decodeURIComponent(token.match(/XSRF-TOKEN=([\S]+)/)[1]);
+
+        request = new Request(new URL('/anime/'+manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request, 'div#app');
+        data = JSON.parse(data[0].dataset['page']);
+        const pageCount = data.props.episode_list.meta.last_page;
+        this.version = data.version;
```
27136773: Seems like `version` could make into a local variable instead of global and provided as parameter to `_getchaptersFromPage(...)` the same way as it is done with `token`

27136773 (approved):

MikeZeDev:
Ahaha they changed api again, gotta fix that anyway

MikeZeDev:
Website returns error 403. I fear changes again, not gonna push for now.

PR #5384, opened 2023-01-17
PDF Export Fix soft lock when encountering broken images

### LINKED ISSUE #5380: [Bug] PDF Export soft locked when at least 1 image on source fails to download
Yes

**Describe the bug**
When downloading certain mangas the download queue gets stuck at 100% forever. This is easily reproduceable bellow.

**To Reproduce**
1. Set the "Chapter File Format" in the settings to "Portable Document (*.pdf)"
2. Go to the Connector "WebToonily"
3. Search for the manga "Leveling Up, By Only Eating!"
4. Attempt to download any of the following chapters: 2, 3 , 4, 5, 6, 9, 10, 11, 12, 14, 16, 23, 25, 26, 27, 28, ... 76, I first observed this on chapter 76.
5. Observe that the download never finishes, in addition, if you are watching the network traffic you should observe that some of the images are broken

**Expected behavior**
The app throws an error on broken images and moves on to the next download

**Screenshots**
N/A

 - OS: Windows 11
 - HakuNeko Version: 6.17@554b0c

**Additional context**
In addition to this I've pulled the current master branch and have fixed the issue locally, however when attempting to push my branch I am getting a 403.

To fix this the following code needs to be adding to "web/mhs/engine/Storage.mjs":
"_saveChapterPagesPDF"
Encapsulate the entire function in a try catch block, with a throw statement in the catch part.

"_addImageToPDF"
Encapsulate the entire function in a try catch block, with a throw statement in the catch part. As well as for the bitmap promise add: "resolve(img)" to the end of the promise.

That is all that is needed.

himdo:
- Fixes #5380 .

With this pull request downloads that get stuck when exporting to PDF will now error out as expected and continue with the rest of the download.

### DIFF
--- a/src/web/mjs/engine/Storage.mjs
+++ b/src/web/mjs/engine/Storage.mjs
@@ -504,43 +504,52 @@ export default class Storage {
      * Callback will be executed after completion and provided with an array of errors (or an empty array when no errors occured).
      */
     async _saveChapterPagesPDF(pdf, pageData) {
-        var doc = new PDFDocument({ autoFirstPage: false });
-        doc.pipe(this.fs.createWriteStream(pdf));
-        for (let page of pageData) {
-            await this._addImageToPDF(doc, page);
+        try {
+            var doc = new PDFDocument({ autoFirstPage: false });
+            doc.pipe(this.fs.createWriteStream(pdf));
+            for (let page of pageData) {
+                await this._addImageToPDF(doc, page);
+            }
+            doc.end();
+        } catch (error) {
+            throw error
         }
-        doc.end();
     }
 
     /**
      * Add a single image as PDF page to the given document.
      */
     async _addImageToPDF(pdfDocument, page) {
-        let bitmap = await new Promise(resolve => {
-            let img = new Image();
-            img.onload = () => resolve(img);
-            img.src = URL.createObjectURL(page.data);
-        });
-        let pdfImgType = this._pdfImageType(page);
-        let blob;
-        if (!pdfImgType) {
-            pdfImgType = 'JPEG';
-            let canvas = document.createElement('canvas');
-            canvas.width = bitmap.width;
-            canvas.height = bitmap.height;
-            let ctx = canvas.getContext('2d');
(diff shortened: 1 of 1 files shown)


MikeZeDev:
LGTM. Dont know this code enough to validate tho. Ill wait for other reviewers.

MikeZeDev:
In the meantime you can fix lint errors https://github.com/manga-download/hakuneko/actions/runs/3941947326/jobs/6757610800

27136773:
Looks to me like there are way too many changes which are not related to the problem.
Wouldn't it just be enough to make sure the `Promise` for the bitmap is correctly rejected when an error occurs?

e.g.,
```javascript
let bitmap = await new Promise((resolve, reject) => {
    let img = new Image();
    img.onerror = () => reject(new Error('Failed to load image!'));
    img.onload = () => resolve(img);
    img.src = URL.createObjectURL(page.data);
});
```

src/web/mjs/engine/Storage.mjs
```diff
@@ -504,43 +504,52 @@ export default class Storage {
...
-            let ctx = canvas.getContext('2d');
-            ctx.drawImage(bitmap, 0, 0);
-            blob = await new Promise(resolve => {
-                canvas.toBlob(data => resolve(data), 'image/jpeg', 0.90);
+        try {
+            let bitmap = await new Promise(resolve => {
+                let img = new Image();
+                img.onload = () => resolve(img);
+                img.src = URL.createObjectURL(page.data);
+                resolve(img)
```
27136773: This will always resolve the promise immediately, even if the image has not yet finished loading or an error occurs during loading.
=> Would be better to reject the promise properly, see other comment for suggestion

himdo:
> Looks to me like there are way too many changes which are not related to the problem. Wouldn't it just be enough to make sure the `Promise` for the bitmap is correctly rejected when an error occurs?

I can confirm that the try catches are not needed so I have removed those.
@ronny1982 after testing your suggestion I can confirm that fix does work for me, so I have implemented it and pushed it up.

27136773 (approved):

PR #5399, opened 2023-01-20
Bilibili : Language, login support and clipboard fixes

MikeZeDev:
* Only 2 connectors needed : Bibilicomics and BibiliManhua
* Clipboard feature works again now (only one Bilibilicomics.com connector) 
* Bilibilicomics  is the multilingual one and handle Bilibilicomics  .com languages through a setting
* Deletion of Bibili ES, FR, ID as they are handled by Bilibilicomics  
* Bilibilimanhua ,is Chinese only and url is manga.bilibili.com. 
* Bilibilicomics extends Bilibilimanhua, it only override its config so user can select language.
* Login pages are different
* Language setting affect fetched manga list.
* Login & downloading unlocked/free chapter was tested for Bilibilicomics  . Free download was tested for Bilibilimanhua
* May require testing for paid stuff on Bilibilimanhua


src/web/mjs/connectors/BilibiliManhua.mjs
```diff
@@ -9,8 +9,15 @@ export default class BilibiliManhua extends Connector {
...
+        this.token_expires_at = -1;
+        this.auth = {
+            accessToken : '',
+            refreshToken : '',
+        };
+        this.links = {
+            login: 'https://passport.bilibili.com/login'
+        };
+        //default config doesnt need language switch, its fixed to CN
         this.config = {
```
27136773, line 21: May add `language` property with `input: 'disabled'` and value set to `cn`.
With this all the ternary checks for `this.config.language` could be removed.
Disadvantage would be that this is [shown in settings](https://github.com/manga-download/hakuneko/blob/master/src/web/lib/hakuneko/frontend%40classic-light/input.html#L32-L36)...

27136773:
The token expiration handling looks quite over-engineered, couldn't we just always request a new token?
I'm pretty sure the backend will handle expired tokens and also verify the cookies that are send with the request and deliver either a new token, or the last valid token.

27136773 (approved):
May simplify token handling

MikeZeDev:
That may be over the top but like i've said, i'd rather mimicking what the websites does in order to avoid further problems. And i think its the proper way : website doesn't seems to ask for a new token until its expired. You may take that as a training on how to handle JWT if you want.

Tbh if i'd wanted to make that more complicated i would have made sure that the cookie stored is updated with the new values.
I couldnt do that because electron didnt let me write a cookie for ".bilibili.com' , only for "www.bilibili.com" .

src/web/mjs/connectors/BilibiliManhua.mjs
MikeZeDev: I'll do that.

PR #5404, opened 2023-01-21
FIX: Qiman5 - Add Url Detection

### LINKED ISSUE #5402: [Qiman5] Connector not working
Yes

Yes

**Describe the bug**
The chapter lists are not loading. Old downloaded chapters appear but are greyed out eg http://m.qiman58.com/16041/
hakuneko2

**To Reproduce**
1. Qiman5

**Screenshots**
hakuneko

 - Version 6.1.7@554b0c

MikeZeDev:
* Get correct url from default url
* Fixes https://github.com/manga-download/hakuneko/issues/5402

### DIFF
--- a/src/web/mjs/connectors/Qiman5.mjs
+++ b/src/web/mjs/connectors/Qiman5.mjs
@@ -1,21 +1,45 @@
-import Connector from '../engine/Connector.mjs';
-import Manga from '../engine/Manga.mjs';
+import SinMH from './templates/SinMH.mjs';
 
-export default class Qiman5 extends Connector {
+export default class Qiman5 extends SinMH {
 
     constructor() {
         super();
         super.id = 'qiman5';
         super.label = '奇漫屋 (Qiman5)';
         this.tags = [ 'webtoon', 'chinese'];
-        this.config = {
-            url: {
-                label: 'URL',
-                description: `This website's main domain doesn't always work, but has alternate domains.\nThis is the default URL which can also be manually set by the user.`,
-                input: 'text',
-                value: 'https://www.qiman59.com'
-            }
-        };
+        this.url = 'http://qimanw.com';
+
+        this.path = '/rank/1-%PAGE%.html';
+        this.queryMangas = 'div.mainForm div.updateList div.bookList_3 div.ib p.title a';
+        this.queryChapters = 'div.chapterList div#chapter-list1 a.ib';
+        this.queryManga = 'div.container div.mainForm div.comicInfo div.ib.info h1';
+        this.queryChaptersScript = `
+            new Promise((resolve, reject) => {
+                setTimeout(() => {
+                    try {
+                        let button = document.querySelector('.moreChapter');
+                        if(button) {
+                            button.click();
+                        }
+                        setTimeout(() => {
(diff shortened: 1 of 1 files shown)


MikeZeDev:
Still need to investigate, people miss content, i  need to reproduce that

MikeZeDev:
User said its ok.

src/web/mjs/connectors/Qiman5.mjs
```diff
@@ -53,6 +53,19 @@ export default class Qiman5 extends SinMH {
         `;
     }
 
+    canHandleURI(uri) {
+        super.initialize()
+            .then(() => {
```
27136773: This does not work as intended, `canHandleURI` is expected to return a `bool`, but the current implementation returns a `Promise<bool>`, therefore it will always be true even if the promise resolves to false.

src/web/mjs/connectors/Qiman5.mjs
```diff
@@ -53,6 +53,19 @@ export default class Qiman5 extends SinMH {
...
+                return uri.href.includes(this.url);
+            });
+    }
+
+    async _initializeConnector() {
+        const request = new Request(this.url, this.requestOptions);
+        this.url = (await this.fetchDOM(request, 'a#href')).pop().href;
```
27136773: This may be error prone in case the last found `anchor` element is a cross reference to a different domain.

27136773: Aslo shouldn't it use `.origin` instead of `.href`?

src/web/mjs/connectors/Qiman5.mjs
```diff
@@ -53,6 +53,19 @@ export default class Qiman5 extends SinMH {
...
 
+    canHandleURI(uri) {
+        super.initialize()
+            .then(() => {
+                return uri.href.includes(this.url);
+            });
+    }
+
+    async _initializeConnector() {
+        const request = new Request(this.url, this.requestOptions);
```
27136773, line 69: What if the current `this.url` is already wrong, will this still work?

MikeZeDev: this.url is the "portal" url, like for many connectors. The difference is its doesnt use redirect, people must click on a single link on the page. How should i do?

src/web/mjs/connectors/Qiman5.mjs
MikeZeDev: Hum, but we must initialize the connector before otherwise copypaste wont work. I cant call initialize with await without async and i shouldnt prefix canHandleURI with async, right? 

Btw its a common occurrence when connector have to assign an url and first thing user do is pasting.
What should i do ?

src/web/mjs/connectors/Qiman5.mjs
MikeZeDev: 1) There is only one 'a' element on the page, and its 'id' is 'href'. There shouldnt be more than one 'a' node with the same 'id' 🤷 .  And since its a portal page, its a given that it will be to another domain, i dont get it.

2) True, origin is more error resistant.

PR #5419, opened 2023-01-25
Mangareader.to : Fix timeout problems

### LINKED ISSUE #5054: [mangareader.to] Volume download doest work
Yes

**Describe the bug**
I can't download manga in Volumes from mangareader.to connector. It's all good on site's side, but when i try to download volume, it always times out. Chapters on other hand are downloading easier

**To Reproduce**
1. Go to mangareader.to connector (i'm not sure if there are any other connector with volumes to download)
2. Click on Naruto (that's where i encountered a problem, but i think it works with everything else too)
3. Try to download VOL 1 (EN)
4. It will timeout, but if you try to download any chapter, it will download correctly

My guess that it's something with the timeout feature and increased size of volumes, compared to chapters. Is there any way to increase time for timeout manually?

**Context:**
 - OS: Windows 7
 - HakuNeko Version 6.17

### LINKED ISSUE #5229: [Feature] Volumes download support
When iam trying to download volumes some of them work the other one doesn't 

Maybe expanding download size if that is possible 
Or when you can't download the file to be split into separate parts 

I know downloading the chapters are the better idea but honestly if there are Volumes available I prefer them because i have less files and its more easy to read than the chapters (opening, closing, opening the next one....)

Thanks for your hard work

### LINKED ISSUE #5418: [Mangareaderto] failed to load within the given timeout of sixty seconds when downloading volumes.
The Bug
I downloaded Hakuneko to download mange. I got it working and then found out I could also download in volumes, but when I click download after a minute it says that it failed to load whitin the given timeout of sixty seconds. I also can't view the manga as it gives me the same error. It is not a connector issue, because the chapters do work. I think it is because the volumes are to big to load within sixty seconds. Is there a way to change the sixty second time period or maybe something else?

To Reproduce
1. Load a connector that has volumes( i tried mangareader as i couldn't find a other one that could).
2. earch for a manga ( I tried The quintessential quintuplets, my hero academia, fire force and ayakashi triangle).
3. Click download and after sixty seconds it gives a error.

Additional context
My os is windows 10 and I run version 6.1.7 of hakuneko

Thanks in advance.

MikeZeDev:
This arbitrary value seems to be okay. If you have any objection feel free to voice it.

Fixes https://github.com/manga-download/hakuneko/issues/5418, Fixes https://github.com/manga-download/hakuneko/issues/5054, Fixes https://github.com/manga-download/hakuneko/issues/5229

### DIFF
--- a/src/web/mjs/connectors/MangaReaderTo.mjs
+++ b/src/web/mjs/connectors/MangaReaderTo.mjs
@@ -110,6 +110,12 @@ export default class MangaReaderTo extends Connector {
         let data = await this._blobToBuffer(blob);
         this._applyRealMime(data);
         return data;
+
+    }
+
+    async base64toArraybuffer(dataURL) {
+        const arr = dataURL.split(',');
+        return Uint8Array.from(atob(arr[1]), c => c.charCodeAt(0));
     }
 
     /**********************/
@@ -127,6 +133,9 @@ export default class MangaReaderTo extends Connector {
         });
     }
 
+    //Slightly modified from Mangareader lambda that extends Math by adding Math.seedrandom  function.
+    //1) Modified because we dont want to mess with modules.exports (i bet they have a node application to scramble and unscramble their pics)
+    //2) Main purpose of init is to fill this.seedrandom with an adequate lambda to use for unscramblng pictures. Too lazy too debug all that maybe there is still useless code.
     init() {
         let _0xb2d0a3 = [];
         var _0x221936,
@@ -153,7 +162,7 @@ export default class MangaReaderTo extends Connector {
                 return (_0x16995d + _0x11164e) / _0x374e1e;
             }
             var _0x357c70 = [],
-                _0xd1c90b = _0x459744(
+                _0xd1c90b = _0x459744(// eslint-disable-line
                     (function _0x53f196(_0x598061, _0xa97de4) {
                         var _0x4cb136,
                             _0x588b46 = [],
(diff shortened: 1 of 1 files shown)


27136773:
Instead of processing all images at once, may try a different approach.
Use `_handleConnectorURI` and do the descrambling one by one per image.
The overall processing time will likely increase, but the work is split into shorter running tasks.

MikeZeDev:
> Instead of processing all images at once, may try a different approach. Use `_handleConnectorURI` and do the descrambling one by one per image. The overall processing time will likely increase, but the work is split into shorter running tasks.

Fine. How should i return the canvas data in _handleConnectorURI?

MikeZeDev:
> Instead of processing all images at once, may try a different approach. Use `_handleConnectorURI` and do the descrambling one by one per image. The overall processing time will likely increase, but the work is split into shorter running tasks.

```
        const type = Engine.Settings.recompressionFormat.value;
    	const url = payload.requestChapter;
        const quality = parseFloat(Engine.Settings.recompressionQuality.value) / 100;
        const script = `
            new Promise(async (resolve, reject) => {
                try {
                        const canvas = await imgReverser('${payload.url}');
                        const uri = canvas.toDataURL('${type}', ${quality});
                        resolve(uri);
                    }
                 catch(error) {
                    reject(error);
                }
            });
        `;
        const request = new Request(url, this.requestOptions);
        const fixedImage = await Engine.Request.fetchUI(request, script, 60000, true);
        let array = await this.base64toArraybuffer(fixedImage);
        return {data : array, mimetype: type};

    }

    async base64toArraybuffer(dataURL) {
        const arr = dataURL.split(',');
        return Uint8Array.from(atob(arr[1]), c => c.charCodeAt(0));
    }

```

This one is slow and keeps giving me timeouts on volumes. While the previous solution allowed me to download volumets IN BULK without a hassle.
Edit : Hakuneko even crashed while getting images.

Doing the work in _handleConnectorURI means multiplying requests to the page with script execution meaning the full page is executed each time and load stuff that bloat memory. I may have done something wrong .

The solution would be to recode their unscrambling, its not heavily obfuscated but fragmented (many functions) but damn i ma lazy.

MikeZeDev:
Copied descrambling from their javascript but there is some LINT ERRORS idk how to fix. However, its fast and stable .

no-loss-of-precision : not sure its a problem here
no redefine : yeah they redefine argument from the function but the initial value seems to be used in the process and code is a mess.

27136773 (approved):

PR #5433, opened 2023-01-29
[Pixiv] Fix getting chapter list

Druzai:
If manga has less than 5 chapters, no chapters will be found.
Example: https://www.pixiv.net/user/46393/series/183896
Also I added artworks processing, because there are many mangas that are single chapter.

### DIFF
--- a/src/web/mjs/connectors/Pixiv.mjs
+++ b/src/web/mjs/connectors/Pixiv.mjs
@@ -35,25 +35,25 @@ export default class Pixiv extends Connector {
             const chapters = await this._getChaptersFromPage(manga.id, page);
             chapters.length > 0 ? chapterList.push(...chapters) : run = false;
         }
-        return chapterList.filter((chapter, index) => {
-            return index === chapterList.findIndex(item => chapter.id === item.id);
-        });
+        return chapterList;
     }
 
     async _getChaptersFromPage(mangaId, page) {
         const uri = new URL(`series/${mangaId}?p=${page}&lang=en`, this.apiURL);
         const request = new Request(uri, this.requestOptions);
         const data = await this.fetchJSON(request);
-        if (data.body.thumbnails.illust.length < 5)
-            return [];
-        return data.body.thumbnails.illust
-            .filter(chapter => chapter.seriesId === mangaId)
+        return data.body.page.series
             .map(chapter => {
-                return {
-                    id: chapter.id,
-                    title: chapter.title.trim()
-                };
-            });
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Pixiv.mjs
```diff
@@ -31,29 +41,37 @@ export default class Pixiv extends Connector {
...
+                title: manga.title
+            });
+        } else {
+            for (let page = 1, run = true; run; page++) {
+                const chapters = await this._getChaptersFromPage(manga.id, page);
+                chapters.length > 0 ? chapterList.push(...chapters) : run = false;
+            }
         }
-        return chapterList.filter((chapter, index) => {
-            return index === chapterList.findIndex(item => chapter.id === item.id);
```
MikeZeDev, line 39: Hum i am curious about this filter.

src/web/mjs/connectors/Pixiv.mjs
```diff
@@ -31,29 +41,37 @@ export default class Pixiv extends Connector {
...
-                };
-            });
+                const chapterContents = data.body.thumbnails.illust.find(c => c.id === chapter.workId);
+                if (chapterContents != undefined) {
+                    return {
+                        id: chapterContents.id,
+                        title: chapterContents.title.trim()
+                    };
+                } else {
+                    return null;
```
MikeZeDev: Why not just mapping data.body.page.series to manually populate an array of chapter you can return later? 
I am not fond of letting null elements in chapter array and use a filter after, its just useless to add them first.
```

let chapterlist = []
const data = await this.fetchJSON(request);

data.body.page.series .map(chapter => {
    const chapterContents = data.body.thumbnails.illust.find(c => c.id === chapter.workId);
    if (chapterContents) {
         chapterlist .push(  { id: chapterContents.id,     title: chapterContents.title.trim()  });
      }// else do nothing 
}
return chapterlist 

```

src/web/mjs/connectors/Pixiv.mjs
Druzai: I added it because before I just filtered chapters by `mangaId`.
Using `data.body.page.series` won't produce duplicates. So i removed this filter.
Well, as far as I've tested, I haven't found any duplicates yet...

src/web/mjs/connectors/Pixiv.mjs
Druzai: Changed the algorithm to populate the list in 12f9713 as you suggested, except I used `forEach` instead of `map`.

MikeZeDev (approved):
LGTM

PR #5455, opened 2023-02-06
SaveAsPdf : Workaround for PDFKIT png lib bug

### LINKED ISSUE #5450: [Download as PDF] Maximum Call stack exceeded.
Yes

Yes

**Describe the bug**

**To Reproduce**
1. Name of connector:https://bato.to/
2. Name of manga: The Villainess Turns the Hourglass  (https://bato.to/series/78871/the-villainess-turns-the-hourglass) 
3. Name of chapter 17-10
Maximum Call stack exceeded. 

**Screenshots**
image

 - Version 6.1.7@554b0c

MikeZeDev:
What error : 
---------------
Max stack exceeded when creating PDF

Reproduce the error : 
--------------------------------
Try so save as PDF this chapter : https://bato.to/chapter/1452754. We get a stack overflow when adding the last page which is a 4.5MB png.

It turns out the error is in the PNG lib included in our PDFKIT version (when parsing tExT chunk). 

JPG files doesnt have this problem apparently. Maybe because PDFKit handle them properly or just because we dont find exceeding 4MB easily ? Need further testing .

Proposed workaround : 
---------------------------
Force JPG conversion if image is not jpeg (its already the case for WEBP, now it will be the same for PNG).

Another possibility would be try.catch page addition and convert the png only in case of error.

References : Fixes https://github.com/manga-download/hakuneko/issues/5450

### DIFF
--- a/src/web/mjs/engine/Storage.mjs
+++ b/src/web/mjs/engine/Storage.mjs
@@ -855,9 +855,9 @@ export default class Storage {
         if (image.type === 'image/jpeg') {
             return 'JPEG';
         }
-        if (image.type === 'image/png') {
-            return 'PNG';
-        }
+        //if (image.type === 'image/png') {
+        //    return 'PNG';
+        //}
         return undefined;
     }
 


Druzai:
This error occurs because the `png-js` library uses recursion to concatenate an array of characters into a string.
If file [pdfkit.standalone.js](https://github.com/manga-download/hakuneko/blob/master/src/web/js/pdfkit.standalone.js) can be changed, you can change this algorithm from recursive to iterative implementation, e.g.
From
```js
        case 'tEXt':
          var text = this.read(chunkSize);
          var index = text.indexOf(0);
          var key = String.fromCharCode.apply(String, text.slice(0, index));
          this.text[key] = String.fromCharCode.apply(
            String,
            text.slice(index + 1)
          );
          break;
```
To
```js
        case 'tEXt':
          var text = this.read(chunkSize);
          var index = text.indexOf(0);
          var key = String.fromCharCode.apply(String, text.slice(0, index));
          this.text[key] = '';
          text.slice(index + 1).forEach(element => {
            this.text[key] += String.fromCharCode(element);
          });
          break;
```
About that `tEXt` chunk in png image:
- I checked several images and not all of them have this chunk (for example [this chapter](https://mangadex.org/chapter/cba49c22-6882-422d-8277-eae663b90bae) was successfully converted to pdf)
- I also tried to make pdf from 1 png image that doesn't contain a `tEXt` chunk and is 11MB in size. The pdf was created successfully, so the problem isn't the file size.

So if the suggestion above can't be implemented, it's better to use `try catch` instead of converting all png images to jpeg.

MikeZeDev:
I mentioned it was a tExT chunk issue in the first place but thanks for the alternate (and better) solution.
Should i close this PR and you correct this issue properly?

Druzai:
> Should i close this PR and you correct this issue properly?

Yes, I'll open a new PR for it. Although you can edit that file in this PR (line 59051), but okay.

MikeZeDev:
Yeah but it would be cleaner if done in another PR. Thanks :)

PR #5463, opened 2023-02-11
Hiperdex : url as setting

### LINKED ISSUE #5409: [Hiperdex] Connector not working
Hiperdex is now changed to following address:
https://1sthiperdex.com/ [New address]
https://hiperdex.com/ [Old address]

**To Reproduce**
1. Connect to Hiperdex
2. Search any manga
3. Select any chapter
4. Nothing will load

 - Version 6.1.7

MikeZeDev:
They change between 2 url too frequently and when they do that one of the two may be unreachable.
Fixes : https://github.com/manga-download/hakuneko/issues/5409

### DIFF
--- a/src/web/mjs/connectors/Hiperdex.mjs
+++ b/src/web/mjs/connectors/Hiperdex.mjs
@@ -21,13 +21,6 @@ export default class Hiperdex extends WordPressMadara {
         return this.config.url.value;
     }
 
-    set url(value) {
-        if (this.config && value) {
-            this.config.url.value = value;
-            Engine.Settings.save();
-        }
-    }
-
     canHandleURI(uri) {
         return this.url.includes(uri.hostname);
     }
--- a/src/web/mjs/connectors/Hiperdex.mjs
+++ b/src/web/mjs/connectors/Hiperdex.mjs
@@ -10,9 +10,14 @@ export default class Hiperdex extends WordPressMadara {
         this.config = {
             url: {
                 label: 'URL',
-                description: 'This website changes their URL regularly.\nThis is the last known URL which can also be manually set by the user.',
-                input: 'text',
+                description: 'This website changes between those two URL regularly.\n Please select the one to use',
+                input: 'select',
+                options: [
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Hiperdex.mjs
```diff
@@ -7,6 +7,29 @@ export default class Hiperdex extends WordPressMadara {
...
+                value: 'https://1sthiperdex.com'
+            }
+        };
     }
+
+    get url() {
+        return this.config.url.value;
+    }
+
+    set url(value) {
```
27136773, line 29: Is this setter used?
If not, may be removed...

src/web/mjs/connectors/Hiperdex.mjs
```diff
@@ -7,6 +7,29 @@ export default class Hiperdex extends WordPressMadara {
         super.id = 'hiperdex';
         super.label = 'Hiperdex';
         this.tags = [ 'hentai', 'webtoon', 'english' ];
-        this.url = 'https://hiperdex.com';
+        this.config = {
```
27136773, line 10: Seems to me once pushed to the user, the URL can no longer be automatically be updated by the developers, the user must always manually update the URL.
Is this by intention?

MikeZeDev: Its like this for connectors where url is in settings afaik. Like i've said i would have detected the url  in initialize instead of putting a setting but Hiperdex also sometimes fire Cloudflare and one of the two domains may be unreachable (unrelated to CF)

I can ofc change to usual domain detection but there is no "master domain".

27136773: Maybe use a `'select'` [type setting](https://github.com/manga-download/hakuneko/blob/20c5e61c46b71de379a4635a21018f406d32c42b/src/web/mjs/engine/Settings.mjs#L135-L141) with pre-defined mirrors instead of a `'text'` type to make it more suitable for users?

27136773 (approved):

PR #5471, opened 2023-02-13
Add new website cosmicscansID, komikbree, rrscans, moonscans

KazielZoe:
Add new website cosmicscans.id, komikbree.xyz, rrscans.my.id, moonscans.xyz


src/web/mjs/connectors/CosmicScansID.mjs
```diff
@@ -0,0 +1,13 @@
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
+
+export default class CosmicScans extends WordPressMangastream {
```
MikeZeDev: * Class must match filename and be unique

src/web/mjs/connectors/KomikBree.mjs
```diff
@@ -0,0 +1,12 @@
+import WordPressMadara from './templates/WordPressMadara.mjs';
+
+export default class Aiinscan extends WordPressMadara {
```
MikeZeDev: * Class must match filename and be unique

src/web/mjs/connectors/MoonScans.mjs
```diff
@@ -0,0 +1,13 @@
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
+
+export default class CosmicScans extends WordPressMangastream {
```
MikeZeDev: * Class must match filename and be unique

src/web/mjs/connectors/RRScans.mjs
```diff
@@ -0,0 +1,13 @@
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
+
+export default class CosmicScans extends WordPressMangastream {
```
MikeZeDev: * Class must match filename and be unique

MikeZeDev (changes requested):
* Class names fixes required
* Please provide 4 icons for the 4 websites.

MikeZeDev:
Icons files are way to big. Ideally they should be png, < 16kb and dimension max 96x96 px

MikeZeDev:
Cosmics Scans pictures are behind Cloudflare , you should add refer and origin headers otherwise it doesnt work.

```
        this.requestOptions.headers.set('x-referer', this.url);
        this.requestOptions.headers.set('x-origin', this.url);
```

image
image

MikeZeDev:
Many komikbree mangas have 'Bahasa indonesia' in name, You may want to get rid of that
Same for RRscans

Also RR scans chapters randomly doesnt work because ads are messing with us (Girl und panzer)
Not sure something can be done about that. No problem if you cant fix btw.

image

27136773:
@KazielZoe 
No response since more than `6 weeks` after the last review feedback, is this still worked on?

Would be sad if we had to abandon the changes because of inactivity...

PR #5537, opened 2023-03-05
SaveAsPdf : Fixed reading 'tEXt' chunk for 'png-js' lib

### LINKED ISSUE #5450: [Download as PDF] Maximum Call stack exceeded.

Druzai:
Continuation of PR #5455
Fixes #5450

### DIFF
--- a/src/web/js/pdfkit.standalone.js
+++ b/src/web/js/pdfkit.standalone.js
@@ -59048,10 +59048,10 @@ module.exports = class PNG {
           var text = this.read(chunkSize);
           var index = text.indexOf(0);
           var key = String.fromCharCode.apply(String, text.slice(0, index));
-          this.text[key] = String.fromCharCode.apply(
-            String,
-            text.slice(index + 1)
-          );
+          this.text[key] = '';
+          text.slice(index + 1).forEach(element => {
+            this.text[key] += String.fromCharCode(element);
+          });
           break;
 
         case 'IEND':


MikeZeDev:
I'll approve this since i've tested it and it works.

MikeZeDev (approved):

27136773:
Library code should not be patched, otherwise we have to apply all the changes again whenever the library is updated.
I suggest to directly commit the fix to https://github.com/foliojs/pdfkit instead and then update the library in HakuNeko.

Druzai:
I don't think I can commit the fix to `pdfkit` lib because this error occurs in constructor of `PNG` class. That class is taken from `png-js` lib as a dependency.

I can think of 2 ways to solve the problem:
1. Committing the fix to [pdfkit](https://github.com/foliojs/pdfkit) lib
    - Inherit and extend `PNG` class and change its constructor to fix this problem
    - I don't think it's the right way, because code of `png-js` lib can change. Also it's just duplicating constructor code with 1 small fix... (if there is no other way I will do it)
2. Committing the fix to [png-js](https://github.com/foliojs/png.js) lib
    - Fix constructor of `PNG` class
    - It's more appropriate, but author of that repository has been inactive since 2019 (PRs from 2019 and 2020 haven't been reviewed at all). I think this fix won't be merged.

I think each way is not good enough.

Druzai:
Also the third way was suggested in previous PR as a workaround:
> Another possibility would be try.catch page addition and convert the png only in case of error.

MikeZeDev:
I wanted to submit a PR to pdf-js but like Druzai said judging by folio lack of activity it will most likely never be merged.

On the other hand, it means it probably wont be updated anymore. For that reason i deemed patching the lib an acceptable solution 🤷

PR #5545, opened 2023-03-08
[MangaPark] Updated connectors for MangaPark v5

Druzai:
Noticed that MangaPark finally switched to v5.
Updated connectors to use graphQL endpoints.

### DIFF
--- a/src/web/mjs/connectors/MangaParkEN.mjs
+++ b/src/web/mjs/connectors/MangaParkEN.mjs
@@ -27,9 +27,13 @@ export default class MangaParkEN extends Connector {
 
     async _getMangas() {
         try {
-            return await this._getMangasV3();
+            const mangaList = await this._getMangasV3();
+            if (mangaList.length == 0) {
+                throw new Error('Got 0 mangas from site using v3 version! Using v5 version as a fallback!');
+            }
+            return mangaList;
         } catch (error) {
-            return this._getMangasV5();
+            return await this._getMangasV5();
         }
     }
 
@@ -109,9 +113,13 @@ export default class MangaParkEN extends Connector {
 
     async _getChapters(manga) {
         try {
-            return await this._getChaptersV3(manga);
+            const chapterList = await this._getChaptersV3(manga);
+            if (chapterList.length == 0) {
+                throw new Error('Got 0 chapters from site using v3 version! Using v5 version as a fallback!');
+            }
+            return chapterList;
         } catch (error) {
-            return this._getChaptersV5(manga);
+            return await this._getChaptersV5(manga);
         }
     }
 
--- a/src/web/mjs/connectors/MangaParkEN.mjs
+++ b/src/web/mjs/connectors/MangaParkEN.mjs
@@ -11,8 +11,6 @@ export default class MangaParkEN extends Connector {
         this.url = 'https://mangapark.net';
         this.apiURL = `${this.url}/apo/`;
 
-        this.queryMangaTitle = 'main h3 a';
-
         this.requestOptions.headers.set('x-origin', this.url);
         this.requestOptions.headers.set('x-referer', `${this.url}/`);
         this.requestOptions.headers.set('x-cookie', 'set=h=1;');
@@ -20,7 +18,8 @@ export default class MangaParkEN extends Connector {
 
     async _getMangaFromURI(uri) {
         const request = new Request(uri, this.requestOptions);
-        const data = await this.fetchDOM(request, this.queryMangaTitle);
+        const queryMangaTitle = /\/title\/\d+/.test(uri.pathname) ? 'main h3 a' : 'meta[property="og:title"]';
+        const data = await this.fetchDOM(request, queryMangaTitle);
         const id = uri.pathname.match(/\/(\d+)\/?/)[1];
         const title = (data[0].textContent || data[0].content).trim();
         return new Manga(this, id, title);
--- a/src/web/mjs/connectors/MangaParkEN.mjs
+++ b/src/web/mjs/connectors/MangaParkEN.mjs
@@ -20,21 +20,55 @@ export default class MangaParkEN extends Connector {
         const request = new Request(uri, this.requestOptions);
         const queryMangaTitle = /\/title\/\d+/.test(uri.pathname) ? 'main h3 a' : 'meta[property="og:title"]';
         const data = await this.fetchDOM(request, queryMangaTitle);
-        const id = uri.pathname.match(/\/(\d+)\/?/)[1];
+        const id = parseInt(uri.pathname.match(/\/(\d+)\/?/)[1]);
         const title = (data[0].textContent || data[0].content).trim();
         return new Manga(this, id, title);
     }
 
     async _getMangas() {
+        try {
+            return await this._getMangasV3();
+        } catch (error) {
+            return this._getMangasV5();
+        }
+    }
+
+    async _getMangasV3() {
+        let mangaList = [];
+        const uri = new URL('/browse?sort=name', this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, 'nav.d-md-block ul.pagination li:nth-last-child(2) a');
+        const pageCount = parseInt(data[0].text.trim());
+        for(let page = 1; page <= pageCount; page++) {
+            let mangas = await this._getMangasFromPageV3(page);
+            mangaList.push(...mangas);
+        }
+        return mangaList;
+    }
+
+    async _getMangasFromPageV3(page) {
+        const uri = new URL(`/browse?sort=name&page=${page}`, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, '#subject-list div.item a.fw-bold');
+        return data.map( element => {
+            this.cfMailDecrypt(element);
+            return {
+                id: parseInt(this.getRootRelativeOrAbsoluteLink(element, request.url).match(/\/(\d+)\/?/)[1]),
+                title: element.text.trim()
+            };
+        });
+    }
+
+    async _getMangasV5() {
         let mangaList = [];
         for(let page = 1, run = true; run; page++) {
-            const mangas = await this._getMangasFromPage(page);
+            const mangas = await this._getMangasFromPageV5(page);
             mangas.length > 0 ? mangaList.push(...mangas) : run = false;
         }
         return mangaList;
     }
 
-    async _getMangasFromPage(page) {
+    async _getMangasFromPageV5(page) {
         const gql = `
             query get_content_browse_search($select: ComicSearchSelect) {
                 get_content_browse_search(select: $select) {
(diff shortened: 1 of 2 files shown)


MikeZeDev:
Nice work.

Better than using innerHTML to get __NEXT_DATA__ , why not using fetchui with a script since its a global variable? 
And doing the groundwork in the script maybe? 
There are plenty of exemples with __NEXT_DATA__  or __NUXT__ in connectors.

Also, i may be blind but there is not such variable on https://mangapark.net/comic/10953/one-piece/c1076-en (tried this randomly website gave me this url)

Then it gave me this one https://mangapark.net/title/10953-one-piece/7828587-en-ch.1076 with a NEXT_DATA.

It seems to me they didnt fully switched to v5. Old links still works. Thats a problem.

Druzai:
> It seems to me they didnt fully switched to v5. Old links still works. Thats a problem.

Yes, my bad. I got 404 errors. With vpn those titles accessible.

For v5 I can move this logic to new connector and revert existing 2 connectors.
Or I can use scrapping logic for v3 by default. If that fails then fallback to use logic for v5 like I did in this [connector](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/WebComicsApp.mjs).

> Better than using innerHTML to get **NEXT_DATA**

Changed.

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -64,8 +133,23 @@ export default class MangaParkEN extends Connector {
             const chapters = [...this.createDOM(data.html).querySelectorAll('div.episode-item > div > a.chapt')].map(element => {
                 const link = this.getRootRelativeOrAbsoluteLink(element, this.url);
                 let lang = link.match(/c[\d.]+-(\w+)-i\d+/)[1];
-                if (lang)
-                    lang = lang.replace(/\D(_)\D/g, (match, g1) => match.replace(g1, '-'));
+                switch (lang) {
+                    case 'zh_hk':
+                        lang = 'zh-Hans';
+                        break;
+                    case 'zh_tw':
```
MikeZeDev: Just a shower thought : It means default lang is english and its unspecified? Maybe then its worth and cleaner to add a default value in switch too?

Druzai: Well, default language is most likely English, yes. But in those links they always specify language code, so it can't be null.
Example: `https://mangapark.net/comic/16009/tsuredure-children/c212.1-en-i3407661`.

On MangaPark some language codes doesn't match ISO standard, so I have to correct specific cases. I took these replacements from [here](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/5a8618f004a4a8560e6d0d480805bffe4794bd10/src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaParkFactory.kt).
If I add default case all other languages will be set as that default string, which is bad.

MikeZeDev:
Gotta take time to test and review this.

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -10,43 +9,113 @@ export default class MangaParkEN extends Connector {
...
         const data = await this.fetchDOM(request, queryMangaTitle);
-        return new Manga(this, uri.pathname.match(/\/(\d+)\/?/)[1], (data[0].textContent || data[0].content).trim());
+        const id = parseInt(uri.pathname.match(/\/(\d+)\/?/)[1]);
+        const title = (data[0].textContent || data[0].content).trim();
+        return new Manga(this, id, title);
     }
 
     async _getMangas() {
+        try {
+            return await this._getMangasV3();
```
MikeZeDev: There are chances that _getMangasV3 wont trigger an error but just return an empty array. In this case the V5 fallback wont be called. Maybe catching errors in _getMangasV3 return an empty array in this case and change the try catch here to a test if _getMangasV3 return [] to force v5 fallback?

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -10,43 +9,113 @@ export default class MangaParkEN extends Connector {
...
+                id: manga.data.id,
+                title: manga.data.name
+            };
+        });
+    }
+
     async _getChapters(manga) {
+        try {
+            return await this._getChaptersV3(manga);
+        } catch (error) {
```
MikeZeDev, line 129: Same remark than for getting mangas. Assuming V3 will create an error for using the fallback is uncertain.

Edit : Just found there are mangas without chapters at all : https://mangapark.net/comic/53695/fushigi-toriatsukaimasu-tsukumodou-kottouten . It will always fallback to v5 in my suggestion but thats not a problem at all.

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -10,43 +9,113 @@ export default class MangaParkEN extends Connector {
...
+        const id = parseInt(uri.pathname.match(/\/(\d+)\/?/)[1]);
+        const title = (data[0].textContent || data[0].content).trim();
+        return new Manga(this, id, title);
     }
 
     async _getMangas() {
+        try {
+            return await this._getMangasV3();
+        } catch (error) {
+            return this._getMangasV5();
```
MikeZeDev: Miss "await" on _getMangasV5 call i think.

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -10,43 +9,113 @@ export default class MangaParkEN extends Connector {
...
+                title: manga.data.name
+            };
+        });
+    }
+
     async _getChapters(manga) {
+        try {
+            return await this._getChaptersV3(manga);
+        } catch (error) {
+            return this._getChaptersV5(manga);
```
MikeZeDev: Miss await maybe?

src/web/mjs/connectors/MangaParkEN.mjs
Druzai: Good find! Thanks. Added if block to check array length.

src/web/mjs/connectors/MangaParkEN.mjs
Druzai: Technically these methods can return a promise, because they are called with await. But yes, it's better to return an array. Thanks again. :)

MikeZeDev (approved):
LGTM. That said, Ronny keen eyes are better than mine to see wrong coding practice or errors.

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -10,43 +9,121 @@ export default class MangaParkEN extends Connector {
         super.label = 'MangaPark';
         this.tags = [ 'manga', 'multi-lingual' ];
         this.url = 'https://mangapark.net';
+        this.apiURL = `${this.url}/apo/`;
+
+        this.requestOptions.headers.set('x-origin', this.url);
+        this.requestOptions.headers.set('x-referer', `${this.url}/`);
         this.requestOptions.headers.set('x-cookie', 'set=h=1;');
```
27136773, line 16: Does this work with non URL encoded component (multiple equal signs)?
`'set=h%3D1'`

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -10,43 +9,121 @@ export default class MangaParkEN extends Connector {
...
+        const id = parseInt(uri.pathname.match(/\/(\d+)\/?/)[1]);
+        const title = (data[0].textContent || data[0].content).trim();
+        return new Manga(this, id, title);
     }
 
     async _getMangas() {
+        try {
+            const mangaList = await this._getMangasV3();
+            if (mangaList.length == 0) {
+                throw new Error('Got 0 mangas from site using v3 version! Using v5 version as a fallback!');
```
27136773, line 40: Suggestion to call and return V5 directly instead of using `try catch`, since `catch` will also be performed for other errors (out of memory, no network connection, ...)

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -10,43 +9,121 @@ export default class MangaParkEN extends Connector {
...
+                title: manga.data.name
+            };
+        });
+    }
+
     async _getChapters(manga) {
+        try {
+            const chapterList = await this._getChaptersV3(manga);
+            if (chapterList.length == 0) {
+                throw new Error('Got 0 chapters from site using v3 version! Using v5 version as a fallback!');
```
27136773, line 126: Same as described in comment above.

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -64,8 +141,23 @@ export default class MangaParkEN extends Connector {
             const chapters = [...this.createDOM(data.html).querySelectorAll('div.episode-item > div > a.chapt')].map(element => {
                 const link = this.getRootRelativeOrAbsoluteLink(element, this.url);
                 let lang = link.match(/c[\d.]+-(\w+)-i\d+/)[1];
-                if (lang)
-                    lang = lang.replace(/\D(_)\D/g, (match, g1) => match.replace(g1, '-'));
+                switch (lang) {
```
27136773: Avoid long `switch case` structures and use dictionaries instead
```typescript
const language = {
    'en': 'en-US',
    'pt': 'pt-BR',
}[lang] || 'default';
const title = `{chapter.title} {language}`;
```

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -79,22 +171,112 @@ export default class MangaParkEN extends Connector {
...
+        `;
+        const vars = {
+            "comicId": parseInt(mangaId),
+            "dbStatuss": ["normal"],
+            "haveChapter": true
+        };
+        const data = await this.fetchGraphQL(this.apiURL, 'get_content_comic_sources', gql, vars);
+        return data.get_content_comic_sources.map(source => {
+            let language = source.data.lang;
+            switch (language) {
```
27136773: See comment above

src/web/mjs/connectors/MangaParkEN.mjs
```diff
@@ -79,22 +171,112 @@ export default class MangaParkEN extends Connector {
...
+                        } else if (typeof amWord !== 'undefined' && typeof amPass !== 'undefined' && typeof imgHostLis !== 'undefined' && typeof imgPathLis !== 'undefined') {
+                            const params = JSON.parse(CryptoJS.AES.decrypt(amWord, amPass).toString(CryptoJS.enc.Utf8));
+                            resolve(imgHostLis.map((data, i) => \`\${data}\${imgPathLis[i]}?\${params[i]}\`));
+                        } else {
+                            const images = __NEXT_DATA__.props.pageProps.dehydratedState.queries
+                                .map(query => {
+                                    const { httpLis, wordLis } = query.state.data.data.imageSet;
+                                    return httpLis.map((path, i) => \`\${path}?\${wordLis[i]}\`);
+                                });
+                            resolve([].concat(...images));
```
27136773, line 247: Consider just returning the `images`, no need to spread and concat to empty array.

27136773 (approved):

src/web/mjs/connectors/MangaParkEN.mjs
Druzai: I tested with these strings (versions v3 and v5): `'set=h=1;'`,  `'set=h%3D1'` and without setting this cookie at all. In each case everything worked fine.
Maybe this cookie is no longer needed?

src/web/mjs/connectors/MangaParkEN.mjs
Druzai: I would do that if in my environment I wasn't getting 404 errors when trying to receive manga/chapters via `v3` endpoint.

For example, using code below I can't fetch manga via `v5` endpoint because `this._getMangasV3()` method throws a 404 error:
```js
async _getMangas() {
  const mangaList = await this._getMangasV3();
  if (mangaList.length == 0) {
	  return await this._getMangasV5();
  }
  return mangaList;
}
```
Log:
image

When using VPN `v3` endpoint works, but I think other people can get this problem too.

src/web/mjs/connectors/MangaParkEN.mjs
Druzai: If I simply return images, I return an array of array of strings - `[ [ 'link1', 'link2' ] ]`.

I can take first parsed query after `map` (i.e. array).
But what if there are several queries for pages in some chapter in the future?

27136773: I wasn't aware `images` is a nested array

PR #5601, opened 2023-03-22
fix(connector): Ganma - vacant chapter titles become undefined

yinyanfr:
A fix for the Ganma connector: When a chapter doesn't have a subtitle, it shows and downloads as `undefined`.

Example: [https://ganma.jp/kotori](https://ganma.jp/kotori)

截屏2023-03-22 21 23 02

### DIFF
--- a/src/web/mjs/connectors/Ganma.mjs
+++ b/src/web/mjs/connectors/Ganma.mjs
@@ -7,7 +7,7 @@ export default class Ganma extends Connector {
         super();
         super.id = 'ganma';
         super.label = 'GANMA!';
-        this.tags = [ 'manga', 'japanese' ];
+        this.tags = ['manga', 'japanese'];
         this.url = 'https://ganma.jp';
         this.requestOptions.headers.set('x-from', this.url);
     }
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Ganma.mjs
```diff
@@ -7,7 +7,7 @@ export default class Ganma extends Connector {
         super();
         super.id = 'ganma';
         super.label = 'GANMA!';
-        this.tags = [ 'manga', 'japanese' ];
+        this.tags = ['manga', 'japanese'];
```
MikeZeDev: Unnecessary change should be removed.

src/web/mjs/connectors/Ganma.mjs
```diff
@@ -71,7 +71,7 @@ export default class Ganma extends Connector {
         const uri = new URL('/api/1.0/magazines/web/' + chapter.manga.id, this.url);
         const request = new Request(uri, this.requestOptions);
         let data = await this.fetchJSON(request);
-        data = data.root.items.find(item => item.id === chapter.id ).page;
+        data = data.root.items.find(item => item.id === chapter.id).page;
```
MikeZeDev: Unnecessary change should be removed.

MikeZeDev (changes requested):
LGTM, requires minor changes

MikeZeDev (approved):

PR #5612, opened 2023-03-25
Kakaopage : switch to graphql

MikeZeDev:
Their api changed 🤷

### DIFF
--- a/src/web/mjs/connectors/kakaopage.mjs
+++ b/src/web/mjs/connectors/kakaopage.mjs
@@ -76,18 +76,6 @@ export default class kakaopage extends Connector {
                       startCursor
                       __typename
                     }
-                    selectedSortOption {
-                      id
-                      name
-                      param
-                      __typename
-                    }
-                    sortOptionList {
-                      id
-                      name
-                      param
-                      __typename
-                    }
                     edges {
                       cursor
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/kakaopage.mjs
```diff
@@ -28,52 +28,459 @@ export default class kakaopage extends Connector {
...
+    async _getChaptersFromPage(manga, nextCursor) {
+        const gql = {
+            operationName: 'contentHomeProductList',
+            variables: {
+                boughtOnly: false,
+                seriesId : manga.id,
+                sortType : 'asc',
+                after : nextCursor,
+            },
+            query: `query contentHomeProductList($after: String, $before: String, $first: Int, $last: Int, $seriesId: Long!, $boughtOnly: Boolean, $sortType: String) {
```
27136773, line 61: GQL query can probably be simplified, I'm pretty sure the connector only needs a very small amount of all those fields.

Same for other queries in this file

MikeZeDev: I tried of but its fragments of fragments of fragments :D. Well, i'll see what i can do

27136773 (approved):

PR #5621, opened 2023-03-27
fix(connector): Copymanga

yinyanfr:
General fixes and improvements for Copymanga

- Updated the website URL and the API URL
- Updated the API entries
- Added additional headers to bypass the API authentification
- Added multiple options
  - Website URL ~~& API URL~~
  - Image format
  - Use oversea CDN
- Re-order pages

### DIFF
--- a/src/web/mjs/connectors/Copymanga.mjs
+++ b/src/web/mjs/connectors/Copymanga.mjs
@@ -25,27 +25,23 @@ export default class Copymanga extends Connector {
                 label: 'Website URL',
                 description: 'This website may change their URL.\nThis is the last known URL which can also be manually set by the user.',
                 input: 'text',
-                value: 'https://www.copymanga.site'
+                value: 'https://www.copymanga.site',
             },
             format:  {
                 label: 'Preferred format',
                 description: 'format of images\nwebp \njpg',
                 input: 'select',
                 options: [
-                    { value: "0", name: 'jpg' },
-                    { value: "1", name: 'webp' },
+                    { value: 0, name: 'jpg' },
+                    { value: 1, name: 'webp' },
                 ],
-                value: "0"
+                value: 0,
             },
             useGlobalCDN: {
                 label: 'Use Global CDN',
                 description: 'Requesting from the Global CDN',
-                input: 'select',
-                options: [
-                    { value: "0", name: 'Yes' },
-                    { value: "1", name: 'No' },
-                ],
-                value: "0"
+                input: 'checkbox',
+                value: true, // true for 0, false for 1
             }
         };
         this.requestOptions.headers.delete('accept');
@@ -65,7 +61,7 @@ export default class Copymanga extends Connector {
     updateHeaders() {
         this.requestOptions.headers.set("Referer", this.url);
         this.requestOptions.headers.set("webp", this.config.format.value);
-        this.requestOptions.headers.set("region", this.config.useGlobalCDN.value);
+        this.requestOptions.headers.set("region", this.config.useGlobalCDN.value ? 0 : 1);
     }
 
     async _getMangaFromURI(uri) {
@@ -145,6 +141,16 @@ export default class Copymanga extends Connector {
         const uri = new URL(`/api/v3/comic/${chapter.manga.id}/chapter2/${chapter.id}?platform=3`, this.apiurl);
         const request = new Request(uri, this.requestOptions);
         const data = await this.fetchJSON(request);
-        return data.results.chapter.contents.map(item => item.url);
+        /**
+         * Copymanga now messes up the order of images,
+         * but in the API data, there's a chapter.words array that contains the correct page order.
+         */
+        const imageUrls = data.results.chapter.contents.map(item => item.url);
+        const imageOrder = data.results.chapter.words;
+        const orderedPages = [];
+        imageOrder.forEach((order, index) => {
+            orderedPages[order] = imageUrls[index];
+        });
+        return orderedPages;
     }
 }
--- a/src/web/mjs/connectors/Copymanga.mjs
+++ b/src/web/mjs/connectors/Copymanga.mjs
@@ -1,6 +1,21 @@
 import Connector from '../engine/Connector.mjs';
 import Manga from '../engine/Manga.mjs';
 
+function formatDate() {
+    // Example: 2023.03.27
+    const now = new Date();
+    const year = now.getFullYear();
+    const month = now.getMonth() + 1;
+    const day = now.getDate();
+    return `${year}.${month < 10 ? "0" : ""}${month}.${day}`;
+}
+
+const API_HEADERS = {
+    'User-Agent': '"User-Agent" to "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.124 Safari/537.36 Edg/102.0.1245.44"',
+    'version': formatDate(),
+    "platform": "1",
+};
+
 export default class Copymanga extends Connector {
 
     constructor() {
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
 import Connector from '../engine/Connector.mjs';
 import Manga from '../engine/Manga.mjs';
 
+function formatDate() {
```
27136773, line 4: Can be simplified
`return new Date().toISOString().replaceAll('-', '.').split('T').shift()`

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+    // Example: 2023.03.27
+    const now = new Date();
+    const year = now.getFullYear();
+    const month = now.getMonth() + 1;
+    const day = now.getDate();
+    return `${year}.${month < 10 ? "0" : ""}${month}.${day}`;
```
27136773: The `day` may be a single digit, is this by intention?

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+                label: 'Preferred format',
+                description: 'format of images\nwebp \njpg',
+                input: 'select',
+                options: [
+                    { value: 'webp', name: 'webp' },
+                    { value: 'jpg', name: 'jpg' },
+                ],
+                value: 'webp'
+            },
+            useOverseaCDN: {
```
27136773: Use a more expressive name, depending on the users location, oversea may have a different meaning.

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
 
     constructor() {
         super();
         super.id = 'copymanga';
         super.label = '拷貝漫畫 (Copymanga)';
         this.tags = [ 'manga', 'webtoon', 'chinese' ];
-        this.url = 'https://copymanga.com';
-        this.apiurl = 'https://api.copymanga.com';
+        this.config = {
+            url: {
```
27136773, line 24: Is the URL expected to frequently change?

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+                label: 'Website URL',
+                description: 'This website may change their URL.\nThis is the last known URL which can also be manually set by the user.',
+                input: 'text',
+                value: 'https://www.copymanga.site'
+            },
+            apiurl: {
```
27136773: Is the API URL expected to frequently change?

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+            },
+            format:  {
+                label: 'Preferred format',
+                description: 'format of images\nwebp \njpg',
+                input: 'select',
+                options: [
+                    { value: 'webp', name: 'webp' },
+                    { value: 'jpg', name: 'jpg' },
+                ],
+                value: 'webp'
```
27136773: The `jpeg` format should be default, almost all HakuNeko users are asking to get JPEG instead of WEBP

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+                description: 'This website may change their API URL.\nThis is the last known API URL which can also be manually set by the user.',
+                input: 'text',
+                value: 'https://api.copymanga.org'
+            },
+            format:  {
+                label: 'Preferred format',
+                description: 'format of images\nwebp \njpg',
+                input: 'select',
+                options: [
+                    { value: 'webp', name: 'webp' },
```
27136773: Consider using `0` and `1` as values, so they don't need to be mapped later.

src/web/mjs/connectors/Copymanga.mjs
yinyanfr: Hello @ronny1982 , this is a better solution, but when I replace it in the code, the connector itself disappears.

截屏2023-03-28 12 13 57

截屏2023-03-28 12 13 38

Edit: This one is good. It seems that this version of electron doesn't support `replaceAll`, by using `replace` with RegExp it works again.

src/web/mjs/connectors/Copymanga.mjs
yinyanfr: No it's not intended, my mistake.

src/web/mjs/connectors/Copymanga.mjs
yinyanfr: Is `useGlobalCDN` a better name?

src/web/mjs/connectors/Copymanga.mjs
yinyanfr: Not frequently.

Edit: The mirror site URL for Chinese mainland users may change frequently.

src/web/mjs/connectors/Copymanga.mjs
yinyanfr: Not frequently.

yinyanfr:
My config setters don't work as intended.

For example, the format setter:

```js
    set format(value) {
        if (this.config && value) {
            this.config.format.value = value;
            this.requestOptions.headers.set("webp", value);
            Engine.Settings.save();
        }
    }
```

I added `this.requestOptions.headers.set("webp", value);` hoping that it would update the request header but it didn't.

Edit: I just realized that those are not default class members, are there a method that is called when the configs are updated?

Edit2: It seems that there's not such a method, now headers are updated before each request.

src/web/mjs/connectors/Copymanga.mjs
MikeZeDev: Often or not, dont expect users to go change an API url in settings. Its contributor's work.

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+            this.config.apiurl.value = value;
+            Engine.Settings.save();
+        }
+    }
+
+    get format() {
+        return this.config.format.value;
+    }
+
+    set format(value) {
```
27136773: Setter seems not to be used => remove

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+            this.config.url.value = value;
+            Engine.Settings.save();
+        }
+    }
+
+    get apiurl() {
+        return this.config.apiurl.value;
+    }
+
+    set apiurl(value) {
```
27136773: Setter seems not to be used => remove

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+        this.requestOptions.headers.append("Referer", this.url);
+        this.requestOptions.headers.append("webp", this.format === "webp" ? 1 : 0);
+        this.requestOptions.headers.append("region", this.useOverseaCDN === "yes" ? 0 : 1);
+    }
+
+    get url() {
+        return this.config.url.value;
+    }
+
+    set url(value) {
```
27136773: Setter seems not to be used => remove

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+            this.config.format.value = value;
+            Engine.Settings.save();
+        }
+    }
+
+    get useOverseaCDN() {
+        return this.config.useOverseaCDN.value;
+    }
+
+    set useOverseaCDN(value) {
```
27136773: Setter seems not to be used => remove

src/web/mjs/connectors/Copymanga.mjs
```diff
@@ -1,16 +1,117 @@
...
+        if (this.config && value) {
+            this.config.useOverseaCDN.value = value;
+            Engine.Settings.save();
+        }
+    }
+
+    canHandleURI(uri) {
+        return /copymanga/.test(uri.origin);
```
27136773: The URL is configured to be changed by the user, but this check does not scale very well with changing the URL

MikeZeDev:
Just wondering : For Yes/No value in settings why not using a checkbox?

yinyanfr:
> Just wondering : For Yes/No value in settings why not using a checkbox?

I didn't find the documentation or an example of this, could you please show me an example?

Edit: I think that I find one.

27136773 (approved):

PR #5629, opened 2023-04-02
MangaFox/MangaHere : fix ad removing & cleaning code

MikeZeDev:
* Ad removing wasnt working because fetching image need a referrer.
* Ad removing now use picture hash instead of image dimensions
* Removed useless code
* Works for MangaFox and MangaHere

### DIFF
--- a/src/web/mjs/connectors/MangaFox.mjs
+++ b/src/web/mjs/connectors/MangaFox.mjs
@@ -101,27 +101,24 @@ export default class MangaFox extends Connector {
         let request = new Request(uri, this.requestOptions);
         const pages = await Engine.Request.fetchUI(request, this.script);
 
-        try {
-        //fetch last image data
-            request = new Request(pages.slice('-1'));
-            request.headers.set('x-referer', uri);
-            const response = await fetch(request);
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/MangaFox.mjs
```diff
@@ -94,58 +97,38 @@ export default class MangaFox extends Connector {
...
-            img.src = pages.slice(-1);
-        });
-        //Check if the last image seems to be the usual ad
-        if(lastImage && lastImage.naturalHeight === 563 && lastImage.naturalWidth === 1000) {
-            pages.pop();
+        try {
+        //fetch last image data
+            request = new Request(pages.slice('-1'));
+            request.headers.set('x-referer', uri);
+            const response = await fetch(request);
```
27136773: Similar code as in `this. _handleConnectorURI`, suggestion to use the existing method:
```javascript
const pageList = pages.map(link => this.createConnectorURI({
    url: link,
    referer: this.url
}));

const lastImage = await this. _handleConnectorURI(pageList.slice(-1));
// perform advertisement check and may remove last image: pageList.pop()

return pageList;
```

27136773 (approved):

PR #5711, opened 2023-04-22
Update BacaMangaOrg.mjs

MIRVOUS:
Changed domain & Manga lists

### DIFF
--- a/src/web/mjs/connectors/BacaMangaOrg.mjs
+++ b/src/web/mjs/connectors/BacaMangaOrg.mjs
@@ -1,6 +1,6 @@
 import WordPressMangastream from './templates/WordPressMangastream.mjs';
 
-export default class MangaTale extends WordPressMangastream {
+export default class BacaMangaOrg extends WordPressMangastream {
 
     constructor() {
         super();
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/BacaMangaOrg.mjs
```diff
@@ -1,13 +1,13 @@
 import WordPressMangastream from './templates/WordPressMangastream.mjs';
 
-export default class BacaMangaOrg extends WordPressMangastream {
+export default class MangaTale extends WordPressMangastream {
```
MikeZeDev: I am not fond of changing class name but there is probably no consequence.

MikeZeDev (approved):
Its ok, but gotta wait for confirmation there is no problem.

src/web/mjs/connectors/BacaMangaOrg.mjs
Sheepux: file should have the same name as the class and ID. 
changing the ID will lose the user bookmarks related to it as it would be considered as a new connector.

for me, it's a new connector or you let it like that.

MikeZeDev: Oh yeah i forgot. We already got the case. Old id is kept because changing it causes problems. I guess class must stays the same too, to be matching filename (and not because it causes problems)

PR #5717, opened 2023-04-23
Update EightMuses.mjs

subhajit-001:
Previously, we have to paste only the comics URL from 8muses which contain pictures. After my code changes. it is successfully, fetching all the comics from a album URL.

E.g. - If you paste below URL, it will successfully download All immediate child comics (Only if child comics does not have any more child comics i.e. issues)

https://comics.8muses.com/comics/album/ShadBase-Comics

### DIFF
--- a/src/web/mjs/connectors/EightMuses.mjs
+++ b/src/web/mjs/connectors/EightMuses.mjs
@@ -26,12 +26,14 @@ export default class EightMuses extends Connector {
 
     async _getChapters(manga) {
         let request = new Request(new URL(manga.id, this.url), this.requestOptions);
-        let data = await this.fetchDOM(request, 'div#top-menu div.top-menu-breadcrumb ol li a');
-        return [ {
-            id: manga.id,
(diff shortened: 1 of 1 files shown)


MikeZeDev:
There is a problem. For Hakuneko, fundamentally, https://comics.8muses.com/comics/album/ShadBase-Comics is a list of comics, where each can have chapters. And you treat them as chapters.

If i paste https://comics.8muses.com/comics/album/MilfToon-Comics (similar pattern to your link btw). 

I got "the milftoons" as first "chapter", but its actually a book, with 4 issues, and with your fix we cant access them

image

image

Hakuneko only show the covers with your modification. 

I bet this is why it has been coded to work like it currently works and not like you want it to work. I mean, pasting the direct link to the content. Website tree is not constant, we have to adapt :/

subhajit-001:
Yeah, got your point. It should not be like this. I will be working on it to make it perfect. Meanwhile is there any way that I could apply this code change in my already installed Hakuneko desktop client?

MikeZeDev:
Read docs : it talks about plugin folder.

PR #5741, opened 2023-04-29
Update NetTruyen.mjs

sswizzz:


MikeZeDev (approved):
Approved but seriously, if they change url every damn day its annoying :D
Whats with VN websites?

PR #5763, opened 2023-05-06
KomikTap : Correct domain

MIRVOUS:
Changed Url


Sheepux:
HakuNeko is able to handle domain redirect.
If the connector is working fine I suggest to reject this PR.

MikeZeDev:
Yeah its no use, its already working 
image

If the goal is to circumvent host ban by ID internet provider, this is not hakuneko problem . User must change their DNS server .
IP are too subject to change.

PR #5765, opened 2023-05-06
PojokManga : Correct domain

MIRVOUS:
Changed url


MikeZeDev:
Previous domain is geoblocked, second is Cloudflared meaning this PR is pointless (as CF blocks Hakuneko).
Is the previous domain working of also blocked for ID people?  

We shouldnt change a domain that works for some people to a domain that wont work at all in hakuneko because its behind cloudflare anti bot.

PR #5766, opened 2023-05-06
SheaManga : Correct Domain & Tag

MIRVOUS:
Changes Url


Sheepux:
HakuNeko is able to handle domain redirect.
If the connector is working fine I suggest to reject this PR.

MikeZeDev:
Agreed. Its not sane to put ip + its up to the user to change its DNS in case of ban by a country.

And on top of this : 

image

MikeZeDev:
Chapters are broken on the website itself, Maybe because of the host/ip change.

image

I vote to reject this PR. We have to wait for when their fix the website itself, then we can see what to do.

* If they fix the redirect for pictures and it works > no need for PR.
* If they find another host > worth a PR
* If they fix but keep an IP only > hew thats ugly , idk

PR #5776, opened 2023-05-10
modified viz to fetch new viz manga simulpubs

amit34521:


src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -157,9 +176,13 @@ export default class VizShonenJump extends Connector {
     }
 
     async _getChapters(manga) {
+		console.log(manga.id);
```
MikeZeDev: Remove console output  :) (should fix the indent error too)

amit34521: removed. totally forgot about that :P

MikeZeDev:
Gotta fire the vpn and test. Should be ok since its just extending what is already done.

MikeZeDev:
If you fixed paid chapters with https://github.com/manga-download/hakuneko/pull/5776/commits/a30fad247210629999f221ff2db5ba7c9f419d43 maybe its worth including the issue in main PR message as 'Fixes <issueURL>' :)

MikeZeDev:
Quite a complex connector, i am pretty sure i am not skilled enough to approve with the last modifications right now ^^
Gotta take time for that.

amit34521:
closest i can say is get the subscription and try it out. atleast that's what i did

src/web/mjs/connectors/VizShonenJump.mjs
```diff
@@ -156,10 +176,57 @@ export default class VizShonenJump extends Connector {
...
+        const request = new Request(new URL(manga.id, this.url), this.requestOptions);
+        let data = await this.fetchDOM(request);
+
+        if ( data.innerText.includes('MANGA Plus by SHUEISHA') ) {
+            alert('This website is geolocked. It can only be accessed from the USA.\nYou may use MANGA Plus instead.', this.label, 'info');
+            return;
+        }
+
+        return [...data.querySelectorAll('div > a.o_chapter-container[data-target-url], tr.o_chapter td.ch-num-list-spacing a.o_chapter-container[data-target-url]')]
+            .filter(element => {
```
MikeZeDev, line 190: About that filter , test try https://www.viz.com/vizmanga/chapters/fushigi-yugi-genbu-kaiden.

Chapter 1 is free, chapter 2 is free. Its ok
For chapter three i have ""javascript:void('join to read');" cause i am not a subscriber and not even logged.

Since its not what you are testing ((/javascript:tryReadChapter/) it end up being displayed while it shouldnt .

Logged in, it stays the same.

MikeZeDev (changes requested):
Clarification / fixing a filter maybe?

src/web/mjs/connectors/VizShonenJump.mjs
amit34521: fixed the condition. for subscribed logged in members, they always see tryReadChapters so removed that.

amit34521:
test now and see if this is correct. i never tried logging out so didnt come through the issue.

MikeZeDev (approved):

PR #5796, opened 2023-05-14
Yurism : no more reader

MikeZeDev:
Their reader is on DynastyScans now.

* Directing the user to DynastyScans connector
* Connector marked for removal in relevant issue

### DIFF
--- a/src/web/mjs/connectors/YuriISM.mjs
+++ b/src/web/mjs/connectors/YuriISM.mjs
@@ -12,14 +12,14 @@ export default class YuriISM extends FoolSlide {
     }
 
     async getMangas() {
-        throw new Error('There is no mangas here, please use DynastyScans connector.');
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/YuriISM.mjs
```diff
@@ -1,20 +1,25 @@
...
-        this.url = 'https://www.yuri-ism.net';
-        this.path = '/slide/directory/';
+        this.url = 'https://www.yuri-ism.com';
         this.language = 'english';
     }
-}
\ No newline at end of file
+
+    async getMangas() {
+        throw new Error('There is no mangas here, please use DynastyScans connector.');
```
Sheepux: How about

> This website doesn't host anything. Use DynastyScans connector instead

Sheepux (approved):

PR #5819, opened 2023-05-21
[MangaHub] Don't include `reloadKey` parameter

### LINKED ISSUE #5805: [MangaHub.io] Connector not working
Yes

**Did you [try Nightly Build](https://github.com/manga-download/hakuneko/releases) ?**
Yes

Yes

**Describe the bug**

**To Reproduce**
1. Name of connector : MangaHub.io
2. Name of manga (if connector update is working) : All Manga
3. Name of chapter (if manga update is working) : All Chapters
4. Copy any relevant error message that appears in a popup or (advanced) within the console (press F12) : MangaHub errors: API rate limit excessed! Go to mangahub.io to continue reading!

**Screenshots**
Screenshot 2023-05-18 at 08 15 20

 - Version [8.3.4]

Druzai:
Resolves #5805

Now they're checking for parameter `reloadKey` not to be in query and return the same expired key. (request headers don't matter) Everything works if this parameter is not specified.

I added a check to automatically try sending request with and without this parameter to see which option is the right one. (because they may re-enable the requirement to request it)
If check found the right state, then that state's value stored in field `this.useReloadKeyParam`.

This commit (fb6bb20c9b0cf64626349f5b8307afae78a9d994) may be redundant, but they might give not the same expired API key in the future...
So it can be included as well. Otherwise I can revert it.

### DIFF
--- a/src/web/mjs/connectors/MangaHub.mjs
+++ b/src/web/mjs/connectors/MangaHub.mjs
@@ -18,6 +18,8 @@ export default class MangaHub extends Connector {
         this.requestOptions.headers.set('x-origin', this.url);
         this.requestOptions.headers.set('x-referer', `${this.url}/`);
         this.requestOptions.headers.set('Accept-Language', 'en-US,en;q=0.9');
+
+        this.useReloadKeyParam = false;
     }
 
     async _initializeConnector() {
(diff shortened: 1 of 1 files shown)


MikeZeDev (approved):
LGTM. They are troublesome arent they?

Druzai:
Yea, pretty much. :)
But as long as they don't use tools like `devtools-detector`, it's all right.

PR #5843, opened 2023-05-27
Add Cuutruyen.mjs

Khangry:
Fixes https://github.com/manga-download/hakuneko/issues/5226

PR #5858, opened 2023-05-30
CuuTruyen

### LINKED ISSUE #5226: [Site Request] cuutruyen
**Name of the website**
cuutruyen.net

- Site: https://cuutruyen.net/
- Manga List: https://cuutruyen.net/newest?page=
- Manga example: https://cuutruyen.net/mangas/204
- Chapter online viewer example: https://cuutruyen.net/mangas/204/chapters/1686

**Languages**
Vietnamese

beer-psi:
Supersedes #5843
Closes #5226


MikeZeDev:
I am sorry Ronny but i must ask for your review, thats hard

src/web/mjs/connectors/Cuutruyen.mjs
```diff
@@ -0,0 +1,117 @@
...
+export default class Cuutruyen extends Connector {
+    constructor() {
+        super();
+        super.id = 'cuutruyen';
+        super.label = 'Cứu Truyện';
+        this.tags = ['manga', 'vietnamese'];
+        this.url = 'https://cuutruyen.net';
+        this.api = 'https://kakarot.cuutruyen.net';
+
+        this.wasm = init(fetch("https://cuutruyen.net/d660d30a7af3c1ad.module.wasm"));
```
27136773, line 14: I would have some security concerns, loading the `wasm` code directly into HakuNeko's render process which may be used for injection attacks in case the website owners want to attack HakuNeko users.

src/web/mjs/connectors/Cuutruyen.mjs
```diff
@@ -0,0 +1,117 @@
...
+    async _descramble(image, scrambleString) {
+        await this.wasm;
+
+        let canvas = document.createElement('canvas');
+        canvas.width = image.width;
+        canvas.height = image.height;
+        const ctx = canvas.getContext('2d');
+        ctx.textAlign = "right";
+
+        render_image(image, ctx, scrambleString);
```
27136773, line 105: The website's descrambling doesn't look to difficult. Isn't it possible to reverse engineer their algorithm and just implement it in a single function instead of all these `wasm` overhead?

src/web/mjs/connectors/Cuutruyen.mjs
MikeZeDev: Yeah thats the main problem. If wasm capabilities is just doing calculus (no internet, no file access ? ) the only thing they can do is wasting cpu time. Which is already a problem.

src/web/mjs/connectors/Cuutruyen.mjs
MikeZeDev: I know a lot of things about reversing PE files but nothing about Wasm :D

The other solution is to use wasm to js "converters" and that a security concern  : you have to trust the JS output

beer-psi: > The website's descrambling doesn't look to difficult. Isn't it possible to reverse engineer their algorithm and just implement it in a single function instead of all these `wasm` overhead?

I have tried for months to no avail; there is simply no good tooling for reverse engineering WebAssembly.

beer-psi: > I know a lot of things about reversing PE files but nothing about Wasm :D
> 
> The other solution is to use wasm to js "converters" and that a security concern : you have to trust the JS output

There's `wasm2js` from [`binaryen`](https://github.com/WebAssembly/binaryen), a WASM compiler toolchain officially maintained by the WebAssembly working group.

src/web/mjs/connectors/Cuutruyen.mjs
beer-psi: > I would have some security concerns, loading the `wasm` code directly into HakuNeko's render process which may be used for injection attacks in case the website owners want to attack HakuNeko users.

I can place the file inside the source tree instead, but I have no idea where I should place it.

MikeZeDev: Sorry about the late reply but if the internal browser is able to load that WASM shouldnt hakuneko be able to inject a javascript script that calls the WASM function? Like i did in https://github.com/manga-download/hakuneko/pull/6528 ? 

This is how its done in Haruneko so i think it could be ok in Haku.

PR #5896, opened 2023-06-13
FIX: LeviatanScans - add WpMangaProtector bypass

MikeZeDev:


Sheepux (approved):
/LGTM
but i don't really think that having a second function is helpfull 

I just did that on a whim 🤷

PR #5920, opened 2023-06-19
Temporary update nettruyen url

KNN-07:
istg they change their domain so often


MikeZeDev:
Its useless, there is an update pending that will let the user chose url

KNN-07:
> Its useless, there is an update pending that will let the user chose url

Alr actually user can change it them self by editing the mjs but ur update make it much more user-friendly

MikeZeDev:
We have other things to do than update the url of the same website every weeks :/

Issue #5615, opened 2023-03-25
[SushiScan] Connector not working


Euphillya:
**Did you read [the troubleshooting guide](https://hakuneko.download/docs/troubleshoot/)**
Yes

**Did you [try Nightly Build](https://github.com/manga-download/hakuneko/releases) ?**
No
**Is the website of the connector working properly / are you able to see the manga within your browser**
Yes

**Describe the bug**
When we load the Sushiscan connector, we have an error : Failed to load resource : the server responded with a status of 403 ()
When we look in the network, during the update with the url : https://sushiscan.net/manga/list-mode/, we see this message : Enable JavaScript and cookies to continue

**To Reproduce**
1. Sushi Scans
2. all
3. all
4. Failed to load resource : the server responded with a status of 403 ()
https://sushiscan.net/manga/list-mode/ -> Enable JavaScript and cookies to continue

**Screenshots**
![image](https://user-images.githubusercontent.com/26461373/227732120-abaf7363-c568-484f-b543-6216d99ac8ca.png)

**HakuNeko (please complete the following information):**
 - Version [e.g. 6.17] rev master@772101

MikeZeDev:
Cant fix; Cloudflare. Wait if they remove it or change website

MikeZeDev:
Tried a few things. Can get one chapter then get fu*** by Cloudflare and error 403 again. Thats really weird

MikeZeDev:
No more cloudflare

MikeZeDev:
Cloudflare again

Zanguy:
hey ! Again not working, 
![image](https://github.com/user-attachments/assets/e5d14a3c-27e9-4fa7-bfe8-7b91e75fc896)
Seems like it's wrong URL 
![image](https://github.com/user-attachments/assets/106bd554-2c60-42a9-ad3d-2bc85746f6f3)

MikeZeDev:
wont fix until they remove cloudflare, its pointless

MikeZeDev:
USE haruneko

PR #5939, opened 2023-06-23
FIX : SushiScans : cloudflare workaround

MikeZeDev:
Somehow, using the updated HeaderGenerator (https://github.com/manga-download/hakuneko/pull/5938) and  setting the useragent and stuff give good results, like being able to download an an entire volume, or update manga list. 

We still got errors 403 but not like before, its still an improvement

image

OFC this needs https://github.com/manga-download/hakuneko/pull/5938 to work properly

See https://github.com/manga-download/hakuneko/issues/5615


Chikimaro:
image

Sheepux (approved):

PR #5940, opened 2023-06-23
Workaround for Manga Livre (1)

Rodrigo-017:


MikeZeDev:
What are you even trying to do? 
That code is incorrect and wont "correct" anything.

PR #5941, opened 2023-06-23
Workaround for Manga Livre (2)

Rodrigo-017:


MikeZeDev:
No. If you want to correct a `Mangalivre `error you dont modify `Chapter.mjs` as its modifying for every chapters of every websites.

You modify Mangalivre.mjs or the particular template if the error is in the template.

MikeZeDev (changes requested):
Change Mangalivre, not Chapter.mjs

MikeZeDev:
Rotting PR.  I may close it soon if no reaction from contributor.

Issue #4972, opened 2022-09-14
Potentially dead/redirected websites (March 12 2024)


MikeZeDev:
If you ever find a dead website, feel free to report it on this issue.

* Please before doing so make sure its dead for a while, make sure its not just blocked by your ISP or geoblocked.
* Also try to search for alternative addresses as website may change URL without saying.
* In case of website still up but without reader (they went to use Mangadex) make sure to mention it.
* Same for defunct website redirecting to another (supported) manga website.

### DEAD WEBSITES
=========================================================
Ace scans
ahstudios.net
aiinscan
akaiyuhimun
akumanotenshi
alkyonescans
Aloalivn
Alphascans : disbanded
amascan
ancientempirescan
angelofresistance
Animangaes : ad webpage
animecenterscan
AniRune
Anitationarts
anivizyon
anubisscans
apoloteam
arangscans.com : ad webpage
artemisnofansub
asgard1team
assortedscans.com : bad gateway
BacaKomikID
BananaMecanica
BatoScan
bestmanhua
betafox
Biamam Scans
Birmanga
BlackArmy
BlackDragonsFansub
blmanhwaclub
Buka
Bx117
caisemh
cat-translator
CatManga
chiotaku
ComicClear
ComicDom
ComicFX
comico-novel
comiconline
cronosscan
cutiepie
darkskyprojects
darkyuerealm
daum
Deliscan : tachiyomiorg/tachiyomi-extensions#7624
DetectiveConanAR
diabolickscan
DOUJINLAND
doujinmitai
drkmangas
dropescan 'sensainayuri
duniakomik
eclypsescan
edelgardescans
einherjarscans
eromanhwas
eternalscans
evascans
exhentai
fdmscan
FenixScan
fenixscanlator
finalscans
fizmanga
FlameScan.fr
ForgottenScans
Frscans.ws
fudidoscan
Fukushuu no yuusha : parking page
Funbe
Funlist
FurioScan : tachiyomiorg/tachiyomi-extensions#17738
Fusion scanlation : #5813
gabutscans
geassscan/hentai
geceninlordu
genkan
getmanhwa.co
GloryScans
GloryScansHentai
GloryScansTR
goddessmanga
godsrealmscan
gogomanga
golden-manga (arabic)
grafimanga
GrazeScans
guoman8
gurukomik
hakaiscan
happy tea scans
HastaTeamDDT
Hatigarm Scans
hatsukimanga
hellunatoon
HentaiMimi
[...]

PR #5944, opened 2023-06-24
LegacyScans: change to Mangastream

TristanWasTaken:
[Legacyscans](https://legacy-scans.com) temporarily rebranded to [flamescans.fr](https://mangadex.org/group/0f82ec30-3b70-4870-b57c-5ebda3f5c770) but are now back with flamescans mangastream theme.

I can't test it though since they're using cloudflare. But I copied the code from the flamescans.fr connector, so everything *should* work out of the box.

### DIFF
--- a/src/web/mjs/connectors/LegacyScans.mjs
+++ b/src/web/mjs/connectors/LegacyScans.mjs
@@ -1,6 +1,6 @@
-import WordPressMadara from './templates/WordPressMadara.mjs';
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
 
-export default class LegacyScans extends WordPressMadara {
+export default class LegacyScans extends WordPressMangastream {
 
     constructor() {
         super();
@@ -8,5 +8,9 @@ export default class LegacyScans extends WordPressMadara {
         super.label = 'Legacy-Scans';
         this.tags = ['webtoon', 'french'];
         this.url = 'https://legacy-scans.com';
+        this.path = '/manga/list-mode/';
+
+        this.queryMangas = 'div.postbody div.soralist ul li a.series';
+        this.queryChapters = 'div#chapterlist ul li a';
     }
 }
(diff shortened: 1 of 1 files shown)


TristanWasTaken:
also #4972 would have to be updated
> legacyscans : moved to flamescanfr (which is already a connector)

MikeZeDev:
Sorry i ended up editing your message instead if answering :/
But yeah, i was mentioning https://github.com/manga-download/hakuneko/issues/4972

MikeZeDev:
Tried it. 

* no need for this.queryMangas.
* And i cant get pictures, i ma checking why

MikeZeDev:
Suggestions  : 

* Remove this.queryMangas
* add ```  this.requestOptions.headers.set('x-referer', this.url); ``` to the constructor;
* override _getpages so the referer we added is effective

``` 
    async _getPages(chapter) {
        return (await super._getPages(chapter)).map(page => this.createConnectorURI(page));
    }

```

MikeZeDev (changes requested):
Suggestions  : 

* Remove this.queryMangas
* add ```  this.requestOptions.headers.set('x-referer', this.url); ``` to the constructor;
* override _getpages so the referer we added is effective

``` 
    async _getPages(chapter) {
        return (await super._getPages(chapter)).map(page => this.createConnectorURI(page));
    }

```

27136773 (approved):

MikeZeDev:
@ronny1982 I was able to test this. And i think i was not wrong asking for modifications, ie if you manage to bypass Cloudflare you still need the referer for pictures. Thanks for the approving boss, but i think its a bit hasty .

Should i not care since its CF plagued anyway?

TristanWasTaken:
@MikeZeDev I implemented the requested changes.
Though how does the Cloudflare bypass work? I still couldn't test the connector :/

MikeZeDev (approved):

MikeZeDev:
> @MikeZeDev I implemented the requested changes. Though how does the Cloudflare bypass work? I still couldn't test the connector :/

Changing the user agent for a particular one. However, we cant do that publicly as the agent will most likely be blacklisted, and its only giving mixed results (its not really reliable).

Thanks for the fix.

TristanWasTaken:
> Changing the user agent for a particular one. However, we cant do that publicly as the agent will most likely be blacklisted, and its only giving mixed results (its not really reliable).

so pretty much what I told you weeks ago that it works when I change the ua with a mitm proxy :smile:

PR #5947, opened 2023-06-26
Sanitize folder names for Linux compatibility

YurDavid:


src/web/mjs/engine/Storage.mjs
```diff
@@ -779,18 +779,20 @@ export default class Storage {
...
     sanatizePath(path) {
         if (this.platform.indexOf('win') === 0) {
-            // TODO: max. 260 characters per path
+            // Replace forbidden characters for Windows
             path = path.replace(/[\\/:*?"<>|\r\n\t]/g, '');
         }
         if (this.platform.indexOf('linux') === 0) {
-            path = path.replace(/[/\r\n\t]/g, '');
+            // Replace forbidden characters for Linux and replace ":" with "_"
+            path = path.replace(/[/\r\n\t]/g, '').replace(/:/g, '_');
```
27136773, line 787: To the best of my knowledge, colons are supported in linux file names, which distro does not work?
Replace colon with [modifier letter colon](https://github.com/manga-download/haruneko/blob/6f4c06ef9f1d56cdefe2ee76adb30fe8d48bd3de/web/src/engine/StorageController.ts#L39C23-L39C57) for forward compatibility with next version of HakuNeko.

src/web/mjs/engine/Storage.mjs
```diff
@@ -779,18 +779,20 @@ export default class Storage {
      */
     sanatizePath(path) {
         if (this.platform.indexOf('win') === 0) {
-            // TODO: max. 260 characters per path
+            // Replace forbidden characters for Windows
```
27136773, line 782: Comment is redundant and can be removed, it can easily be inferred from the code itself.
Other places as well in this file.

src/web/mjs/engine/Storage.mjs
```diff
@@ -779,18 +779,20 @@ export default class Storage {
      */
     sanatizePath(path) {
         if (this.platform.indexOf('win') === 0) {
-            // TODO: max. 260 characters per path
```
27136773, line 782: The `TODO` is not resolved with this pull request, so it should be kept.
Other places as well in this file.

src/web/mjs/engine/Storage.mjs
YurDavid: sure, would  make modifications using the [modifier letter colon](https://github.com/manga-download/haruneko/blob/6f4c06ef9f1d56cdefe2ee76adb30fe8d48bd3de/web/src/engine/StorageController.ts#L39C23-L39C57)
Reason for the PR:
A dualboot user (ubuntu and windows), so this was a problem that arise for me and may for other potential dual boot users

src/web/mjs/engine/Storage.mjs
YurDavid: Got it

src/web/mjs/engine/Storage.mjs
YurDavid: Sure

src/web/mjs/engine/Storage.mjs
YurDavid: also i wanted to ask, how can the bookmark   get the filename if it has been sanitized ?,

PR #5959, opened 2023-07-01
FIX : Comic boost : handle encrypted Publus version

### LINKED ISSUE #5391: {comicブースト] Connector not working
Yes

Yes

**Describe the bug**
I tried refreshing after picking the comicブースト in the connector, but it didn't work.

**To Reproduce**
1. comicブースト

**Screenshots**
image

 - Version [e.g. 6.17] 6.1.7@554b0c

MikeZeDev:
* JSON is encrypted.  I use bookloader to do the job itself :D
* filename has to be calculated by the system. 
* "blocks" computing procedure has changed its "b0Q" now

I wasnt able to fully unobfuscate the "blocks computing" reliably, leading to random unscrambling errors.
So this code just let PUBLUS calculate the blocks itself. 

Fixes https://github.com/manga-download/hakuneko/issues/5391

### DIFF
--- a/src/web/mjs/connectors/ComicBoost.mjs
+++ b/src/web/mjs/connectors/ComicBoost.mjs
@@ -58,109 +58,118 @@ export default class ComicBoost extends Publus {
 
     async _getPages(chapter) {
 
-        //FIRST : get real viewer location
-        let uri = new URL( chapter.id, this.url );
-        let request = new Request( uri.href, this.requestOptions );
-        uri = await Engine.Request.fetchUI(request, `window.location.href`, 5000);
-
         const script = `
-            new Promise((resolve) => {
-            	  const starttime = Date.now();
-                setTimeout(async () => {
-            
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/ComicBoost.mjs
```diff
@@ -13,32 +13,169 @@ export default class ComicBoost extends Publus {
...
-        });
+        }).filter(el => !el.id.includes('?coin=')); //exclude not accessible chapters
+    }
+
+    async _getPages(chapter) {
+
+        const script = `
+            new Promise((resolve, reject) => {
+                setTimeout(async () => {
+
```
27136773, line 63: This method looks quite big, recommend to wrap this whole method body into a `try {...} catch(error) {  reject(error); }` block statement to ensure the promise correctly delivers an error (otherwise the error is ignored and the promise never finishes correctly)

27136773 (approved):

PR #5997, opened 2023-07-11
FIX: Mangadex: Throttle API requests for loading manga list

ccx2:
Lack of throttling leads to connection refused errors. In tests throttling needed to be larger than 2 000 ms (I used the maximum of 10 000 ms).


MikeZeDev:
That is indeed a mistake. Testing it right now. It will take a while tho.

Eww still got error 400. Ok thats not fixing the problem, but its a bug indeed. Approved

MikeZeDev (approved):

PR #6013, opened 2023-07-15
FIX : Siyahmelek : filter dummy picture

MikeZeDev:


MikeZeDev:
If you are wondering why all that code, its because base code use createConnectorURI, so we have to decode url to actually apply a filter.

27136773 (approved):

PR #6023, opened 2023-07-16
CxC: create connector

CetaceanNation:
Connector for https://cxc.today/.

Works with login credentials to access age-restricted comics, handles token refreshes and image decryption. Theoretically should allow for downloading paid content if your account is authorized, but I have not tested this.

JavaScript is not a language I use often, so any feedback on behavior or function choice is appreciated.


src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,250 @@
...
+            accessToken: null,
+            refreshToken: null,
+            expireTime: null
+        };
+        this.requestOptions.headers.set('accept', 'application/json, text/plain, */*');
+        this.requestOptions.headers.set('content-type', 'application/json;charset=utf-8');
+        this.requestOptions.headers.set('referer', this.url);
+        this.requestOptions.headers.set('origin', this.url + '/');
+        this.requestOptions.headers.set('device', 'web');
+        this.requestOptions.headers.set('uuid', this.uuid);
```
MikeZeDev: not sure this uuid value will be used ^^

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,250 @@
...
+        if (bookData.data.work_category.id != 1) {
+            throw Error('CxC book type must be comic (漫畫).');
+        }
+        var title = bookData.data.name;
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        await this._refreshAccessToken();
+        var books = [];
```
MikeZeDev: General hint : "var" is bad.  Use "let " if variable has to be re-affected or "const" if not. That goes for the whole file

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,250 @@
...
+                input: 'password',
+                value: ''
+            }
+        };
+    }
+
+    async _initializeConnector() {
+        await this._getUUID();
+        if (this.config.username.value && this.config.password.value) {
+            await this._login();
```
MikeZeDev: Login in _initializeConnector ok but.

Imagine if someone fill credentials after first connector use. _initializeConnector wont be called at allso the person wont get logged :/

Maybe you should just replace most _refreshAccessToken calls by login(), since login only refreshed tokens if you are already logged :)

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,250 @@
...
+        var title = bookData.data.name;
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        await this._refreshAccessToken();
+        var books = [];
+        var page = 1;
+        var perPage = 50;
+        var pageLength = 0;
```
MikeZeDev: Maybe put those constants as class properties ? not a big deal tho.

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,250 @@
...
+        do {
+            let request = new Request(`${this.api}/book?page=${page}&per_page=${perPage}&sort_by=updated_at&work_category=1&lang=&word_count=0`, this.requestOptions);
+            let bookData = await this.fetchJSON(request);
+            pageLength = bookData.data.data.length;
+            bookData.data.data.forEach(function(book) {
```
MikeZeDev: Maybe make it more "hakuneko coding style" like ? 

```
    async _getMangas() {
        let mangaList = [];
        }
        return mangaList;
    }

        await this._refreshAccessToken();
        const request = new Request(`${this.api}/book?page=${page}&per_page=${perPage}&sort_by=updated_at&work_category=1&lang=&word_count=0`, this.requestOptions);
        let bookData = await this.fetchJSON(request);
        bookData = bookData.filter(book => book.work_category.id == 1);
        return bookData.map(book => {
       	return {
                id: `/store/${book.store.id}/book/${book.id}`,
                title: book.name
            };
        });
    }

```

In case last page+1 doesnt return empty result but throw an error you should just catch it and return en empty array in _getMangasFromPage. No need to handle page count ^^

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,250 @@
...
+        try {
+            chapterData = await this.fetchJSON(request);
+        } catch (error) {
+            if (!this.auth.accessToken) {
+                throw Error("Could not get chapters, may be age-restricted (18禁). Add an account to view chapters.");
+            } else {
+                throw Error("Could not get chapters, your account may not be configured to view age-restricted works (18禁).");
+            }
+        }
+        var chapters = [];
```
MikeZeDev: For code consistency, maybe make it more Hakuneko like , with array.map => ?

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,250 @@
...
+    async _getPages(chapter) {
+        await this._refreshAccessToken();
+        let request = new Request(`${this.api}${chapter.id}`, this.requestOptions);
+        let contentData;
+        try {
+            contentData = await this.fetchJSON(request);
+        } catch (error) {
+            throw Error("Could not get pages, chapter may require a fan subscription (粉絲免費).");
+        }
+        var imageData = [];
```
MikeZeDev: Same remark about the map syntax.

CetaceanNation:
Fixed the paging code for real this time after reading the updated comment, thanks for your feedback.

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,254 @@
...
+        this.url = 'https://cxc.today';
+        this.api = 'https://api.cxc.today';
+        this.booksPerPage = 50;
+        this.uuid = null;
+        this.auth = {
+            accessToken: null,
+            refreshToken: null,
+            expireTime: null
+        };
+        this.requestOptions.headers.set('accept', 'application/json, text/plain, */*');
```
MikeZeDev, line 20: Not sure its a good idea to put json headers in all requests.

Also i think fetchJSON is taking care of that ?

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,254 @@
...
+            return [];
+        }
+    }
+
+    async _getMangas() {
+        await this._login();
+        let books = [];
+        for (let page = 1, run = true; run; page++) {
+            const booksFromPage = await this._getMangasFromPage(page);
+            booksFromPage.length >= this.booksPerPage ? books.push(...booksFromPage) : run = false;
```
MikeZeDev: No. keep the check as `booksFromPage > 0.` In case last page may have less manga that requested they will be lost with your code.

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,254 @@
...
+        try {
+            chapterData = await this.fetchJSON(request);
+        } catch (error) {
+            if (!this.auth.accessToken) {
+                throw Error("Could not get chapters, may be age-restricted (18禁). Add an account to view chapters.");
+            } else {
+                throw Error("Could not get chapters, your account may not be configured to view age-restricted works (18禁).");
+            }
+        }
+        let chapters = [];
```
MikeZeDev: chapters variable is kinda useless. you can directly return the map results (not tested , its a sample)
```

return chapterData.map ((chapter, index) => {
 return {
                 id: `${manga.id}/chapter/${chapter.id+1}`,
                title: `#${chapterNum.toString()} - ${chapter.name}`
          }
});

```

Same stuff elsewhere in the file

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,254 @@
...
+        await this._login();
+        let request = new Request(`${this.api}${chapter.id}`, this.requestOptions);
+        let contentData;
+        try {
+            contentData = await this.fetchJSON(request);
+        } catch (error) {
+            throw Error("Could not get pages, chapter may require a fan subscription (粉絲免費).");
+        }
+        let imageData = [];
+        contentData.data.content.map((page) => {
```
MikeZeDev: Same remark; imageData  variable is useless  :)

src/web/mjs/connectors/CxC.mjs
CetaceanNation: Thanks for the feedback, not used to doing so much logic in return statements.

src/web/mjs/connectors/CxC.mjs
CetaceanNation: Despite using `fetchJSON`, the accept header for requests made is `text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9`, which is not ideal for guaranteeing JSON when the server can provide it. I imagine this accept header works for more scraper-like connectors, but since this is almost exclusively interacting with a JSON API it shouldn't be necessary.

Since [`fetch`](https://github.com/manga-download/hakuneko/pull/6023/files#diff-74dacebfa1b84f4d1772f27f7308bb1e4830b46f348d9bb2487120792807f92dR136) is used for getting the images without using `this.requestOptions`, the accept header is `image/webp,image/apng,image/*,*/*` instead.

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,254 @@
...
+        };
+        let decryptedImageResponse = this._decrypt(encTransformImage, imageCombo).split(",");
+        // from https://cxc.today/worker/base64ToBlob.js
+        const e = decryptedImageResponse[0].indexOf("base64") >= 0 ? atob(decryptedImageResponse[1]) : decodeURI(decryptedImageResponse[1]);
+        let n = new Uint8Array(e.length);
+        for (let i = 0; i < e.length; i++) {
+            n[i] = e.charCodeAt(i);
+        }
+        return {
+            mimeType: `image/${payload.url.split('.').pop()}`,
```
MikeZeDev: what if the image ends with jpg and not jpeg? 

there is a method we use to apply real mimetype from an array buffer

```
               data = {
                    mimeType: response.headers.get('content-type'),
                    data: n
                };
                this._applyRealMime(data);
                return data;
```

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,254 @@
...
+                try {
+                    let loginRequest = new Request(`${this.api}/auth/login`, {
+                        credentials: 'omit',
+                        method: 'POST',
+                        body: JSON.stringify({
+                            grant_type: 'password',
+                            password: this.config.password.value,
+                            username: this.config.username.value
+                        }),
+                        headers: this.requestOptions.headers
```
MikeZeDev, line 205: better put the json headers here individually than for all requests

src/web/mjs/connectors/CxC.mjs
```diff
@@ -0,0 +1,254 @@
...
+        if (this.auth.accessToken && now > this.auth.expireTime) {
+            try {
+                let refreshRequest = new Request(`${this.api}/auth/refresh_token`, {
+                    credentials: 'omit',
+                    method: 'POST',
+                    body: JSON.stringify({
+                        grant_type: 'refresh_token',
+                        refresh_token: this.auth.refreshToken
+                    }),
+                    headers: this.requestOptions.headers
```
MikeZeDev, line 231: better put the json headers here individually than for all requests

src/web/mjs/connectors/CxC.mjs
CetaceanNation: Since all requests besides the one to get the image are for JSON, and not having the JSON headers can result in a 400 for some endpoints, I don't think it makes sense to add it for every request that isn't for the image rather than exclude it for just the image request.

src/web/mjs/connectors/CxC.mjs
MikeZeDev: Yeah you are right, my bad. That said, putting that accept header in all requests is a bit overkill when you have only 2 or 3 times in the code. 

Oh well, do what you want for this.

MikeZeDev:
Fix LINT and its ok

image

MikeZeDev (approved):

27136773:
[Missing in HaruNeko](https://github.com/search?q=repo%3Amanga-download%2Fharuneko%20%27cxc%27&type=code)

PR #6070, opened 2023-07-26
FIX Alphapolis: picture fallback

### LINKED ISSUE #6067: [alphapolis] Cannot download old comic
Yes

**Describe the bug**
Some old comic cannot download.

**To Reproduce**
When download from https://www.alphapolis.co.jp/manga/official/492000109
Except ep 8 and ep9 cannot download

**Expected behavior**
The image url show 403 but the comic can read from chrome.

**Screenshots**
Clipboard02sfdsafds

Clipboard03safdsdfsadf

OS: Windows 11
HakuNeko Version: 6.17@513508

**Additional context**

MikeZeDev:
old manga doesnt have high res

Fixes https://github.com/manga-download/hakuneko/issues/6067

This fix is simple.

### DIFF
--- a/src/web/mjs/connectors/Alphapolis.mjs
+++ b/src/web/mjs/connectors/Alphapolis.mjs
@@ -47,7 +47,11 @@ export default class Alphapolis extends Connector {
         const data = await this.fetchDOM(request, 'viewer-manga-horizontal');
         try {
             const pages = JSON.parse(data[0].getAttribute('v-bind:pages'));
-            return pages.filter(element => typeof element != 'object' && !element.match('white_page')).map(element => element.replace(/\/[0-9]+x[0-9]+.([\w]+)/, '/1080x1536.$1'));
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Alphapolis.mjs
```diff
@@ -63,4 +67,18 @@ export default class Alphapolis extends Connector {
             };
         });
     }
+
+    async _handleConnectorURI(payload) {
+        let request = new Request(payload.hiresPicture, this.requestOptions);
```
Sheepux, line 72: Should test if highres url is not null/empty before sending a request

MikeZeDev: Its always filled. Its a replace so in worst case it would be the same than normalPicture.

Sheepux (approved):

PR #6084, opened 2023-07-30
Update Tempestfansub.mjs

Furkangcmn:
Domain changed


MikeZeDev (commented):
Thanks but it doesnt work, '/manga/list-mode/' gives 404 error.

MikeZeDev (approved):
This one will work

PR #6086, opened 2023-07-31
Update Tempestfansub.mjs

Furkangcmn:
After the domain address changed, there was an error in the listing mode.


MikeZeDev:
Maybe you should stop making random changes hoping it will work ;)
Dont worry I fixed it for you on your own PR and merged it already.

Thanks for the information.

PR #6102, opened 2023-08-02
Save ComicInfo.xml in CBZ Files (Optional)

apiweb:
**Description:**
This pull request introduces a new feature to enhance the settings functionality by adding an option to save a ComicInfo.xml file within the CBZ files. This option is disabled by default to maintain backward compatibility.

**Purpose:**
The ComicInfo.xml file provides valuable additional information about the comics, benefiting users of popular comic readers like ComicRack, ComicZeal, and Chunky. By including this file, the readers can display detailed information about the comics, leading to improved grouping and organization.

image

image

image

### DIFF
--- a/src/web/mjs/HakuNeko.mjs
+++ b/src/web/mjs/HakuNeko.mjs
@@ -8,6 +8,7 @@ import BookmarkManager from './engine/BookmarkManager.mjs';
 import ChaptermarkManager from './engine/ChaptermarkManager.mjs';
 import Connectors from './engine/Connectors.mjs';
 import DownloadManager from './engine/DownloadManager.mjs';
+import ComicInfoGenerator from './engine/ComicInfoGenerator.mjs';
 //import HistoryWorker from './engine/HistoryWorker.mjs'
 import Request from './engine/Request.mjs';
 import Settings from './engine/Settings.mjs';
@@ -32,6 +33,7 @@ export default class HakuNeko {
         this._connectors = new Connectors(ipc);
         this._storage = new Storage();
         this._bookmarkManager = new BookmarkManager(this._settings, new BookmarkImporter());
+        this._comicInfoGenerator = new ComicInfoGenerator();
         this._chaptermarkManager = new ChaptermarkManager(this._settings);
         this._discordPresence = new DiscordPresence(this._settings);
     }
@@ -61,6 +63,10 @@ export default class HakuNeko {
 
     get BookmarkManager() {
         return this._bookmarkManager;
+    }
+
+    get ComicInfoGenerator() {
+        return this._comicInfoGenerator;
     }
 
     get ChaptermarkManager() {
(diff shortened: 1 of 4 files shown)


MikeZeDev:
Oh, an interesting feature, and easy to review. I like that. Gotta wait for others opinions :)
Maybe change the label to "Include Comic File in CBZ" to make it clear for users?

src/web/mjs/engine/ComicInfoGenerator.mjs
```diff
@@ -0,0 +1,29 @@
...
+<ComicInfo xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <Title>${title}</Title>
+    <Series>${series}</Series>
+    <PageCount>${pagesCount}</PageCount>
+</ComicInfo>`;
+    }
+
+    static escapeXML(str) {
+        return str.replace(/[<>&'"]/g, function (c) {
+            switch (c) {
```
27136773: Consider to use a predefined dictionary for mapping characters instead of `switch ... case`.

src/web/mjs/engine/Settings.mjs
```diff
@@ -125,6 +125,13 @@ export default class Settings extends EventTarget {
             value: extensions.img
         };
 
+        this.includesComicFile = {
```
27136773: I suggest to not add a setting for this to prevent setting pollution.
Just always add this sidecar file.

src/web/mjs/engine/Storage.mjs
```diff
@@ -448,7 +449,9 @@ export default class Storage {
             }
             if (Engine.Settings.chapterFormat.value === extensions.cbz) {
                 this._createDirectoryChain(this.path.dirname(output));
-                promise = this._saveChapterPagesCBZ(output, pageData)
+                let mangaName = chapter.manga.title;
+                let chapterName = chapter.title;
+                promise = this._saveChapterPagesCBZ(output, pageData, mangaName, chapterName)
```
27136773: May use `chapter.manga.name` and `chapter.title` directly as parameter instead of intermediate variables `mangaName` and `chapterName`.

src/web/mjs/engine/ComicInfoGenerator.mjs
```diff
@@ -0,0 +1,29 @@
+export default class ComicInfoGenerator {
```
27136773, line 1: This class only contains `static` methods.
Consider to use instance methods, or remove the class and just export the public methods from this module file.

27136773 (approved):

src/web/mjs/engine/ComicInfoGenerator.mjs
apiweb: I fixed the code by changing it from a static method to an instance method, similar to how other classes, like BookmarkManager, work

src/web/mjs/HakuNeko.mjs
```diff
@@ -32,6 +33,7 @@ export default class HakuNeko {
         this._connectors = new Connectors(ipc);
         this._storage = new Storage();
         this._bookmarkManager = new BookmarkManager(this._settings, new BookmarkImporter());
+        this._comicInfoGenerator = new ComicInfoGenerator();
```
27136773, line 36: This instance seems not to be of global significance to be placed in the engine scope, just inject it into the `Storage` constructor, thats the only place where it is really needed.

27136773 (approved):

PR #6104, opened 2023-08-02
Pause, Resume, and Remove Jobs from the Download Queue

apiweb:
- Users can now pause and resume the download queue, and remove jobs from it. 
 - In 'DownloadJob.mjs', the initial status of the chapter has been recorded to reset it when a job is removed. 

The buttons only are enabled when the queue contains at least 1 item.
Screenshot 2023-08-02 at 23 30 48

The buttons to pause, resume and remove items.
Screenshot 2023-08-02 at 23 31 01

When you click on the "Clear Queue" button, a dialog is shown, asking if you are sure that you want to remove all the items.
The chapters will back to their initial status.
Screenshot 2023-08-02 at 23 31 28

Demo

https://github.com/manga-download/hakuneko/assets/5881027/106744a0-b2dc-4a59-b032-76680931d8d3

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/jobs.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/jobs.html
@@ -70,6 +70,24 @@
             .hide {
                 display: none;
             }
+            .buttonDisabled {
+                cursor: default;
+                pointer-events: none;
+                color: gray;
+            }
+            .remove-button {
+                cursor: pointer;
+                text-align: center;
+                padding: 0.25em;
+            }
+            .remove-button i {
+                color: var(--job-list-button-failed-color);
+            }
+
+            .remove-button:hover i {
+                color: var(--job-list-button-completed-color);
+            }
+
         </style>
         <div class$="list [[ getListClass(popupVisibility) ]]">
             <table cellpadding="0">
@@ -79,6 +97,9 @@
                         <td class="cell">[[ item.labels.connector ]]</td>
                         <td class="cell">[[ item.labels.manga ]]</td>
                         <td class="cell">[[ item.labels.chapter ]]</td>
+                        <td class="remove-button" on-click="removeFromQueue">
+                            <i class="fas fa-times"></i>
+                        </td>
                         <td class="progress" style$="background: linear-gradient(90deg, var(--job-list-progress-color) [[ item.progress ]]%, var(--job-list-progress-background-color) 0%);"></td>
                     </tr>
                 </template>
@@ -88,7 +109,18 @@
             <div class="expander">
                 <i class$="fas [[ getButtonClass(popupVisibility) ]] button" title="Toggle download list" on-click="toggleJobList"></i>
             </div>
-            <div class="status">[[ jobList.length ]] Download(s)</div>
+            <div class="status">
+                <span>[[ jobList.length ]] Download(s)</span>
+                <span>
+                    <i class$="fas fa-fw button [[ getPlayPauseClass(queuePaused) ]] [[ isQueueEmpty(jobList.length) ]]"
+                       title$="[[ getPlayPauseTooltip(queuePaused) ]]"
+                       on-click="playPauseQueue">
+                    </i>
+                </span>
+                <span>
+                    <i class$="fas fa-times remove-button [[ isQueueEmpty(jobList.length) ]]" title="Clear Job List" on-click="clearJobList"></i>
+                </span>
+            </div>
         </div>
     </template>
 
@@ -113,6 +145,10 @@
                         notify: true, // enable upward data flow,
                         //readOnly: true, // prevent downward data flow
                         //observer: 'onSelectedMangaChanged'
+                    },
+                    queuePaused: {
+                        type: Boolean,
+                        value: false,
                     }
                 };
             }
@@ -130,6 +166,13 @@
                 // register callbacks for events published by download manager
                 Engine.DownloadManager.addEventListener('updated', this.onDownloadStatusUpdated.bind(this));
                 this.ipc.on( 'close', this.onClose.bind( this ) );
+            }
+
+            /**
+             *
+             */
+            isQueueEmpty(length) {
+                return length < 1 ? 'buttonDisabled' : '';
             }
 
             /**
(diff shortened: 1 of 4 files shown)


src/web/mjs/engine/DownloadManager.mjs
```diff
@@ -10,6 +10,7 @@ export default class DownloadManager extends EventTarget {
         super();
         this.queue = [];
         this.worker = setInterval( this.processQueue.bind( this ), 250 );
+        this.queuePaused = false;
```
27136773, line 13: Personally i prefer not using flags/fields to control behavior, therefore i would clear the interval and unset the `this.worker` (e.g., to undefined) in case of pause and assign a new `this.worker` interval in case of resume.

src/web/mjs/engine/DownloadManager.mjs
```diff
@@ -36,12 +37,53 @@ export default class DownloadManager extends EventTarget {
         }
     }
 
+    /**
+     * Returns true if the download queue is paused.
+     */
+    isQueuePaused() {
```
27136773, line 43: May use a getter instead of a function to expose the paused state

src/web/mjs/engine/DownloadManager.mjs
27136773, line 13: May use a more appropriate name for the private field e.g., `_isQueuePaused `

src/web/mjs/engine/DownloadManager.mjs
```diff
@@ -36,12 +37,53 @@ export default class DownloadManager extends EventTarget {
...
+    pauseQueue() {
+        this.queuePaused = true;
+    }
+
+    /**
+     * Start the download queue processing if it was paused.
+     */
+    startQueue() {
+        this.queuePaused = false;
+        this.processQueue(); // Start processing the queue again
```
27136773, line 59: This call to `processQueue` seems redundant and may be removed as there is a `this.worker` interval calling this method every 250ms.

src/web/mjs/engine/DownloadManager.mjs
```diff
@@ -52,4 +94,18 @@ export default class DownloadManager extends EventTarget {
             }
         }
     }
+
+    /**
+     * Clear the download queue.
+     */
+    clearQueue() {
+        for (let connectorID in this.queue) {
+            const jobs = this.queue[connectorID];
```
27136773, line 103: Can this be simplified by using existing functionality?
`this.queue[connectorID].slice(0).forEach(this.removeFromQueue);`

src/web/mjs/engine/DownloadManager.mjs
27136773, line 1: Be aware that `this.queue` is not thread safe.
Adding more methods to manipulate `this.queue` increases the probability of errors due to inconsistent access (e.g., accessing a queue entry that already has been removed).

src/web/lib/hakuneko/frontend@classic-dark/jobs.html
```diff
@@ -132,6 +168,13 @@
                 this.ipc.on( 'close', this.onClose.bind( this ) );
             }
 
+            /**
+             *
+             */
+            isQueueEmpty(length) {
```
27136773, line 174: The method name is misleading, i would expect a bool as return type, but it returns the string representation of a related CSS class.

src/web/lib/hakuneko/frontend@classic-light/jobs.html
```diff
@@ -178,6 +221,41 @@
...
+                }
+            }
+
+            /**
+             *
+             */
+            playPauseQueue() {
+                if(this.queuePaused) {
+                    Engine.DownloadManager.startQueue();
+                    this.queuePaused = false;
```
27136773, line 252: In case of `startQueue()` failed, the assignment of `false` to `this.queuePaused` is incorrect, use `DownloadManager.isQueuePaused` instead.

```javascript
if(this.queuePaused) {
    Engine.DownloadManager.startQueue();
} else {
    Engine.DownloadManager.pauseQueue();
}
this.queuePaused = Engine.DownloadManager.isQueuePaused;
```

src/web/lib/hakuneko/frontend@classic-light/jobs.html
```diff
@@ -178,6 +221,41 @@
...
+                } else {
+                    Engine.DownloadManager.pauseQueue();
+                    this.queuePaused = true;
```
27136773, line 255: The methods `DownloadManager.pauseQueue()` and `DownloadManager.startQueue()` can be called from anywhere in the software, so `this.queuePaused` may not reflect the correct state.

Consider to introduce an event mechanism to update `this.queuePaused` whenever the pause state of the queue in the `DownloadManager` is changed instead of setting it manually.

src/web/lib/hakuneko/frontend@classic-light/jobs.html
```diff
@@ -197,6 +275,29 @@
...
 
+            /**
+             *
+             */
+            removeFromQueue(e) {
+                const index = e.model.index;
+                const job = this.jobList[index];
+                if (job) {
+                    Engine.DownloadManager.removeFromQueue(job);
+                    this.splice('jobList', index, 1);
```
27136773, line 286: The method `DownloadManager.removeFromQueue()` can be called from anywhere in the software, so `this.jobList` may not reflect the correct entries.

Consider to introduce an event mechanism to update `this.jobList` whenever the queue in the `DownloadManager` is changed instead of setting it manually.

src/web/lib/hakuneko/frontend@classic-light/jobs.html
```diff
@@ -197,6 +275,29 @@
...
+            }
+
+            /**
+             *
+             */
+            async clearJobList() {
+                if(await confirm( 'Do you really want to clear the download queue?' )) {
+                    Engine.DownloadManager.clearQueue();
+                    this.splice('jobList', 0, this.jobList.length);
+                    this.queuePaused = false;
```
27136773, line 297: The queue may still be in paused stated, so assigning `false` may be incorrect.

27136773 (changes requested):
I recommend to address the UI synchronisation issues with `isQueuePaused` and `jobList`.

src/web/mjs/engine/DownloadManager.mjs
Sheepux: i'd also prefer having a "stratus" field instead of "isPaused".
List of status in haruneko: https://github.com/manga-download/haruneko/blob/master/web/src/engine/DownloadTask.ts#L6

27136773:
Stale for almost 2 month => closed
Feel free to re-open or create a new PR in case the feedback will be addressed

PR #6106, opened 2023-08-03
Fixed failed test, due to error message not matching.

omar908:
Fixed failed test, due to error message not matching.

### DIFF
--- a/src/app/__tests__/UpdateServerManager.test.js
+++ b/src/app/__tests__/UpdateServerManager.test.js
@@ -208,11 +208,12 @@ describe('UpdateServerManager', function () {
 
         it('should throw error when URL is invalid', async () => {
             let testee = fixture.createTestee();
+            let invalidURL = 'foobar';
             try {
-                await testee.getUpdateArchive({link: 'foobar'});
+                await testee.getUpdateArchive({link: invalidURL});
                 assert.fail('Expected error not thrown!');
             } catch(error) {
-                assert.equal(error.message, 'Invalid URL: foobar');
+                assert.equal(error.message + ':' + invalidURL, 'Invalid URL:'+ invalidURL);
             }
         });
 


src/app/__tests__/UpdateServerManager.test.js
```diff
@@ -208,11 +208,12 @@ describe('UpdateServerManager', function () {
...
         it('should throw error when URL is invalid', async () => {
             let testee = fixture.createTestee();
+            let invalidURL = 'foobar';
             try {
-                await testee.getUpdateArchive({link: 'foobar'});
+                await testee.getUpdateArchive({link: invalidURL});
                 assert.fail('Expected error not thrown!');
             } catch(error) {
-                assert.equal(error.message, 'Invalid URL: foobar');
+                assert.equal(error.message + ':' + invalidURL, 'Invalid URL:'+ invalidURL);
```
27136773, line 216: This assertion does not look correct to me.
The error message seems to not contain the URL anymore.
The old assertion looks correct to me, so the production code probably needs to be fixed and not the test.

27136773 (changes requested):
Submit feedback that must be addressed before merging.

src/app/__tests__/UpdateServerManager.test.js
omar908: Hi @ronny1982!

The error is coming from node.js internals, not the repo of the application.
image


I just finished double checking on this to be safe.
Changes were made on the Node.js internals to remove the invalidURL from the error message, please see below of the change that was made:
https://github.com/nodejs/node/commit/417c31b69a2ee63649600ece471d9e1a5bba9f55

Thank you for the feedback!

27136773: So this means the test output depends on the NodeJS version on which the test is performed?

In this case, the test should be adjusted to work with any NodeJS version and not just with the latest, or the test should be kept as is and the node version for this project needs to be locked to the same version that is shipped with the HakuNeko Desktop Client (Electron 8.3.4 => NodeJS 12.13.0).

May also affect the [build pipeline](https://github.com/manga-download/hakuneko/blob/master/.azure/pipelines/ci.yml#L37)

27136773: OK, it seems the build agents are installing NodeJS 16.x instead of the required 10.x which explains why the test is breaking now on the pipeline.

image

I suggest to leave the test unchanged, but properly adjust the build pipeline to use NodeJS 12

omar908: Correct, they were 12.x previously. Those changes within the github actions workflow were made here, I THINK:
https://github.com/manga-download/hakuneko/pull/4949/commits/73826d5d9013128b23658727f374af718e74ba44

omar908: I can adjust the test to work with either version, but first we need to decide, do we want to go back to 12.x on the github workflow as well, in that case I would double check to see if the issue even exists once we go back.

omar908: Sorry, I didn't see the bottom of you message, my mistake.
Let me make a separate PR to change the build pipeline and double check the test then.

PR #6151, opened 2023-08-16
Add ignore 404 error option for page downloads

### LINKED ISSUE #5731: [Feature] option to ignore 404 errors
I'm always frustrated when a single image like the credit page is missing and the whole chapter download fails. I will have to get the chapter manually anyways.

An option in the settings to ignore 404 errors and save the chapter anyways.

An "ignore errors" option when you retry to download a failed chapter.

 - Version: 6.1.7

**Additional context**

elena-lorente:
Fixes https://github.com/manga-download/hakuneko/issues/5731.

### DIFF
--- a/src/web/mjs/engine/DownloadJob.mjs
+++ b/src/web/mjs/engine/DownloadJob.mjs
@@ -124,7 +124,7 @@ export default class DownloadJob extends EventTarget {
         for(let page of pages) {
             await this._wait(this.throttle);
             const response = await fetch(page, this.requestOptions);
-            if(response.status !== 200) {
+            if(!(response.status === 200 || response.status === 404 && Engine.Settings.ignore404.value)) {
                 throw new Error(`Page " ${page}" returned status: ${response.status} - ${response.statusText}`);
             }
             result.push(await response.blob());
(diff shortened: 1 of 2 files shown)


TristanWasTaken:
fixes #5731

src/web/mjs/engine/DownloadJob.mjs
```diff
@@ -124,7 +124,7 @@ export default class DownloadJob extends EventTarget {
         for(let page of pages) {
             await this._wait(this.throttle);
             const response = await fetch(page, this.requestOptions);
-            if(response.status !== 200) {
+            if(!(response.status === 200 || response.status === 404 && Engine.Settings.ignore404.value)) {
```
Sheepux: The feature should be "ignore errors" and not just 404.
404 seems to be for a specific connector case
Something like that 
```javascript
         if(response.status !== 200 && !Engine.Settings.ignoreErrorOnDownload.value)) {
```

elena-lorente: Updated it!

Sheepux: don't use not OR, use AND, it's easier to read

elena-lorente: Fixed

Sheepux:
@ronny1982 requesting a tiny review

src/web/mjs/engine/Settings.mjs
```diff
@@ -93,6 +93,13 @@ export default class Settings extends EventTarget {
             value: false
         };
 
+        this.ignoreErrorOnDownload = {
+            label: 'Ignore errors on download',
+            description: 'Download pages ignoring errors',
```
27136773: I recommend to use a more detailed description like `The download will be treated as successful, even if all or some files failed to download and the chapter is in a broken state.`

Otherwise i have a bad feeling that this setting may backfire and users will start complaining about broken downloads.

27136773 (approved):

src/web/mjs/engine/Settings.mjs
MikeZeDev: Changed it myself since OP isnt answering. I hope you dont mind

PR #6212, opened 2023-09-07
template : clarification on issue searching

MikeZeDev:


Sheepux:
It's already written in the description

MikeZeDev:
But i feel like people are reading the checkbox and not the subtitle where its written in a smaller font.

We have less duplicates now but i think this change wont hurt, i got a couple of dupe because people missed the sentence.

PR #6223, opened 2023-09-15
feat(reader): display connecteur manga page in reader

serge-reboul:
I thought it is so helpful to rely on the hakeneko to 'browse' connector web site in reader view when reader is visible + manga is selected + no chapter selected.

As such it is super easy to be done via simple iframe, I created PR to propose this evolution.

Let me know if you feel it may be a good idea and/or if this PR is good enoughor deserved to be worked out further.

Thanks,
S.


serge-reboul:
proposed logic proposed is uncomplete. It does not work with most of connector and would require connectors extension to store manga pages or API call when any
Should rather be dropped as is

PR #6224, opened 2023-09-15
feat(downlaod): one click button to download every visible manga

serge-reboul:
I thought it would be useful to mass download every available chapter for every visible (filtered ) manga.
This helps me a lot to sync locally all my bookmarks or some subset of filtered manga

I don't know if you agree that this could be a valid improvement.
If so, let me know if you feel this PR is enough or if it deserves some evolutions

Thanks,
Serge

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
@@ -94,6 +94,9 @@
                         <input id="paste" />
                         <i class="fas fa-fw fa-paste button" on-click="onPasteClick" title="Click to paste manga links from the clipboard (CTRL + V)"></i>
                     </td>
+                    <td width="1">
+                        <i class="fas fa-fw fa-download button" on-click="onDownloadMangaClick" title="Click to download all chapters currently available for all mangas shown in the filtered and sorted list"></i>
+                    </td>
                 </tr>
             </table>
         </div>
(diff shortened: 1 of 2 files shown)


MikeZeDev:
>The philosophy is ad-hoc consumption, get it when you want to read/watch it. It is not meant to be a mass downloader to stock up thousands of chapters that are just collected and will probably never be read.


Pretty sure this wont be accepted.

Sheepux:
We are against mass downloads.
But i see where this feature could be possibly acceptable : if it's only available with bookmarked mangas. This allows someone who goes on WE to quickly sync his current reading progress.
Need other contributors point of view on this one.

serge-reboul:
Got it and thanks for the feedbacks.
Let me know if it is worth that I restrict this feature to the bookmarks

27136773:
I vote to reject this PR for the following reasons:
- Avoid adding features in HakuNeko as it is supposed to be in end-of-life/maintenance-mode
- It does not comply with HakuNeko philosophy
- It may make the app unresponsive if a user accidentally click this button when the list shows 50k mangas

Sheepux:
Yeah but what if it's restrained to bookmarks. I don't think people will have so much mangas (i currently have 560 and that's quite a lot)...

I'm mostly thinking for something in Haru

MikeZeDev:
Closing this as we have 2 votes against it, mostly because we wont add  features to Hakuneko anymore.
Its time for haruneko.
Thanks for your contribution anyway.

PR #6258, opened 2023-09-22
Fix Tappytoon : get best picture quality if available

### LINKED ISSUE #6257: [Feature] (tappytoon.com)  chapter download in high definition
No, I just want the option to download the raw in high definition to be added, since tapytoon has the option of medium quality than normal, and the high quality which is 1280

have an option to download in normal quality or in high definition

that automatically downloads the high definition and no longer the normal one.

 - HakuNeko: 8.3.4

**Additional context**

Please implement that option...both high definition and normal...

MikeZeDev:
Fixes https://github.com/manga-download/hakuneko/issues/6257

There is a new endpoint to get pictures with chosen quality Tested on EN FR DE, works

### DIFF
--- a/src/web/mjs/connectors/templates/Tappytoon.mjs
+++ b/src/web/mjs/connectors/templates/Tappytoon.mjs
@@ -68,9 +68,15 @@ export default class Tappytoon extends Connector {
     }
 
     async _getPages(chapter) {
-        let uri = new URL(`/chapters/${chapter.id}?includes=images`, this.apiurl);
+        //check if super quality is supported
+        let uri = new URL(`/comics/${chapter.manga.id}`, this.apiurl);
         let request = new Request(uri, this.requestOptions);
         let data = await this.fetchJSON(request);
-        return data.images.map(image => image.url);
+        const quality = data.isSuperHighQualitySupported ? 'super_high': 'high';
+        uri = new URL(`/content-delivery/contents?chapterId=${chapter.id}&variant=${quality}&locale=${this.lang}`, this.apiurl);
+        request = new Request(uri, this.requestOptions);
+        data = await this.fetchJSON(request);
+        return data.media.map(image => image.url);
     }
+
 }


src/web/mjs/connectors/templates/Tappytoon.mjs
```diff
@@ -68,9 +68,15 @@ export default class Tappytoon extends Connector {
...
 
     async _getPages(chapter) {
-        let uri = new URL(`/chapters/${chapter.id}?includes=images`, this.apiurl);
+        //check if super quality is supported
+        let uri = new URL(`/comics/${chapter.manga.id}`, this.apiurl);
         let request = new Request(uri, this.requestOptions);
         let data = await this.fetchJSON(request);
-        return data.images.map(image => image.url);
+        const quality = data.isSuperHighQualitySupported ? 'super_high': 'high';
+        uri = new URL(`/content-delivery/contents?chapterId=${chapter.id}&variant=${quality}&locale=${this.lang}`, this.apiurl);
```
27136773, line 76: It seems `locale` parameter can be omitted since `this.lang` is never set to something other than `undefined`

27136773 (approved):

src/web/mjs/connectors/templates/Tappytoon.mjs
MikeZeDev: ```this.lang``` is always defined in connectors using this template. 

https://github.com/manga-download/hakuneko/blob/e9d2abe4f18f9576ae4270e4889bb8ef1f6a0006/src/web/mjs/connectors/TappytoonDE.mjs#L15

https://github.com/manga-download/hakuneko/blob/e9d2abe4f18f9576ae4270e4889bb8ef1f6a0006/src/web/mjs/connectors/TappytoonEN.mjs#L15

https://github.com/manga-download/hakuneko/blob/e9d2abe4f18f9576ae4270e4889bb8ef1f6a0006/src/web/mjs/connectors/TappytoonFR.mjs#L15


image

PR #6290, opened 2023-10-04
feat(GalaxyAction, GalaxyManga): create & update connector

TristanWasTaken:
https://galaxyaction.site/
https://flixscans.com


MikeZeDev (changes requested):
It redirect to flixscans, and flixscans use an api and its not mangastream

src/web/mjs/connectors/GalaxyManga.mjs
```diff
@@ -1,13 +1,67 @@
...
+
+    async _getChapters(manga) {
+        const uri = new URL(`/api/v1/webtoon/chapters/${manga.id}-desc`, this.api);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchJSON(request);
+
+        return data.map((chapter) => {
+            return {
+                id: chapter.id,
+                title: chapter.title ? `${chapter.name} ${chapter.title.trim()}` : chapter.name
```
27136773: May be shortened to
```javascript
title: `${chapter.name} ${chapter.title || ''}`.trim()
```

27136773 (approved):

TristanWasTaken:
I just wrote with the site operator and he said that multiple scan groups will post to the site.
From what I see in the api manga response, each group would "own" a title like on bato.to
should the group name be attached to the manga name in the format `<manga> [<group>]` or is it fine since it's not multiple groups sharing one manga like on MangaDex?

MikeZeDev (approved):

PR #6298, opened 2023-10-07
Fix LineMangaJP

Elastic1:
Support "portal" chapter.
Manga example: https://manga.line.me/product/periodic?id=S122922

### DIFF
--- a/src/web/mjs/connectors/LineMangaJP.mjs
+++ b/src/web/mjs/connectors/LineMangaJP.mjs
@@ -62,6 +62,10 @@ export default class LineMangaJP extends Connector {
                 });
             });
         }
+        if (option.isPortal) {
+            return Object.values(option.portalPages)
+                .map(pp => this.createConnectorURI({ pp }));
+        }
         if (images.length == 0) {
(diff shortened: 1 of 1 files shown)


MikeZeDev:
Seems good. I'll see if i can find a good proxy to test it

src/web/mjs/connectors/LineMangaJP.mjs
```diff
@@ -62,6 +62,10 @@ export default class LineMangaJP extends Connector {
                 });
             });
         }
+        if (option.isPortal) {
+            return Object.values(option.portalPages)
+                .map(pp => this.createConnectorURI({ pp }));
```
27136773: May use a more expressive variable name instead of `pp`

ronny1982: Also it seems `pp` is not used, only `pp.url` is used, so instead of providing `pp` as parameter, may provide `pp.url` as parameter for connectorURI

27136773 (approved):

PR #6302, opened 2023-10-08
feat(download): add manager and UI for updating bookmarked manga chapters

russaa:
add functionality for checking & queuing new chapters in __Bookmark__ list/connector

> NOTE this feature is similar to the PR #6224 but restricted to the Bookmark list _(plus some other features)_

main features:
 * start / cancel updating bookmarked manga chapters
 * include / exclude bookmarked mangas from update
 * show status / progress when checking for new manga chapters

The start / cancel button (and the progress information) is added at the bottom next to download-list button.

_(also note that _canceling_ in this context means to stop checking for new chapters -- chapters that were already queued for downloading would not be effected)_

The Bookmark list has additional include/exclude buttons for each bookmarked manga (for including/excluding during the update).

The current implementation always waits until `Manga.getChapters()` is finished, before checking the next manga -- this could be changed, e.g. to not waiting at all or only waiting if `Manga.getChapters()` is active for the the same Connector.

### DIFF
--- a/src/web/index.html
+++ b/src/web/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
-<head> 
+<head>
     <meta charset="utf-8">
     <title>HakuNeko</title>
     <link rel="icon" href="./img/logo.png">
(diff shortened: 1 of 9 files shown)


manga-download-owner:
Unfortunately this version of HakuNeko is in feature freeze, only critical website maintenance PR changes will be merged
→ Closed
→ See [manga-download/haruneko](https://github.com/manga-download/haruneko) for future development

PR #6361, opened 2023-10-26
FIX MangaFire : filter api junk

MikeZeDev:
sometimes querying for Volumes returns Chapters.

### DIFF
--- a/src/web/mjs/connectors/MangaFire.mjs
+++ b/src/web/mjs/connectors/MangaFire.mjs
@@ -101,15 +101,16 @@ export default class MangaFire extends Connector {
                 const data = await this.fetchJSON(request);
                 const dom = this.createDOM(data.result.html);
                 const chaptersNodes = [...dom.querySelectorAll('a')];
-                chaptersNodes.map(chapter => {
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/MangaFire.mjs
```diff
@@ -101,15 +101,16 @@ export default class MangaFire extends Connector {
...
                 const chaptersNodes = [...dom.querySelectorAll('a')];
-                chaptersNodes.map(chapter => {
-                    const id = {itemid : chapter.dataset.id, itemtype : type};
-                    const title = chapter.text.trim();
-                    chapterList.push ({
-                        id : JSON.stringify(id),
-                        title : title,
-                        language : language
+                chaptersNodes.filter(anchor=> anchor.pathname.includes(`/${type}-`))
+                    .map(chapter => {
```
27136773: Recommend to use `forEach` instead of `map` to express the intend (it updates `chapterList` by pushing within the callback instead of pushing the result from `map`).

27136773 (approved):

27136773 (approved):

PR #6392, opened 2023-11-04
Fixes AllAnimeto : websites changes

MikeZeDev:
* Change manga website url
* change api url and calls
* Remove StreamSB as source for anime since StreamSB is dead.
* Fixed video url decryption links


marwan7112:
Can you  please change the domain for scanvf "https://www.vfscan.cc/"

PR #6396, opened 2023-11-05
Update Tempestfansub.mjs

Furkangcmn:
has been updated again so that you can read the manga list.


MikeZeDev:
Superseed by https://github.com/manga-download/hakuneko/pull/6402

PR #6426, opened 2023-11-12
Ridibooks Connector

### LINKED ISSUE #5387: [Site Request] Ridi Books
**Name of the website**
Ridi Webtoon

**_Website urls_**
Site: https://ridibooks.com/webtoon/recommendation
Manga List: https://ridibooks.com/comics/ebook
Manga example: https://ridibooks.com/books/2066004870
Chapter online viewer example: https://view.ridibooks.com/books/2066004870

Thanks! Zunshine#5203 on discord if there's anything else I can provide or help out with. This is just a request to connect to this korean webtoon site.

danspinola:
Add a connector for Ridibooks.

Fixes https://github.com/manga-download/hakuneko/issues/5387


src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
+import Connector from "../engine/Connector.mjs";
+import Manga from "../engine/Manga.mjs";
+import Chapter from "../engine/Chapter.mjs";
+
+/**
```
MikeZeDev: May want to remove those comments

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+
+        this.apiUrl = "https://api.ridibooks.com";
+        this.pagesApi = "https://view.ridibooks.com";
+        this.bookUrlPattern = /^https:\/\/ridibooks\.com\/books\/(?<id>[a-zA-Z0-9]+)/;
+    }
+
+    // OVERRIDE
+
+    /**
+     * Get all mangas from the website.
```
MikeZeDev: Again we dont need such comments, even if its good practice

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+     * @returns {Promise<IManga[]>}
+     * @override
+     */
+    async _getMangas() {
+        let mangaList = [];
+
+        // legacy takes some time and makes quite a lot of requests
+        // and still it doesn't return the full catogue
+        let legacyList = await this._getMangasLegacy();
```
MikeZeDev: You could use const instead of let if the variable isnt affected again. Its valable everywhere in the file.

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+        // Webtoons not being one of them;
+        let mangaListGql = await this._getMangasGql();
+        mangaList.push(...mangaListGql);
+
+        return mangaList;
+    }   
+
+    /**
+     * Method to get all chapters for a webtoon.
+     * @param {IManga} manga
```
MikeZeDev: same remark

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+     * @returns {IChapter[]}
+     * @override
+     */
+    async _getChapters(manga) {
+        let uri = new URL(`/books/${manga.id}`, this.url);
+
+        let request = new Request(uri, this.requestOptions);
+
+        return Engine.Request.fetchUI(request, 'seriesBookListJson')
+            .then(data => data.map(item => ({
```
MikeZeDev: We would rather use await when possible and avoid the .then syntax. In that case maybe you can directly do 
```    return await Engine.Request.fetchUI(request, 'seriesBookListJson') ``` since its an array of {id : string , title: string}

File: src/web/mjs/connectors/Ridibooks.mjs [hunk 6]
```diff
@@ -0,0 +1,247 @@
...
+                id: item.id,
+                title: item.title
+            })));
+    }
+
+    /**
+     * Method to get all pages of a chapter.
+     * @param {Chapter} chapter
+     * @returns {Promise<string[]>}
```
MikeZeDev: same remark

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+    async _getPages(chapter) {
+        let uri = new URL(
+            `/generate/${chapter.id}`,
+            this.pagesApi
+        );
+
+        let request = new Request(uri, this.requestOptions);
+
+        return this.fetchJSON(request, 3)
+            .then(res => {
```
MikeZeDev: same remark about .then syntax. Also not sure its required to try 3 times :D

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+                }
+
+                return res.pages.map(page => page.src);
+            });
+    }
+
+    /**
+     * Get the manga from the provided resource.
+     * @param {URL|string} uri
+     * @returns {Promise<Manga>}
```
MikeZeDev: same remark

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+     * @override
+     */
+    async _getMangaFromURI(uri) {
+        if (!this.isBookPageURL(uri)) {
+            alert("Not a valid URL.");
```
MikeZeDev: do not use Alert. throw an exception instead.

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+            return;
+        }
+
+        // first check if it already exists in cache
+        let id = this.getIdFromUri(uri);
+        let manga = await this.findMangaById(id);
```
MikeZeDev: dont bother using cache. Connectors are made to ignore engine shenanigans underneath, like storage stuff :D

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+        if (manga) return manga;
+
+        // else, try new request
+        let request = new Request(uri, this.requestOptions);
+
+        return Engine.Request.fetchUI(request, "bookDetail")
+            .then(data => new Manga(
```
MikeZeDev: same remark about .then syntax.

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+                data.series_id,
+                data.series_title
+            ));
+    }
+
+    // -----------------------------
+
+    /** 
+     * Helper function that checks if a given URI is a valid book page URL for the Ridibooks website.
+     * @param {string|URL} uri - The URI from which to extract the ID.
```
MikeZeDev: same remark

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+     * @returns {boolean}
+     */
+    isBookPageURL(uri) {
+        return this.bookUrlPattern.test(String(uri));
```
MikeZeDev: you dont need that. If you really want to validate pasted url, override     canHandleURI(uri) :D

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /**
+     * Finds a manga by its ID.
+     * 
+     * @async
+     * @param {string} id - The ID of the manga to find.
+     * @returns {Promise<Manga|undefined>}
+     */
+    async findMangaById(id) {
```
MikeZeDev: Connector shouldnt mess with cache, never.

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+        return this.bookUrlPattern.test(String(uri));
+    }
+
+    /**
+     * Extracts the ID from a URI.
+     * 
+     * @param {string|URL} uri - The URI from which to extract the ID.
+     * @returns {string|undefined} The extracted ID, or undefined if no match is found.
+     */
+    getIdFromUri(uri) {
```
MikeZeDev: Not sure you really need a function for this. If its used only once dont bother.

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+            tab: "books",
+            platform: "web",
+            order_by: "popular",
+            limit: 60,
+            offset: 0 
+        }).toString();
+
+        while(nextPage) {
+
+            let request = new Request(
```
MikeZeDev: You can put the arguments on the same line :p That goes for the whole file.

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+                nextPage,
+                this.requestOptions
+            );
+
+            let { data } = await this.fetchJSON(request, 3);
```
MikeZeDev: No need to try 3 times?

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+     * 
+     * @returns {Promise<IManga[]>} The list of manga items.
+     * @async
+     */
+    async _getMangasGql() {
+        // const url = new URL('/graphql', this.apiUrl);
+
+        // const operation = '';
+
+        // const query = ``;
```
MikeZeDev: this needs to be coded, good thing i already gave you the code

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+            let pageMangaList = data.items.map(
+                ({ book }) => ({
+                    id: book.bookId,
+                    title: book.title,
+                })
+            );
+
+            mangaList.push(...pageMangaList);
+
+            if (nextPage = data.pagination.nextPage) {
```
MikeZeDev: While working, it could have been more "hakuneko style" like :D
```
        const mangaList = [];

        let uri = new URL("/v2/category/books", this.apiUrl)
         uri .search = new URLSearchParams({
                category_id: 1600,
                tab: "books",
                platform: "web",
                order_by: "popular",
                limit: 100,
                offset: 0 
           }).toString();


         while (uri) {
            constrequest = new Request(uri, this.requestOptions);
            const { data } = await this.fetchJSON(request);

           const pageMangaList = data.items.map(
                ({ book }) => ({
                    id: book.bookId,
                    title: book.title,
                })
            );
            mangaList.push(...pageMangaList);
            uri = data.pagination.nextPage ? new URL(data.pagination.nextPage, this.apiUrl) : null
        }
        return mangaList;

File: src/web/mjs/connectors/Ridibooks.mjs [hunk 20]
```diff
@@ -0,0 +1,247 @@
+import Connector from "../engine/Connector.mjs";
+import Manga from "../engine/Manga.mjs";
+import Chapter from "../engine/Chapter.mjs";
```
MikeZeDev: You dont need to import Chapter

src/web/mjs/connectors/Ridibooks.mjs
```diff
@@ -0,0 +1,247 @@
...
+
+    /**
+     * Method to get all pages of a chapter.
+     * @param {Chapter} chapter
+     * @returns {Promise<string[]>}
+     * @override
+     */
+    async _getPages(chapter) {
+        let uri = new URL(
+            `/generate/${chapter.id}`,
```
MikeZeDev: Valable for the whole file : we would rather put everything on the same line
```        let uri = new URL(`/generate/${chapter.id}`,  this.pagesApi   ); ```

File: src/web/mjs/connectors/Ridibooks.mjs [hunk 22]
```diff
@@ -0,0 +1,247 @@
...
+     * @override
+     */
+    async _getMangaFromURI(uri) {
+        if (!this.isBookPageURL(uri)) {
+            alert("Not a valid URL.");
+            return;
+        }
+
+        // first check if it already exists in cache
+        let id = this.getIdFromUri(uri);
```
MikeZeDev: No need to use a function for that validation, just override canhandleURI and you dont need to call anything

MikeZeDev (changes requested):
* A couple of things to remove and change
* Maybe use single quotes instead of double quotes
*  Some coding style to change to be like other connectors
*  And last but not least, lint errors to fix https://github.com/manga-download/hakuneko/actions/runs/6842505701/job/18604580350?pr=6426

src/web/mjs/connectors/Ridibooks.mjs
danspinola: I can remove the comments, but if it doesn't hurt why not keep them?

src/web/mjs/connectors/Ridibooks.mjs
danspinola: 👍🏻

src/web/mjs/connectors/Ridibooks.mjs
danspinola: The function is already asynchronous, so the caller will have to await whether I await here or not. And by awaiting here I might be blocking the event loop unnecessarily in exchange for simpler syntax. It is not a good tradeoff in my opinion and it's not like there's a callback hell going on. 

Edit: also, you're wrong, the objects in `seriesBookListJson` has a lot more properties besides `id` and `title`.

src/web/mjs/connectors/Ridibooks.mjs
danspinola: The try 3 times I noticed in other connectors and just copied.

src/web/mjs/connectors/Ridibooks.mjs
danspinola: 👍🏻

src/web/mjs/connectors/Ridibooks.mjs
danspinola: 👍🏻

src/web/mjs/connectors/Ridibooks.mjs
danspinola: I *really* disagree with this statement. Just because a function is used only once doesn't mean it's unnecessary. It's about separation of concerns and clean code. Anyway, the part of the code where I used it will be removed so I'll remove it too, but not because it's "unnecessary".

src/web/mjs/connectors/Ridibooks.mjs
danspinola: I just copied from other connectors.

src/web/mjs/connectors/Ridibooks.mjs
danspinola: Like I explained on Discord. The query you found and copied from the site is **not** for webtoons. I am **not** going to implement it. I'd rather erase the skeleton if it bothers you.

src/web/mjs/connectors/Ridibooks.mjs
danspinola: Same reply about .then syntax

src/web/mjs/connectors/Ridibooks.mjs
Sheepux: Called once + 1 liners functions is a pain.
as a maintainer, it adds a burden when there is an issue on the connector and you have to dig through multiple smalls functions.

src/web/mjs/connectors/Ridibooks.mjs
Sheepux: It adds extra lines on something that won't be used.
Keeping each connector small is better for the maintainers mental-health. 
It's also to align with all thousands connectors coding style.
The function being described in Connector.mjs is enougth

src/web/mjs/connectors/Ridibooks.mjs
danspinola: I'm not trying to be combative, but I wholeheartedly disagree. I'm pretty sure your editor of choice has a way to go to a function's definition and to return to the previous position. It's a non-issue. But anyway, like I said, I'm removing this.

src/web/mjs/connectors/Ridibooks.mjs
Sheepux: we do not use promise.then anymore as they are a burden to read.
using await is exactly the same as using .then but it's more recent and easier to read.

danspinola: They are not exactly the same as I explained and while they may be easier to read I don't think it's a good tradeoff. And in this case it's not like it's that much harder to read. Also, like I also said, the function is async, so the caller is going to await anyway, which makes no difference if I await here or not. But meh 🤷🏻‍♂️.

MikeZeDev: > The function is already asynchronous, so the caller will have to await whether I await here or not. And by awaiting here I might be blocking the event loop unnecessarily in exchange for simpler syntax. It is not a good tradeoff in my opinion and it's not like there's a callback hell going on.
> 
> Edit: also, you're wrong, the objects in `seriesBookListJson` has a lot more properties besides `id` and `title`.

But the calling function should only take care of id and title and discard the rest.
i am aware there is no need to use await on return statement but the .then syntax is a pain to read and its especially avoided in Hakuneko. Its nice having same code pattern everywhere :)

src/web/mjs/connectors/Ridibooks.mjs
MikeZeDev: I figured that out ofc, no pb :)

src/web/mjs/connectors/Ridibooks.mjs
MikeZeDev: I see your point , but code is clean enough. Too much decomposition is painful and we are accustomed to inline stuff like that. And using a function for something called once is overkill. We wouldnt mind if it was even only call twicev  ofc.

src/web/mjs/connectors/Ridibooks.mjs
MikeZeDev: Ah. Its fine then. Better not leaving empty function,yup :)

danspinola:
Requested changes applied.

MikeZeDev:
Everything is good. I would have pushed myself the icon to your repo since i forget to review it but pushing is disabled on your end. It needs to be a png, not a jpeg.

Thanks for your contribution :)

MikeZeDev (approved):

danspinola:
Changed file to `png`.

27136773:
[Missing in HaruNeko](https://github.com/search?q=repo%3Amanga-download%2Fharuneko%20%27ridibooks%27&type=code)

PR #6435, opened 2023-11-16
Create oremangaNet.mjs

YainuWan11:
Create webtoon oremangaNet.mjs

### DIFF
--- a/src/web/mjs/connectors/oremangaNet.mjs
+++ b/src/web/mjs/connectors/oremangaNet.mjs
@@ -1,0 +1,12 @@
+import oremangaNet from './templates/oremangaNet.mjs';
+
+export default class oremanga extends oremangaNet {
+
+    constructor() {
+        super();
+        super.id = 'oremanga';
+        super.label = 'oremanga';
+        this.tags = [ 'manga', 'webtoon', 'thai' ];
+        this.url = 'https://www.oremanga.net';
+    }
+}


YainuWan11 (commented):
Reviews oremanga

src/web/mjs/connectors/oremangaNet.mjs
```diff
@@ -0,0 +1,12 @@
+import oremangaNet from './templates/oremangaNet.mjs';
```
MikeZeDev, line 1: There is no oremangaNet.mjs file in templates what are you trying to import?

src/web/mjs/connectors/oremangaNet.mjs
```diff
@@ -0,0 +1,12 @@
...
+
+export default class oremanga extends oremangaNet {
```
MikeZeDev, line 3: Are you trying to extends a class that doesnt exists? 

Moreover class name 'oremanga' doesnt match filename 'oremangaNet' (case and name)

src/web/mjs/connectors/oremangaNet.mjs
```diff
@@ -0,0 +1,12 @@
...
+
+    constructor() {
+        super();
+        super.id = 'oremanga';
+        super.label = 'oremanga';
```
MikeZeDev, line 8: label is not necessarly  identical to the id :) You could have written something like "Oremanga", or "OreManga".

MikeZeDev (commented):
It seems to me that you tried a bunch of copypaste, didnt test it and threw a pull request.
Thats not how programming works, i am afraid.

* "import" from a template can  be done when the website use a predefined known template. You can see in our mjs we have a bunch of "WordpressMadara". You can import the template and just fill some infos from the website and it works.

*  In case your website doesnt use a known template you have to code  "_getMangas", "_getChapters", "_getPages" to get Manga lists, chapters and pictures links. And import "Connector.mjs" so you can "extends Connector".

Look how its done for other websites there are plenty of examples.

Feel free to ask questions if you want help contributing. I will close this PR for now.

MikeZeDev:
I found that this website is very similar to Neumanga that hakuneko supports. The template is Zmanga. You can look into Neumanga code if you want

PR #6462, opened 2023-11-28
fix(CrazyScans): update url

TristanWasTaken:
old: https://mangacultivator.com
new: https://mangacult.org


MikeZeDev (approved):
Thanks. Already up to date in HaRUneko  :p

PR #6671, opened 2024-01-11
Fix ZeroScans: change url and fix paste support

trendbloat:


src/web/mjs/connectors/ZeroScans.mjs
```diff
@@ -24,11 +24,11 @@ export default class ZeroScans extends Connector {
     }
 
     async _getMangaFromURI(uri) {
-        const request = new Request(uri, this.requestOptions);
-        const script = `new Promise(resolve => resolve(JSON.stringify(window.__ZEROSCANS__)));`;
-        const { data } = await Engine.Request.fetchUI(request, script);
-        const details = data[0].details;
-        return new Manga(this, `${details.id}_${details.slug}`, details.name.trim());
+        const slug = uri.pathname.replace(/[/]+$/, '').split('/').pop();
```
MikeZeDev: i think a regex would be better to extract slug

`const slug = uri.href.match(/\/comics\/([^/]+)$/)[1];`

MikeZeDev (changes requested):
* Minor change requested
* Turn out JS value isnt always filled, nice catch 👍

MikeZeDev (approved):

PR #6679, opened 2024-01-14
Add connector for Nosajj

Tungsten-W-74:
http://shonanjunaigumi.weebly.com/projects.html

### DIFF
--- a/src/web/mjs/connectors/Nosajj.mjs
+++ b/src/web/mjs/connectors/Nosajj.mjs
@@ -1,0 +1,52 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class Nosajj extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'Nosajj';
+        super.label = 'Nosajj (shonanjunaigumi)';
+        this.tags = ['manga', 'english', 'scanlation'];
+        this.url = 'http://shonanjunaigumi.weebly.com';
+    }
+
+    async _getMangaFromURI(uri) {
+        const request = new Request(uri, this.requestOptions);
+        const id = uri.pathname + uri.search;
+        const title = (await this.fetchDOM(request, 'div.paragraph'))[0].textContent.trim();
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        const uri = new URL('/projects.html', this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, 'td[align^="center"] > a');
+        return data.map(element => {
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.textContent.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        const uri = new URL(manga.id, this.url);
+        const request = new Request(uri, this.requestOptions);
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Nosajj.mjs
```diff
@@ -0,0 +1,52 @@
...
+                language: '',
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        const uri = new URL(chapter.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, 'div.wsite-image>a>source');
+        return Array.from(data).map(element => this.url + new URL(element.src).pathname);
```
MikeZeDev, line 50: fetchDOM returns an Array.  no need for Array.from to use map

MikeZeDev: Also, we'd rather use this

```js
        return data.map(image => this.getAbsolutePath(image, request.url));
 ```

File: src/web/mjs/connectors/Nosajj.mjs [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+                title: element.textContent.trim()
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        const uri = new URL(manga.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = (await this.fetchDOM(request, 'td.wsite-multicol-col'))[0].querySelectorAll('a');
+        return Array.from(data).filter(element => element.textContent.includes("Chapter")).map(element => {
```
MikeZeDev, line 37: What if they omit "chapter" in the name? I'd rather use a proper CSS selector than a filter.
Why not
 ```js
 'a[href$=".html"]'
```

since i assume all chapter links end with html? 

Oh and no need to use Array.from, you can do it like this
 ```js
        const data = [...(await this.fetchDOM(request, 'td.wsite-multicol-col'))[0].querySelectorAll('a[href$=".html"]')];
```

src/web/mjs/connectors/Nosajj.mjs
```diff
@@ -0,0 +1,52 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class Nosajj extends Connector {
+
+    constructor() {
+        super();
+        super.id = 'Nosajj';
+        super.label = 'Nosajj (shonanjunaigumi)';
```
MikeZeDev, line 9: I'd rather use "shonanjunaigumi' as id and "Shonan Junai Gumi" as label, i think this website its better known by this name, but its personal taste.

src/web/mjs/connectors/Nosajj.mjs
```diff
@@ -0,0 +1,52 @@
...
+        this.tags = ['manga', 'english', 'scanlation'];
+        this.url = 'http://shonanjunaigumi.weebly.com';
+    }
+
```
MikeZeDev, line 13: Website lacks an icon. Since website have default weebly icon you may copy default hakuneko icon (\\img\connectors\default) and rename it to match the website id.

src/web/mjs/connectors/Nosajj.mjs
MikeZeDev: id MUST be lowercase

MikeZeDev (changes requested):
* Id must be lowercase
* Must provide icon
* minor code changes requested

PR #6690, opened 2024-01-18
Add MangAdventure template

ObserverOfTime:
Used by Assorted Scans & Arc-Relight.


src/web/mjs/connectors/templates/MangAdventure.mjs
```diff
@@ -0,0 +1,89 @@
...
+    constructor() {
+        super();
+        super.id = undefined;
+        super.label = undefined;
+        this.url = undefined;
+        this.requestOptions.headers.set('x-user-agent', this.userAgent);
+    }
+
+    get platform() {
+        switch (process.platform) {
```
MikeZeDev, line 14: Is it really needed to make an user agent? Cant we just use hakuneko default one ?

src/web/mjs/connectors/templates/MangAdventure.mjs
```diff
@@ -0,0 +1,89 @@
...
+                return 'Linux';
+            case 'darwin':
+                return 'Macintosh';
+            default:
+                return process.platform;
+        }
+    }
+
+    get userAgent() {
+        return `Mozilla/5.0 (${this.platform}) Chrome/${process.versions.chrome} Hakuneko`;
```
MikeZeDev, line 28: Is it really needed to make an user agent? Cant we just use hakuneko default one ?

src/web/mjs/connectors/templates/MangAdventure.mjs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /** @param {string} path */
```
MikeZeDev: JSdoc is fine but we tend to remove it in connectors. You may want to remove it too, so its like in the rest of the codebase.

src/web/mjs/connectors/templates/MangAdventure.mjs
```diff
@@ -0,0 +1,89 @@
...
+        const url = new URL('/api/v2' + path, this.url);
+        return new Request(url, this.requestOptions);
+    }
+
+    /**
+     * @param {URL} uri
+     * @override
+     */
+    canHandleURI(uri) {
+        return this.url == uri.origin && uri.pathname.startsWith('/reader/');
```
MikeZeDev, line 37: Personal taste : i'd rather use a regexp than a startswith but its up to you. You may just resolve this conversation and dont change anything.

MikeZeDev (commented):
Minor changes required, and where did you get that MangaAdventure name?

src/web/mjs/connectors/templates/MangAdventure.mjs
ObserverOfTime: Does it include "Hakuneko"?

ObserverOfTime:
> Minor changes required, and where did you get that MangaAdventure name?

https://github.com/mangadventure/MangAdventure

src/web/mjs/connectors/templates/MangAdventure.mjs
ObserverOfTime: A regexp is considerably slower and unnecessary in this case.

src/web/mjs/connectors/templates/MangAdventure.mjs
MikeZeDev: Why do we care?  :D

ObserverOfTime: Curiosity.
image

src/web/mjs/connectors/templates/MangAdventure.mjs
27136773: This makes it very easy to detect requests from HakuNeko and perform individual actions against the user such as IP block/ban, tracking, ...

ObserverOfTime: There won't be any of that.

27136773: How would someone know?
Nobody has control over the website owners. Maybe they will comply now, but what about in 3 month?

Furthermore less source code is better, especially when it is designed to bypass [the mechanism of stripping the user-agent](https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/engine/Request.mjs#L440) that was intentionally implemented in HakuNeko to avoid tracking and detection.

ObserverOfTime: I am the website owner.

27136773: That doesn't affect any of the statements above.

ObserverOfTime: OK, don't take my word for it. If users complain about getting blocked on either site you can remove the UA.

ObserverOfTime:
So?

MikeZeDev:
> So?

Waiting for requested changes to be done. 🤷

ObserverOfTime:
I already explained why they're unnecessary.

MikeZeDev:
And we explained how the user agent code you added 

* bypass an essential Hakuneko feature
* is useless.

If you dont plan to block haku there is no reason to add it in the first place. If its useless, its bloat.
Ofc we aint gonna remove it ourselves as it wouldn't be respectful, and as maintainers we ask PR author politely :D.

ObserverOfTime:
It's useful for analytics. Tachiyomi, Aidoku, Paperback, Kotatsu all support custom UAs. Kotatsu even has one built-in.

MikeZeDev:
> It's useful for analytics. Tachiyomi, Aidoku, Paperback, Kotatsu all support custom UAs. Kotatsu even has one built-in.

You are mixing analytics and Custom UA, those are not the same.

* Custom UA exists to help the user.
 * The other is useful for the dev of the app itself. Its the dev decision to include analytics in its own app. You want to push analytics in an app you dont own  for your own purpose, how can you think its not a problem? 

Oh and when i said useless i meant for the user.

Personal opinion starts here : 

You seems to want to use Hakuneko to push for something you want for yourself. Well ofc like other contributors, they all want a feature or a website added, and its fine 👍  But if that thing can be used against other users, or if they gain no benefit from it i'd say it should not be included at all. Its not personal .

ObserverOfTime:
I don't think it's a problem the same way the maintainers of the other apps I mentioned don't think it's a problem.
The only purpose of the UA is to let us know what percentage of visitors uses each app (Tachiyomi is 31%).
None of the users are personally identified or will ever be blocked because of it.

ObserverOfTime:
I see you have no intention of adding it with the custom UA, but I have no intention of adding it without it.
So, as the site admin, I have to request that you remove the existing scraper from both here and haruneko.
Users can download chapters either directly (after logging in) or by using any other complying app.

MikeZeDev:
Make a Removal request

ObserverOfTime:
Let me know if you change your minds in the future.

PR #6727, opened 2024-01-27
Fixing teamx url

AbdulElahA:
Teamx url has been changed.

### DIFF
--- a/src/web/mjs/connectors/Team1x1.mjs
+++ b/src/web/mjs/connectors/Team1x1.mjs
@@ -14,7 +14,7 @@ export default class Team1x1 extends Connector {
                 label: 'URL',
                 description: 'This website changes their URL regularly.\nThis is the last known URL which can also be manually set by the user.',
                 input: 'text',
-                value: 'https://mnhaestate.com'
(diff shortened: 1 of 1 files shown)


MikeZeDev:
I wonder if its worth accepting this PR, since the point of the option is precisely to avoid such PR because damn websites change domain everytime.

Hum well validating this costs nothing anyway but dont make a PR each time :D

src/web/mjs/connectors/Team1x1.mjs
```diff
@@ -14,7 +14,7 @@ export default class Team1x1 extends Connector {
                 label: 'URL',
                 description: 'This website changes their URL regularly.\nThis is the last known URL which can also be manually set by the user.',
                 input: 'text',
-                value: 'https://mnhaestate.com'
+                value: 'https://teamxnovel.com/'
```
MikeZeDev, line 17: Remove trailing "/'

MikeZeDev (changes requested):

MikeZeDev:
closing an useless PR

PR #6762, opened 2024-02-07
Unify the light and dark versions (except for theme.html)

jabagawee:
Includes the fix in PR #2274.


manga-download-owner:
Unfortunately this version of HakuNeko is in feature freeze, only critical website maintenance PR changes will be merged
→ Closed
→ See [manga-download/haruneko](https://github.com/manga-download/haruneko) for future development

PR #6763, opened 2024-02-07
Fix performance issues while rendering large lists

jabagawee:
Set initialCount for dom-repeat element that can hold many items

The manga list can hold up to hundreds of thousands of items or more, and this causes major performance issues during rendering (see Polymer/polymer#2537). The solution is to ask <dom-repeat> to render a limited subset of the items per frame instead of all in one frame, as explained in https://polymer-library.polymer-project.org/3.0/docs/devguide/templates#large-list-perf.

### DIFF
--- a/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
+++ b/src/web/lib/hakuneko/frontend@classic-dark/mangas.html
@@ -137,7 +137,7 @@
                     and go to the network tab
                 </li>
             </template>
-            <template is="dom-repeat" items="[[ mangaList ]]" filter="[[ filterMangas(mangaPattern) ]]" rendered-item-count="{{ mangaFilteredCount }}">
+            <template is="dom-repeat" items="[[ mangaList ]]" filter="[[ filterMangas(mangaPattern) ]]" rendered-item-count="{{ mangaFilteredCount }}" initial-count="500">
                 <li class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item.id) ]]" title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</li>
             </template>
         </ul>
--- a/src/web/lib/hakuneko/frontend@classic-light/mangas.html
+++ b/src/web/lib/hakuneko/frontend@classic-light/mangas.html
@@ -137,7 +137,7 @@
                     and go to the network tab
                 </li>
             </template>
-            <template is="dom-repeat" items="[[ mangaList ]]" filter="[[ filterMangas(mangaPattern) ]]" rendered-item-count="{{ mangaFilteredCount }}">
+            <template is="dom-repeat" items="[[ mangaList ]]" filter="[[ filterMangas(mangaPattern) ]]" rendered-item-count="{{ mangaFilteredCount }}" initial-count="500">
                 <li class$="manga [[ item.status ]] [[ getMangaClass(selectedManga, item.id) ]]" title$="[[ item.title ]]&#10;[[ item.connector.label ]]" on-click="onMangaClicked">[[ item.title ]]</li>
             </template>
         </ul>


MikeZeDev (commented):
Thanks for your contribution. Unfortunately am against approving this contribution.

I tested it on Mangadex and its not good.

Without it :  list is fully browsable after a 7 seconds load
With it : list is immediatly visible BUT https://i.imgur.com/wiabM5e.gif

* It keeps counting endlessly 
* I cant reach the end of the list forcefully using  END or Pagedown key.

Maybe there are workadounds or fixes for that?

jabagawee:
The counting is just a loading indicator to show how many rows have actually been rendered. You would be able to reach the end of the list after the loading is complete. The question just becomes whether you believe it's a better user experience to be able to browse a partially loaded list immediately or if it's better to browse the entire list after some period of non-responsiveness.

My general inclination is that users would prefer the former rather than the latter, but it depends on which metrics you want to prioritize out of https://web.dev/articles/user-centric-performance-metrics#important_metrics_to_measure. The PR brings improvements to first contentful paint, largest contentful paint, first input delay, and interaction to next paint, whereas the status quo prioritizes minimizing total blocking time. 

In terms of the two points you bring up:

* *It keeps counting endlessly*: If understood as a loading indicator, it makes sense that it would take longer to load a large list. I'm not sure the user would be frustrated to see the number constantly ticking, as they would understand "oh there's like hundreds of thousands of manga to load, at least I can see the ones that have loaded so far!"
* *I cant reach the end of the list forcefully using END or Pagedown key*: Is it common for users to want to scroll to the end of a super long list? I can see wanting to scroll to the end of a list if the list were only a few dozen elements long or wanting to scroll to the end to get an entry that is visible on screen, but I don't know of any use cases for someone to want to scroll to the bottom of a list hundreds of thousands of elements long.

---

One separate matter that I realized while testing locally: this PR helps with the initial render of a list of manga available from a connector, but it actually doesn't help when the filter is changed. I'm not 100% sure, but I believe that this wasn't fixed until Polymer 3.4.0 with Polymer/polymer@b40840b9. I know it's unrealistic to suggest upgrading that since development efforts have turned to haruneko, so I understand if you want to just close this PR without merging it in.

MikeZeDev:
> * it common for users to want to scroll to the end of a super long list?
Not really i was just bruteforcing stuff, its part of testing. I tried to search an item in the bottom of the list too.
If i learned something with users in my previous jobs its "expect the unexpectable" :D

> It keeps counting endlessly: If understood as a loading indicator, it makes sense that it would take longer to load a large list. I'm not sure the user would be frustrated to see the number constantly ticking, as they would understand "oh there's like hundreds of thousands of manga to load, at least I can see the ones that have loaded so far!"

My computer is not so old but not so young either. Loading the whole Mangadex list takes up to 7 seconds, its not really a big deal. Moreover , i may be wrong but i am considering user feedback. I have the feeling that its easier for them to understand "well its frozen (not really there is a loading symbol, albeit quite small) for a couple of seconds because its loading a huge list so be it" and after it can be scrolled & searched flawlessly. And yep, filtering delays stays the same unfortunately.

Imho, if we dont patch the growing counter it will give the impression its taking waaaaaaaaaaaaay longer while its taking 7 seconds without the PR 🤷 . Because its going 500 by 500, and there are nearly 80k mangas. I dont think the average user will think about rendering.

>One separate matter that I realized while testing locally: this PR helps with the initial render of a list of manga available from a connector, but it actually doesn't help when the filter is changed. I'm not 100% sure, but I believe that this wasn't fixed until Polymer 3.4.0 with https://github.com/Polymer/polymer/commit/b40840b9. I know it's unrealistic to suggest upgrading that since development efforts have turned to haruneko, so I understand if you want to just close this PR without merging it in

Its handled quite differently in Haruneko :). I am not working on the UI i am not really familiar with such things.

I think the idea is good but this lacks something, and i am not skilled enough on this matter. I'd like to ask for Ronny or Sheep opinion on that.

Sheepux:
> Its handled quite differently in Haruneko :). I am not working on the UI i am not really familiar with such things.

I'm just using a virtual list https://github.com/sveltejs/svelte-virtual-list
It's only rendering what's usefull. The counter at the bottom is just the number of filtered items / total)

> I think the idea is good but this lacks something, and i am not skilled enough on this matter. I'd like to ask for Ronny or Sheep opinion on that.

I have zero knowledge of polymer and not going to invest time as we're working on haru :D

jabagawee:
Closing as it seems to be not a great use of time for everyone.

MikeZeDev:
Thanks anyway.

PR #6777, opened 2024-02-12
Fix HentaiRead: get chapters and pages

MikeZeDev:


src/web/mjs/connectors/HentaiRead.mjs
```diff
@@ -10,19 +10,32 @@ export default class HentaiRead extends WordPressMadara {
...
-        return pageList.map(link => {
-            let uri = new URL(link);
-            uri.searchParams.set('quality', '100');
-            uri.searchParams.delete('w');
-            return uri.href;
-        });
+        const request = new Request(this.url + chapter.id, this.requestOptions);
+        const script = `
+            new Promise((resolve, reject) => {
+                setTimeout(() => {
```
27136773, line 26: Use `try...catch` inside of `setTimeout`, otherwise the promise may never settle in case of an error

```javascript
try {
    // ...
} catch(error) {
    reject(error);
}
```

27136773 (approved):

PR #6835, opened 2024-02-28
Update Generated User-Agent Headers

27136773:
Use more recent `User-Agent` headers in generator.
This may be helpful when using Google SingleSignOn with certain websites


MikeZeDev:
Obviously no problems in the PR but after i tested a bit i see no difference with Google SingleSignOn .
It still says old browser after i try to log in

MikeZeDev (approved):

PR #6930, opened 2024-04-01
fix MangaSect source

kajack:
* Fix MangaSect source site.

### DIFF
--- a/src/web/mjs/connectors/MangaSect.mjs
+++ b/src/web/mjs/connectors/MangaSect.mjs
@@ -8,7 +8,7 @@ export default class MangaSect extends Connector {
         super.id = 'mangasect';
         super.label = 'MangaSect';
         this.tags = [ 'webtoon', 'english' ];
-        this.url = 'https://mangasect.com';
+        this.url = 'https://mangasect.net';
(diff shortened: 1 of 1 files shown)


MikeZeDev (changes requested):
This is overcomplicated for nothing i think.

1) Mangasect.com is now mangasect.net you should change url. 
2) They changed manga template, its now what we call "MojoPortalComic"
3) Fix lint errors

I'd suggest to check how are coded MojoPortalComic websites we already support in hakuneko.

There is no need for an overcomplicated getpages, at least in Haruneko i managed to code it with standard code (css on page)

kajack:
Ups, right forgot about Lint.

Alright will check those MojoPortalComic

MikeZeDev:
Ok i have made a fix on your repo, a custom getpages was just needed tell me if that works :p

https://github.com/manga-download/hakuneko/pull/6930/commits/51f081cbadcf6e2bfe698293b1240acb998a7968

kajack:
Yeah worked. Indeed, cleaner and easier than the code I had done.

MikeZeDev (approved):

PR #6995, opened 2024-04-18
updated connector drakescans

YurDavid:
fixes for drakescan

### DIFF
--- a/src/web/mjs/connectors/DrakeScans.mjs
+++ b/src/web/mjs/connectors/DrakeScans.mjs
@@ -1,6 +1,6 @@
 import WordPressMangastream from './templates/WordPressMangastream.mjs';
 
-export default class MonzeeKomik extends WordPressMangastream {
+export default class DrakeScans extends WordPressMangastream {
     constructor() {
         super();
         super.id = 'drakescans';
--- a/src/web/mjs/connectors/DrakeScans.mjs
+++ b/src/web/mjs/connectors/DrakeScans.mjs
@@ -1,12 +1,12 @@
-import WordPressMadara from './templates/WordPressMadara.mjs';
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
 
-export default class DrakeScans extends WordPressMadara {
-
+export default class MonzeeKomik extends WordPressMangastream {
     constructor() {
         super();
         super.id = 'drakescans';
         super.label = 'DrakeScans';
         this.tags = [ 'webtoon', 'english'];
         this.url = 'https://drakescans.com';
+        this.path = '/manga/list-mode/';
     }
 }


src/web/mjs/connectors/DrakeScans.mjs
```diff
@@ -1,12 +1,12 @@
-import WordPressMadara from './templates/WordPressMadara.mjs';
-
-export default class DrakeScans extends WordPressMadara {
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
 
+export default class MonzeeKomik extends WordPressMangastream {
```
MikeZeDev: Its not Monzeekomik ;). Do not change class.

MikeZeDev (changes requested):

src/web/mjs/connectors/DrakeScans.mjs
DraconicSalad: Drake scan website dosen't uses the wordpressmadara type anymore it uses one similar to monzeekomik,  so it has to change

reason is, it  dosen't display mangas anymore, so obviously it dosen't download any new chapters
after did the changes it worked can you please check

DraconicSalad: Drake scan website dosen't uses the wordpressmadara type anymore it uses one similar to monzeekomik,  so it has to change

reason is, it  dosen't display mangas anymore, so obviously it dosen't download any new chapters
after did the changes it worked can you please check

YurDavid: @MikeZeDev drake scans are using a similar template, so obviously i changed the template to match its new one, and it now works

YurDavid:
Screenshot from 2024-04-18 09-43-15
screenshot showing the current Drake scans dosen't work

src/web/mjs/connectors/DrakeScans.mjs
MikeZeDev: Your class name is wrong. It needs to be matching the file name, change it.

YurDavid:
after made the changes you can see this works
Screenshot from 2024-04-18 09-49-12
ne of the

src/web/mjs/connectors/DrakeScans.mjs
MikeZeDev: Stop spamming the same useless screenshot and fix what needs to be fixed 🙏 . I perfectly know it doesnt work, since they changed template you dont have to convince anyone :D. How hard it is to understand that "MonzeeKomik" should be "DrakeScans" 🤔

YurDavid: @MikeZeDev i apologize i didn't noticed that, i change it now

YurDavid: @MikeZeDev done

MikeZeDev (approved):
Thanks for your contribution 👍

PR #7015, opened 2024-04-24
FIx WebtoonHatti: fix getting pages, again

MikeZeDev:
this time, with an added bonus.


manga-download-owner (approved):

PR #7016, opened 2024-04-24
Fix MangahubRU: various fixes

MikeZeDev:
*   Fix Mangalist CSS selector
 *   Get pages normally (no json)
 *   Fix Clipboard support
 *   Use referrer for pages
 *   Set adult cookie for chapters


manga-download-owner (approved):

PR #7025, opened 2024-04-24
FIX UzayManga : website revamping

MikeZeDev:
Fixes https://github.com/manga-download/hakuneko/issues/6503


manga-download-owner (approved):

PR #7057, opened 2024-05-06
feat(CulturedWorks): create connector

TristanWasTaken:
https://culturedworks.com/


src/web/mjs/connectors/CulturedWorks.mjs
```diff
@@ -0,0 +1,12 @@
+import WordPressMangastream from './templates/WordPressMangastream.mjs';
+
+export default class CulturedWorksnpm extends WordPressMangastream {
```
MikeZeDev: 'CulturedWorksnpm' ? Should no it be 'CulturedWorks'?

MikeZeDev:
Perhaps it time to add it to haruneko? Its not hard :)

MikeZeDev (changes requested):

src/web/mjs/connectors/CulturedWorks.mjs
TristanWasTaken: yeah, I was wondering what ate the "npm" 😅

TristanWasTaken:
> Perhaps it time to add it to haruneko? Its not hard :)

sure, but I think I already shared my disgust for some parts of the new api when I asked on Discord.
Way too much boilerplate for what could be handled by shared components, like templates in the current version.

TristanWasTaken:
> Perhaps it time to add it to haruneko? Its not hard :)

sure, but I'm not using haru as long as it's pretty much unusable
1. can't even move the window
2. forces you to read long strip/webtoon or 2 page even for single page manga
3. clunky ui
4. can't even retry downloads
5. no way to pause the download -> doesn't detect if no disk space is left
6. no way to reorder downloads or give them a priority
7. doesn't detect if chapters were already downloaded
8. local search didn't work

dev:
9. some functions are still called "mangas". The plural of manga is manga.
10. would've been nice if the dectorators, etc weren't a sub-folder of the connectors/websites folder
11. the 2 files per connector could've been moved to their own folder

and a few more that I forgot

MikeZeDev:
You do not have to use it :D.

Adding a Madara website can be done in like 10 minutes. Need file, icon and test 🤷

MikeZeDev (approved):

PR #7069, opened 2024-05-07
chore: consistently enforce line endings 

TristanWasTaken:
Consistently enforce line endings regardless of what operating system is used.
Good article about this: https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git

If you think this is unnecessary, just close the pr. Though it wouldn't hurt to adopt this.

Existing files won't be changed (unless you specifically use a git cmd like `git add --renormalize .`).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# file is at the root of the project
+root = true
+
+# enforce lf line ending
+[*]
+end_of_line = lf
+
+# crlf for files that are incompatible with lf
+[*.bat]
+end_of_line = crlf
+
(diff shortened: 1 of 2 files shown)


TristanWasTaken:
note that I added .bat & .cmd in case the repo will someday have a bat or cmd file (so it doesn't become a pitfall)
should that be removed?

27136773:
I'm not sure what impact these changes would have to existing repositories of developers. Since there is no real benefit i don't think this change should be merged.
May be more appropriate for HaruNeko.

PR #7105, opened 2024-05-25
Fix klm website change

### LINKED ISSUE #7104: [KissAway] Connector not working
### Is there an existing issue for this connector?

- [X] I have searched the existing issues

### Did you read the troubleshooting guide ?

- [X] I read it but nothing seems related

### Is the website of the connector working properly ?

- [X] Yes i'm able to see the manga within my standard browser

### Make sure to try Nightly Build

- [X] Yes nightly build didn't fix the issue

### What kind of issue are you encountering

Manga list isn't refreshing

### Describe the bug

The website has changed the way it gets chapter lists and image lists

### To Reproduce

1. KissAway.mjs
2. https://klz9.com/ybed-tensei-shitara-slime-datta-ken-raw.html  chapters not update
3. https://klz9.com/jxsh-tensei-shitara-slime-datta-ken-raw-chapter-118.html pages not update

realabertian:
fix #7104
The webpage dynamically generates a 'token' on the previous page through some unknown method, and it uses the 'token' to access the next-level page. The token is one-time use. 
I analyzed it for a long time and couldn't figure it out.

### DIFF
--- a/src/web/mjs/connectors/KissAway.mjs
+++ b/src/web/mjs/connectors/KissAway.mjs
@@ -55,7 +55,7 @@ export default class KissAway extends FlatManga {
                 const data = await response.text();
                 const dom = new DOMParser().parseFromString(data, "text/html");
                 const nodes = [...dom.querySelectorAll('img.chapter-img')];
-                resolve(nodes.map(picture => picture.src));
+                resolve(nodes.map(picture => picture.src).filter(image => !image.match(/olimposcan/)));
             });
         `;
         const request = new Request(uri, this.requestOptions);
--- a/src/web/mjs/connectors/KissAway.mjs
+++ b/src/web/mjs/connectors/KissAway.mjs
@@ -14,17 +14,12 @@ export default class KissAway extends FlatManga {
     async _getChapters(manga) {
         const script = `
             new Promise(async resolve => {
-                const uri = new URL('app/manga/controllers/cont.listChapter.php', window.location.origin);
-                uri.searchParams.set('slug', dataL);
-                const response = await fetch(uri);
-                data = await response.text();
-                const dom = new DOMParser().parseFromString(data, "text/html");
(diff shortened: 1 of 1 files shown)


MikeZeDev:
i think there is an easier solution.

https://github.com/keiyoushi/extensions-source/pull/3208

the endpoint is just randomized but ends with a fixed extension

src/web/mjs/connectors/KissAway.mjs
```diff
@@ -14,17 +14,12 @@ export default class KissAway extends FlatManga {
...
-                const dom = new DOMParser().parseFromString(data, "text/html");
-                const nodes = [...dom.querySelectorAll('a.chapter[title]')];
-                const chapters= nodes.map(chapter => {
+                const nodes = document.querySelectorAll("div#list-chapter a")
+                const chapters = Array.prototype.map.call(nodes, item => {
                     return {
-                        id : chapter.pathname,
-                        title : chapter.title
-                    };
+                        id : item.href,
```
MikeZeDev, line 20: never use full url for identifiers unless absolute necessity. But yeah, there may be a typo here

MikeZeDev (changes requested):
Changes are not ideal and are prone to fail because of a race condition. We don't wait for the chapter/page gathering XHR request to gather the html nodes.

Solution : 

Make the request ourselves and parse the html like before. 
Random endpoints can be used, like in the Mihon extension i mentioned.

realabertian:
> Changes are not ideal and are prone to fail because of a race condition. We don't wait for the chapter/page gathering XHR request to gather the html nodes.
> 
> Solution :
> 
> Make the request ourselves and parse the html like before. Random endpoints can be used, like in the Mihon extension i mentioned.

Got it, I'll handle it

MikeZeDev (approved):

PR #7139, opened 2024-06-02
Voidscans : Change domain

### LINKED ISSUE #7135: [<Void Scans>] Domain Change
### Is there an existing issue for this connector?

- [X] I have searched the existing issues

### Did you read the troubleshooting guide ?

- [X] I read it but nothing seems related

### Is the website of the connector working properly ?

- [X] Yes i'm able to see the manga within my standard browser

### Make sure to try Nightly Build

- [X] Yes nightly build didn't fix the issue

### What kind of issue are you encountering

Other

### Describe the bug

So there has been existing issues about voidscans.com releated to infernal void scans but the issue isn't resolved and it is listed as closed (in hakuneko if you pick void scans it still gives 403 error since it is now https://hivescans.com/) and it still says voidscans.com error so it hasn't been updated even tho the issue is listed as closed.

### To Reproduce

Not Needed

MikeZeDev:
Fixes https://github.com/manga-download/hakuneko/issues/7135


src/web/mjs/connectors/NtsVoidScans.mjs
```diff
@@ -1,12 +1,24 @@
 import WordPressMadara from './templates/WordPressMadara.mjs';
-//dead?
 export default class NtsVoidScans extends WordPressMadara {
 
     constructor() {
         super();
         super.id = 'ntsvoidscans';
-        super.label = 'Void Scans';
+        super.label = 'Void Scans ';
```
27136773: Why was a space added?
This may lead to problems when sub-folders are created in download directory since windows does not allow a name to end with`.` or ` ` .

src/web/mjs/connectors/NtsVoidScans.mjs
```diff
@@ -1,12 +1,24 @@
 import WordPressMadara from './templates/WordPressMadara.mjs';
-//dead?
 export default class NtsVoidScans extends WordPressMadara {
```
27136773: Shouldn't this website be removed instead of throwing errors everywhere?

27136773 (changes requested):

src/web/mjs/connectors/NtsVoidScans.mjs
MikeZeDev: There was 2 websites, they rebranded, fused, etc..., its a mess.

I did it like that so people are not lost. Ofc, it can be deleted if you want.

27136773: Should be removed for the following reasons:
- Existing bookmarks for `ntsvoidscans` will not be preserved/migrated by this approach
- Usually dead websites are removed, so it would be inconsistent to keep just this one in the list
- If this approach is kept there will be dozens of websites in the furture throwing `Website is dead, use ... instead`
- There is no benefit in keeping a non-functional website, if a user ask in support about `NTS Void Scans` there is the generic answer: Website is dead or user has to provide working url

27136773 (approved):

Issue #5864, opened 2023-06-01
[Site Request] fr-scan


Tankdur:
**Name of the website**
fr-scan.com

**Website urls (examples below)**
- Site: https://fr-scan.com/
- Manga List: https://fr-scan.com/manga/
- Manga example: https://fr-scan.com/manga/one-punch-man/
- Chapter online viewer example: https://fr-scan.com/manga/one-punch-man/chapitre-1-vf/

**Languages**
French

**Website relationship*
If applicable describe the relation with any other website (eg : "alternative domain, copy of ...").

**Additional details**
Add any other context details that you may have found (like template to reuse)

MikeZeDev:
Madara template.

flochocinco:
fixed proposed in PR #7141

MikeZeDev:
✅ Is added [HaruNeko](https://docs.hakuneko.download/user-manual/)
🚫 Will not be added to HakuNeko

PR #7141, opened 2024-06-02
fixed FRscan

flochocinco:
Hello,

I fixed FRScans.mjs. fixing issue #5864 

Note: I may contribute to other connectors, especially crunchyscan. Could you make me contributor ?

Note 2: I'm a newbie in JS, I'm pretty sure  _getPages(chapter) may be written in a better way

### DIFF
--- a/src/web/mjs/connectors/FRScans.mjs
+++ b/src/web/mjs/connectors/FRScans.mjs
@@ -1,28 +1,74 @@
-import MangaReaderCMS from './templates/MangaReaderCMS.mjs';
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
 
-export default class FRScan extends MangaReaderCMS {
+export default class FRScan extends Connector {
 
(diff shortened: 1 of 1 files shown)


MikeZeDev:
Before i even review it there is a problem. You used the id (and file) of a dead website with a similar name. 
I dont think those two are the same website since they coexisted.

Check this https://github.com/manga-download/haruneko/commit/3e4e16d7f7301a7e19857cdd2e92b2d318440201.

My fellow frogs have no imagination when it comes to name their websites. Its FRSCAN or SCANVF everywhere

I'd suggest you NOT overwriting a dead connector but make a new one reusing the same ID of mine from the mentioned PR.

flochocinco:
ok, I will create a new file "fr-scan.mjs"

MikeZeDev:
Please avoid '-" in file name 
Also, well, how should i say it ? Its just a madara website, you should just apply a template

Check other madara websites :D

PR #7143, opened 2024-06-04
fixed FR-Scan with madara template

flochocinco:
Hello,

Now supporting fr-scan website using madara template. fixing issue https://github.com/manga-download/hakuneko/issues/5864


src/web/mjs/connectors/FrScan.mjs
```diff
@@ -0,0 +1,31 @@
+import WordPressMadara from './templates/WordPressMadara.mjs';
+
+export default class FrScan extends WordPressMadara {
+
+    constructor() {
+        super();
+        super.id = 'fr-scan';
```
MikeZeDev, line 7: dont put '-' in id. Bad idea. 

use the same than me https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/FRTiretScan.ts, for portability: `frtiretscan` (i should have used dash and not tiret, blame my baguette)

src/web/mjs/connectors/FrScan.mjs
```diff
@@ -0,0 +1,31 @@
...
+        super.label = 'FR-Scan';
+        this.tags = [ 'manga', 'webtoon', 'french' ];
+        this.url = 'https://fr-scan.com/';
+        this.requestOptions.headers.set('x-referer', this.url);
+        this.language = 'fr';
+
+        this.queryTitleForURI = 'div.profile-manga div.post-title h1';
+    }
+
+    get icon(){
```
MikeZeDev, line 17: Fix all lint errors

src/web/mjs/connectors/FrScan.mjs
```diff
@@ -0,0 +1,31 @@
...
+        return '/img/connectors/fr-scan';
+    }
+
+    async _getPages(chapter) {
+        const request = new Request(new URL(`${chapter.id}`, this.url), this.requestOptions);
+        const data = await this.fetchDOM(request, 'div.page-break');
+        return data.map(element => {
+            let srcUrl = element.innerHTML;
```
MikeZeDev, line 25: any reason for this? 

If the reason is to get the image in best quality, madara Template is taking care of that.

https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/templates/WordPressMadara.mjs#L146

MikeZeDev (changes requested):
* Change id
* Fix lint errors
* Remove getpages override or explain why

src/web/mjs/connectors/FrScan.mjs
flochocinco: super function doesn't retrieve image (null), this is the only I achieved to get Image. Probably not the best one :) I'll try on my side with your input

MikeZeDev: I see. 

image

Suggestion : 

* Use a proper CSS selector for images that allows you to get the images inside the noscript tag only, because their src attribute got the good url. You just have to override this.queryPages.

or 

* override _getpages, get only images with [data-lazy-src] attribute (using an appropriate CSS selector), and return that attribute value.

27136773:
Suggestion to close PR after **Tuesday, 27 August 2024** (12 weeks) if not approved by a reviewer.

PR #7182, opened 2024-06-13
Use Node 20 for CI

MikeZeDev:
Use last puppeteer version that support node 16. This will fix CI and let us use linter results.

image

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install NodeJS
         uses: actions/setup-node@v1
         with:
-          node-version: 16
+          node-version: 20
       - name: Install NPM Packages
         run: npm install
       - name: Lint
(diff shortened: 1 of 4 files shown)


27136773:
Any reason for sticking to `Node 16` instead of `Node 20`?
https://github.com/manga-download/hakuneko/blob/2d4a87298ec4b33419d934577fccb98328ecf4a1/.github/workflows/ci-pr.yml#L30
https://github.com/manga-download/hakuneko/blob/2d4a87298ec4b33419d934577fccb98328ecf4a1/.github/workflows/continuous-integration.yml#L31
https://github.com/manga-download/hakuneko/blob/2d4a87298ec4b33419d934577fccb98328ecf4a1/.github/workflows/continuous-deployment.yml#L25

MikeZeDev:
I see no problem upgrading to 20, except it hasnt been tested. Not a big deal its just CI 🤷 
The goal is to fix CI so we can still have lint checks.

Gotta give 20 a try

27136773:
> I see no problem upgrading to 20, except it hasnt been tested. Not a big deal its just CI 🤷 The goal is to fix CI so we can still have lint checks.
> 
> Gotta give 20 a try

I'm using `Node 20.12` and have no issues so far with developing HakuNeko

27136773 (approved):

PR #7196, opened 2024-06-17
Aniwave : fix checksums & getting playlist

MikeZeDev:
1) Fix Aniwave itself
2) Fix Filemoon video plugin
3) Properly handle WAF anti scrapper


src/web/mjs/videostreams/FileMoon.mjs
```diff
@@ -7,9 +7,9 @@ export default class FileMoon {
 
     async getPlaylist() {
         const script = `
-            new Promise((resolve, reject) => {
+            new Promise(resolve => {
                 setTimeout(() => {
-                      resolve(window.videop.hls.url); 
+                      resolve(videop.getConfig().sources[0].file);
```
27136773: Should `try ... catch` the non-synchronous callback and reject the promise in case of an error

27136773 (approved):

PR #7219, opened 2024-06-28
add connector mikoroku

akn19:


MikeZeDev:
This is totally copy pasted from a single connector where i had to use hacky stuff because the website was crappy.
Not sure it is anymore. 

Was it tested? 
In case it works try to remove the hacky stuff and see if it works again. 

I ported it to Haruneko like this, maybe you can find some inspiration here : https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/NextScan.ts ?

akn19:
I've cleaned up the code. The code has been tested on several titles and there are no issues.

src/web/mjs/connectors/Mikoroku.mjs
```diff
@@ -0,0 +1,80 @@
...
+
+    async _getChapters(manga) {
+        let uri = new URL(manga.id, this.url);
+        let request = new Request(uri, this.requestOptions);
+        let dom = (await this.fetchDOM(request, 'body'))[0];
+        let data = [...dom.querySelectorAll('script')];
+
+        const clwdrun = data.filter(el => el.text.trim().startsWith('clwd.run'));
+        let mangaid= '';
+        if (clwdrun.length > 0) {
```
MikeZeDev: Do we really need to check for 'clwd.run' AND div#epx? 

Isn't the check for " 'clwd.run'" enough? Thats what i was implying when i said "simplify the script"

Edit : apparently `clwd` variable contains chapters list, perhaps its easier to exploit this by injecting a script (using this.fetchUI? 

image
```

    async _getChapters(manga) {
        let request = new Request(new URL(manga.id, this.url), this.requestOptions);
        const mangaid = await Engine.Request.fetchUI(request, 'clwd.settings.cat');

        request = new Request(new URL('/feeds/posts/default/-/'+mangaid+'?orderby=published&alt=json&max-results=9999', this.url), this.requestOptions);
        const { feed } = await this.fetchJSON(request);

        const chapterslist = feed.entry.map(entry => {
            const goodLink = entry.link.find(link => link.rel === 'alternate');
            return {
                id: this.getRootRelativeOrAbsoluteLink(goodLink.href, request.url),
                title: goodLink.title.replace(manga.title, '').trim()
            };
        }).filter(chap => chap.id != manga.id);
        return chapterslist;
    }
```

MikeZeDev (changes requested):
Perhaps _getChapters can be simplified?

27136773:
Suggestion to close PR after **Friday, 20 September 2024** (12 weeks) if not approved by a reviewer.

src/web/mjs/connectors/Mikoroku.mjs
akn19: done

MikeZeDev (approved):

PR #7275, opened 2024-07-21
ReadComicOnline.li: fix website

peternoordijk:
This pull request updates the script to get image URLs as the old JS 'lstImages' variable is removed. It also catches the error instead of failing with a timeout.

### DIFF
--- a/src/web/mjs/connectors/KissComic.mjs
+++ b/src/web/mjs/connectors/KissComic.mjs
@@ -56,7 +56,11 @@ export default class KissComic extends Connector {
             new Promise((resolve, reject) => {
                 setTimeout(() => {
                     try {
-                        resolve(Array.from(document.querySelectorAll('#divImage img')).map(img => img.src));
+                        resolve(
+                            Array.from(document.querySelectorAll('#divImage img'))
+                                .map(img => (img.src || '').replace(/=s\d+/, '=s0'))
+                                .filter(Boolean)
+                        )
                     } catch (err) {
                         reject(err);
                     }
--- a/src/web/mjs/connectors/KissComic.mjs
+++ b/src/web/mjs/connectors/KissComic.mjs
@@ -58,7 +58,7 @@ export default class KissComic extends Connector {
                     try {
                         resolve(
                             Array.from(document.querySelectorAll('#divImage img'))
-                                .map(img => (img.src || '').replace(/=s\d+/, '=s0'))
+                                .map(img => (img.src || '').replace(/=s\\d+/, '=s0'))
                                 .filter(Boolean)
                         )
                     } catch (err) {
--- a/src/web/mjs/connectors/KissComic.mjs
+++ b/src/web/mjs/connectors/KissComic.mjs
@@ -55,7 +55,11 @@ export default class KissComic extends Connector {
         const script = `
             new Promise((resolve, reject) => {
                 setTimeout(() => {
-                    resolve(lstImages);
+                    try {
+                        resolve(Array.from(document.querySelectorAll('#divImage img')).map(img => img.src));
+                    } catch (err) {
+                        reject(err);
+                    }
                 }, 1000);
             });
         `;


src/web/mjs/connectors/KissComic.mjs
```diff
@@ -55,7 +55,11 @@ export default class KissComic extends Connector {
         const script = `
             new Promise((resolve, reject) => {
                 setTimeout(() => {
-                    resolve(lstImages);
+                    try {
+                        resolve(Array.from(document.querySelectorAll('#divImage img')).map(img => img.src));
```
27136773: Consider to get full size images for blogspot links by e.g., replacing the [=s1600](https://2.bp.blogspot.com/pw/AP1GczPTI9j61LUiXFovnXAEq14c3TWtkCxDo2axbgltCR6hcflCjO8QTnYWlIT__6n_YgFgBsRhzHe0Na9BmOhB9tO7he_fRcDuEJSTc57WNDVRMq38oAI=s1600?rhlupa=NDYuMTI2LjIuMTk1LjcvMjEvMjAyNCAzOjIxOjUxIFBN&rnvuka=TW96aWxsYS81LjAgKE1hY2ludG9zaDsgSW50ZWwgTWFjIE9TIFggMTBfMTVfNykgQXBwbGVXZWJLaXQvNTM3LjM2IChLSFRNTCwgbGlrZSBHZWNrbykgQ2hyb21lLzEyNi4wLjAuMCBTYWZhcmkvNTM3LjM2IEVkZy8xMjYuMC4wLjA=) end of the URL pathname with [=s0](https://2.bp.blogspot.com/pw/AP1GczPTI9j61LUiXFovnXAEq14c3TWtkCxDo2axbgltCR6hcflCjO8QTnYWlIT__6n_YgFgBsRhzHe0Na9BmOhB9tO7he_fRcDuEJSTc57WNDVRMq38oAI=s0?rhlupa=NDYuMTI2LjIuMTk1LjcvMjEvMjAyNCAzOjIxOjUxIFBN&rnvuka=TW96aWxsYS81LjAgKE1hY2ludG9zaDsgSW50ZWwgTWFjIE9TIFggMTBfMTVfNykgQXBwbGVXZWJLaXQvNTM3LjM2IChLSFRNTCwgbGlrZSBHZWNrbykgQ2hyb21lLzEyNi4wLjAuMCBTYWZhcmkvNTM3LjM2IEVkZy8xMjYuMC4wLjA=)

peternoordijk: Done, but I honestly think that was quite unnecessarily as the URLs are already set to `=s0` since the `quality` parameter has been set to `hq`. You can check it out on a page like [this one](https://readcomiconline.li/Comic/Batman-Three-Jokers/Issue-1?id=175017&readType=1&quality=hq)

MikeZeDev: I wonder if 1 second wait is enough for big chapters or full comics issues.

peternoordijk: Yes, it is enough. Just to be clear: this script was working fine until a month or so ago. I’ve used it for lots of comics, and I used this updated version for even more. This PR is just fixing the issue which made it fail. Also, the one second wait is not to download the whole chapter. All that is needed is just the HTML of the page to be loaded. Honestly I don’t know what the one second is for exactly, I can imagine it might still work if removed. But still: I don’t feel like changing the code which seemed to work without a problem

27136773 (approved):

MikeZeDev:
I am afraid but i still got problems with this code

image

MikeZeDev:
Using that script i still get 0 pages returned on many chapters of random comics for no obvious reason 🤔 
No promise error, no rejection, nothing in console, just [0] pages returned.

Edit : forget about it. Some chapters may trigger cloudflare and thats the cause  , it cant be helped.

peternoordijk:
@MikeZeDev yikes, you're right about that. For some reason I didn't get the cloudflare pages before. One thing I did notice though is that one selector is not correct: `meta[http-equiv="refresh"][content*="="]` should be `meta[http-equiv="refresh"][content]` as the element we're looking for looks like this: `<meta http-equiv="refresh" content="390">`. This change doesn't quite fix the complete scenario though...

MikeZeDev:
Dont bother, Hakuneko cant pass CF captcha anyway.

PR #7302, opened 2024-08-03
Update ColoredManga connector

OkabeZeroLink:
Update the URL from "coloredmanga.com" to "coloredmanga.net"


MikeZeDev (changes requested):
ColoredManga.net works totally differently than ColorandManga.com,
An Url change wont do.

Have you even tested it? I guess not

MikeZeDev:
It doest work at all and it wont work, websites are totally different. Feel free to reopen if with tested code next time

PR #7307, opened 2024-08-05
CuuTruyen

### LINKED ISSUE #5226: [Site Request] cuutruyen

beer-psi:
Take two, with 100% less WASM blob now that I figured out the descrambling mechanism.

Closes #5226


src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+        super.label = 'Cứu Truyện';
+        this.tags = ['manga', 'vietnamese'];
+        this.url = 'https://cuutruyen.net';
+    }
+
+    async _getMangaFromURI(uri) {
+        const mangaid = uri.href.match(/\/mangas\/([0-9]+)/)[1];
+        const req = new URL('/api/v2/mangas/' + mangaid, this.url);
+        const request = new Request(req, this.requestOptions);
+        const data = await this.fetchJSON(request);
```
MikeZeDev: Since we only use the name from the api, perhaps we can directly affect it? 

```
        const { data : {name } } = await this.fetchJSON(request);
        return new Manga(this, mangaid, name.trim());
```

src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+
+    _getMangasFromPage(data) {
+        return data.data.map((c) => ({
+            id: c.id,
+            title: c.name.trim(),
+        }));
+    }
+
+    async _getChapters(manga) {
+        const uri = new URL('/api/v2/mangas/' + manga.id + '/chapters', this.url);
```
MikeZeDev: For consistency sake, maybe we could use interpolation on every requests? 

const uri = new URL(\`/api/v2/mangas/${manga.id}/chapters\`, this.url);

src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchJSON(request);
+        return data.data
+            .filter((c) => c.status === 'processed')
```
MikeZeDev: Please avoid shortening variable names. 'chapter' is fine .
I suppose data.data is an array of chapter, just do  const { data } then return data.filter ....

src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+                }
+
+                return { id: c.id, title };
+            });
+    }
+
+    async _getPages(chapter) {
+        const uri = new URL('/api/v2/chapters/' + chapter.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchJSON(request);
```
MikeZeDev: If we are just using pages, we may as well do 

`        const { data : { pages }} = await this.fetchJSON(request);`

You can chain .filter and .map after

```
        return pages
            .filter(filter((image) => image.status === 'processed')
            .map((image) => {

```

src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+        return pages.map((image) => {
+            return this.createConnectorURI({
+                url: image.image_url,
+                drmData: image.drm_data,
+            });
+        });
+    }
+
+    async _handleConnectorURI(payload) {
+        const resp = await fetch(payload.url, {
```
MikeZeDev: resp is a bit short, response is fine.

src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+            redirect: 'follow',
+            credentials: 'same-origin',
```
MikeZeDev: Are those parameters necessary? fetch default behavior is follow, and ive tested without it, without credentials cache and changing accept. Its still working. 

Perhaps just the referrer is enough, and just one. 
X-referrer is useless in the fetch context and is used when we use Request iirc

src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+    async _getPages(chapter) {
+        const uri = new URL('/api/v2/chapters/' + chapter.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchJSON(request);
+
+        if (data.data.pages.some((image) => image.status !== 'processed')) {
+            throw new Error('This chapter is still processing, please try again later.');
+        }
+
+        const pages = data.data.pages.filter((image) => image.status === 'processed');
```
MikeZeDev: Is there a need to filter here ? Because of the previous check, wont the image all be 'processed' ?

MikeZeDev (changes requested):
Minor changes requested.

Btw, nice work figuring out the image encryption.

src/web/mjs/connectors/CuuTruyen.mjs
```diff
@@ -0,0 +1,140 @@
...
+        const data = await this.fetchJSON(request);
+        const pages = data._metadata.total_pages;
+
+        const mangaList = this._getMangasFromPage(data);
+
+        for (let page = 2; page <= pages; page++) {
+            const uri = new URL(`/api/v2/mangas/recently_updated?page=${page}&per_page=30`, this.url);
+            const request = new Request(uri, this.requestOptions);
+            const data = await this.fetchJSON(request);
+            const mangas = await this._getMangasFromPage(data);
```
MikeZeDev: _getMangasFromPage is not async, i guess there is no need for await.

MikeZeDev (approved):

TraLeeee:
@beer-psi Sorry for asking this silly question?! But how you know the key is "_3141592653589793_" to decode the xor cipher? How can i find this key?

PR #7309, opened 2024-08-06
Update comment from twitter.com to x.com

KyeOnDiscord:


27136773:
This website no longer exists, it was killed by Sony after acquisition of Crunchyroll

PR #7386, opened 2024-09-05
added new connector 'Hattori Manga'

ghost:


src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+        super.label = 'Hattori Manga';
+        this.tags = ['webtoon', 'turkish'];
+        this.url = 'https://www.hattorimanga.net';
+        this.links = {
+            login: 'https://www.hattorimanga.net/giris'
+        };
+    }
+
+    async _getMangas() {
+        const mangas = [];
```
MikeZeDev, line 16: Prefer a simpler approach

```js
   async _getMangas() {
        const mangaList = [];
        }
        return mangaList;
    }

    .....
    }
```

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            } else {
+                hasMorePages = false;
+            }
+        }
+
+        return mangas;
+    }
+
+    async _getChapters(manga) {
+        manga.id = manga.id.replace('manga/', '');
```
MikeZeDev, line 40: dont "replace"

Make sure you keep only the needed part when you gather mangas.

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+        let allChapters = [];
+        let page = 1;
+        const lastPage = await this._getLastPage(manga.id);
+
```
MikeZeDev, line 44: Use this.fetchJSON. you can even use it to populate many variables directly
```js

const chapterList= [];

for (let page = 1, run = true; run = false; page ++ {

 const { chapters, lastPage } = await this.fetchJSON(.....);
 chapterList.push(... chapters.map());
 run = lastPage != page;
}

return chapterList;
```

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            } catch (error) {
+                console.error('Hata:', error);
+                break;
+            }
+        }
+        return allChapters;
+    }
+
+    async _getLastPage(mangaId) {
+        let uri = `${this.url}/load-more-chapters${mangaId}?page=1`;
```
MikeZeDev, line 67: useless

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            return veri.lastPage;
+        } catch (error) {
+            console.error('Hata:', error);
+            return 1;
+        }
+    }
+
+    async _getPages(chapter) {
+        try {
+            // Verilen chapterUrl'den HTML i�eri�ini al�r
```
MikeZeDev, line 80: do not use fetch directly. We have this.fetdhDOM for that

Moreover, what are you doing? 

chapter.id, like manga.id are NEVER supposed to be full flegged url.
We only save relative paths, or when the website allows it, the identifier used in the own website api.

 If you try to  ``` await fetch(chapter.id); ``` i dont see how its supposed to work.

At least, if you save chapter PATH as chapter.id (something like /manga/mymanga/chapterXX" you should create a real url using new URL() and this.url.

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            const response = await fetch(chapter.id);
+            if (!response.ok) {
+                throw new Error('Network response was not ok');
```
MikeZeDev, line 83: we dont care. Network error ill be thrown in console without that and thats what we want.
use Fetchdom by all means.

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            }
+            const html = await response.text();
+
+            // HTML i�eri�i bir DOM nesnesine d�n���r
+            const parser = new DOMParser();
+            const doc = parser.parseFromString(html, 'text/html');
```
MikeZeDev, line 89: The goal of using this.Fetchdom is to get the nodes we wants as an array. use it and remove all that.
One particular thing : in Hakuneko context, CSS selector for img is "source", not "img" .

Use "div.image-wrapper source".

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            // CSS se�iciyi kullanarak t�m resim elementlerini se�me
+            const imgElements = doc.querySelectorAll('div.image-wrapper img');
+
+            // Resim URL'leri bir diziye d�n���r
+            const imageUrls = Array.from(imgElements).map(img => `${this.url}${img.getAttribute('data-src') || img.src}`);
+            console.log(imageUrls);
+
+            return imageUrls;
+        } catch (error) {
+            console.error('Error fetching the pages:', error);
```
MikeZeDev, line 100: Remove console log
No need for a try catch

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            return [];
+        }
+    }
+
+    // async i�levi fetchMangaPages bir s�n�f y�ntemi olmal�d�r
```
MikeZeDev, line 105: If you really want to comment, dont do it in turkish. No offense but it wont really help in that case.

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+    async fetchMangaPages(manga) {
```
MikeZeDev, line 106: whats the purpose of this function?

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    async _getMangas() {
+        const mangas = [];
+        let page = 1;
+        let hasMorePages = true;
+
+        while (hasMorePages) {
+            let request = new Request(`${this.url}/manga?page=${page}`, this.requestOptions);
+            let data = await this.fetchDOM(request, 'div.manga-card h5 a'); // CSS se�icisini �zelle�tirir
```
MikeZeDev, line 22: we really dont need comment for trivial things :)

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+
+            if (data.length > 0) {
```
MikeZeDev, line 24: No need for that. Map will just do nothing after in case its empty.

src/web/mjs/connectors/HattoriManga.mjs
```diff
@@ -0,0 +1,117 @@
...
+            const html = await response.text();
+
+            // HTML i�eri�i bir DOM nesnesine d�n���r
+            const parser = new DOMParser();
+            const doc = parser.parseFromString(html, 'text/html');
+
+            // CSS se�iciyi kullanarak t�m resim elementlerini se�me
+            const imgElements = doc.querySelectorAll('div.image-wrapper img');
+
+            // Resim URL'leri bir diziye d�n���r
```
MikeZeDev, line 94: no need to comments obvious things

MikeZeDev (changes requested):
* A lot a things to changes
* This need an icon. 64x64 px, png, no extension to be added in /imgs/connectors.
* This lacks clipboard support. Look for _getMangaFromURI

27136773:
Suggestion to close PR after **Friday, 29 November 2024** (12 weeks) if not approved by a reviewer.

PR #7388, opened 2024-09-09
SushiScans : update manga path

aboudiom59:
Just the link that changed it's '/catalogue/list-mode/' instead of '/manga/list-mode/'


MikeZeDev:
Wont work anyway because of cloudflare, but PR is valid.

MikeZeDev (approved):

vinlandsagascan:
Hello, 
is it working with this website, because the version I have it's not even shows mangas ?
thanks

PR #7504, opened 2024-10-19
fix title fetch and image fetch in hitomi.la

noname2noname:
Two observations:
- for some reason a direct DOM fetch from the request to get the title does not work (anymore). This is not related to the fetch but rather to the requested page itself, as the title element to be just gone (reproducable in curl). The suspicion I have is that there might be some javascript run afterwards that populates the title instead, hence resorting to js injection to get the title directly (unfortunate, but better than completely broken).
- the source code on the site for fetching images has changed, updating to be consistent to what is on the site.

### DIFF
--- a/src/web/mjs/connectors/Hitomi.mjs
+++ b/src/web/mjs/connectors/Hitomi.mjs
@@ -13,11 +13,21 @@ export default class Hitomi extends Connector {
     }
 
     async _getMangaFromURI(uri) {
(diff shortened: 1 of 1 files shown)


MikeZeDev:
You could also have watched how its done in Haruneko https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/Hitomi.ts. Thanks for that contribution <3

MikeZeDev (approved):

PR #7542, opened 2024-11-07
Fix Niconico Manga

viliml:


MikeZeDev (changes requested):
Dont submit PR if you havent tested them yourself. Because it doesnt work at all, NicoNico has an API .
There is no way an url change would be enough.

Closing this in a week if nothing is done.

viliml:
> Dont submit PR if you havent tested them yourself. Because it doesnt work at all, NicoNico has an API . There is no way an url change would be enough.
> 
> Closing this in a week if nothing is done.

@MikeZeDev Of course I tested it. I re-test it every week as I download new chapters from Niconico on my personal-use fork of Hakuneko where I apply all the patches that you reject.

PR #7720, opened 2025-02-03
Add Weeb Central connector/website

### LINKED ISSUE #7407: [Site Request] weebcentral | weeb central
**Name of the website**
weebcentral

- Site: https://weebcentral.com/
- Manga List: https://weebcentral.com/search
- Manga example: https://weebcentral.com/series/01J76XYFXF05BB2S3MQM4F59H9/Cheat-Mode-Farming-in-Another-World
- Chapter online viewer example: https://weebcentral.com/chapters/01J7FKG87XNWX1P0CBY8BVNQKT

**Languages**
English

**Website relationship*

**Additional details**

kiseki999:
Add weebcentral.com connector/website into HakuNeko.

I'm not that familiar with github (or git in general). So, feel free to correct me if I'm wrong.

Closes #7407


src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,84 @@
+import Connector from '../engine/Connector.mjs';
+
+export default class WeebCentral extends Connector {
+
```
MikeZeDev, line 5: Since WeebCentral is having issues with manga list its very important to implement clipboard feature.

```js
    // Recommend to overwrite the manga extractor for copy & paster URLs ...
    async _getMangaFromURI(uri) {
        const [ title ] = await this.fetchDOM(request, 'meta[property="og:title"]');
        return new Manga(this, uri.pathname, title.content.trim());
    }
```

src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,84 @@
...
+        const limit = 32;
+        let offset = 0;
+        let mangaList = [];
+        let run = true;
+
+        while (run) {
+            const uri = new URL(`${this.url}/search/data?display_mode=Minimal+Display&limit=${limit}&offset=${offset}`);
+
+            const request = new Request(uri, this.requestOptions);
+            const data = await this.fetchDOM(request, 'article.bg-base-300');
```
MikeZeDev: ```article.bg-base-300``` doesnt seems to be a good CSS selector. 'bg-base-300' seems to be autogenerated class name based on style.  Better doing it like in Haruneko https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/WeebCentral.ts#L14

src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+    constructor() {
+        super();
+        super.id = 'weebcentral';
+        super.label = 'Weeb Central';
+        this.tags = [ 'manga', 'manhua', 'manhwa', 'english' ];
+        this.url = 'https://weebcentral.com';
+    }
+
+    async _getMangas() {
```
MikeZeDev: Sure, that code works and its cool. We'd rather doing it like this : 

```js 
    // A variation that will iterate through a multi-page manga list (with abort condition)
    /*
    async _getMangas() {
        const mangaList = [];
        }
        return mangaList;
    }
        const uri = new URL('/manga-list/' + page, this.url);
        const data = await this.fetchDOM(request, 'div.mangas > a');
        return data.map(element => {
            return {
                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
                title: element.text.trim()
            };
        });
    }
```
Please recode it accordingly <3. Using similar coding style is important after all.

It will be way easier to read.

src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+            // Next page
+            offset += limit;
+        }
+
+        return mangaList;
+    }
+
+    async _getChapters(manga) {
+        const mangaChapterListUrl = manga.id.replace(/\/[^/]+$/, '/full-chapter-list');
```
MikeZeDev: why replace? In case you fear someone pasted a bad link, you just need to control the link format using 

```js
    // May overwrite the default copy & paste matcher in case to support various URL patterns with this connector ...
    canHandleURI(uri) {
        return /https?:\/\/(?:www\.)?template.net/.test(uri.origin);
    }
```

Id from manga list should be fine, providing you used the CSS selector i used in Haruneko.

src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,84 @@
...
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: element.querySelector('span.grow span').textContent.trim(),
+                language: ''
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        const uri = new URL(`${chapter.id}/images?is_prev=False&current_page=1&reading_style=long_strip`, this.url);
```
MikeZeDev: Where does this `images`  comes from? 

Just execute the script on chapter page, using Haruneko selector for code consistency.

```js
  const uri = new URL(chapter.id, this.url);
```

https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/WeebCentral.ts#L11

src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,84 @@
...
+                        resolve([...document.querySelectorAll('section img')].map(img => img.src));
+                    } catch(error) {
+                        reject(error);
+                    }
+                }, 2500);
+            });
+        `;
+
+        const data = await Engine.Request.fetchUI(request, script);
+        return data.map(element => this.createConnectorURI(this.getAbsolutePath(element, request.url)));
```
MikeZeDev: `createConnectorURI` is meant to be used when we need to use referer to download images.

In which case you usually needs to write custom handler, with referer

```js

        const request = new Request(payload, this.requestOptions);
        request.headers.set('x-referer', this.url);
        let data = await response.blob();
        data = await this._blobToBuffer(data);
        this._applyRealMime(data);
        return data;
    }
```

If not needed, dont use createConnectorURI and just return image url.

src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,84 @@
...
+                language: ''
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        const uri = new URL(`${chapter.id}/images?is_prev=False&current_page=1&reading_style=long_strip`, this.url);
+        const request = new Request(uri, this.requestOptions);
+
+        let script = `
```
MikeZeDev: use const instead of let (script is a constant after all)

MikeZeDev (changes requested):
Good job. Some issues needs to be addressed tho.

Also, you can write in main message of the PR "Closes #7407 " so it will auto close weebcentral issue when is done :)

src/web/mjs/connectors/WeebCentral.mjs
kiseki999: I don't fully understand this "clipboard feature" yet, so I'm just including this code as is.

MikeZeDev: *user paste manga url
* function extracts manga title from manga page. Title comes from a meta tag in the html.
* return a new Manga with url pathname as ID ('/series/x2dqsdsd4q3sd4/sd5qs4d5qs4d')

src/web/mjs/connectors/WeebCentral.mjs
```diff
@@ -0,0 +1,77 @@
...
+        super();
+        super.id = 'weebcentral';
+        super.label = 'Weeb Central';
+        this.tags = [ 'manga', 'manhua', 'manhwa', 'english' ];
+        this.url = 'https://weebcentral.com';
+    }
+
+    async _getMangaFromURI(uri) {
+        const request = new Request(uri, this.requestOptions);
+        const [ title ] = await this.fetchDOM(request, 'meta[property="og:title"]');
```
MikeZeDev, line 16: image

Remove the junk in title.

[L8](https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/WeebCentral.ts#L8)

MikeZeDev (changes requested):
Nearly done. Fix pasted manga title and its good to go

src/web/mjs/connectors/WeebCentral.mjs
kiseki999: I'm sorry, I've never used this "clipboard feature" before, so I didn't know how to use it.
I've re-read the guide [here](https://github.com/manga-download/hakuneko/wiki/Developer-Website-Tutorial#provide-copy-and-paste-support) to understand it.

MikeZeDev (approved):
Thanks for your contribution <3

PR #7784, opened 2025-02-27
Create harimanga.mjs

onkey111:


src/web/mjs/connectors/harimanga.mjs
```diff
@@ -0,0 +1,16 @@
+import WordPressMadara from './templates/WordPressMadara.mjs';
+
+export default class HARIMANGA extends WordPressMadara {
+
+    constructor() {
+        super();
+        super.id = 'HARIMANGA';
```
MikeZeDev, line 7: * id must be lowercase
* class name & file name should match. Use normal case i.e Harimanga
* Same for label, you can use normal case.
* Please add an icon.  64x64 px png

MikeZeDev (changes requested):

MikeZeDev:
Suggestion : closing this on April the 1st if no reaction from contributor.

PR #7802, opened 2025-03-06
Yanmaga: Fix error handling and restore _getMangaFromURI

clfws9:
If a user tries to download a chapter that requires rental or login, an exception is thrown in the `_getPageList` function.
However, because the exception is not handled, there is no indication in the UI that the download has failed.
I have fixed this by handling the exception.

Also, since switching from `ComiciViewer` to `SpeedBinb` (https://github.com/manga-download/hakuneko/commit/e93e7c1744ee7713a1024aff6271dc3a4fe95a44), copy and paste support seems to be broken.
I have restored the `_getMangaFromURI` function from the following part.
https://github.com/manga-download/hakuneko/blob/5f8979d8453a7570c5f01b797681811f71c630fd/src/web/mjs/connectors/Yanmaga.mjs#L23-L29

**To Reproduce**

1. URL of manga: https://yanmaga.jp/comics/%E3%81%BF%E3%82%87%E3%83%BC%E3%81%A1%E3%82%83%E3%82%93%E5%85%88%E7%94%9F%E3%81%AF%E3%81%8B%E3%81%8F%E8%AA%9E%E3%82%8A%E3%81%8D
2. When you paste the URL, the manga title is displayed as `Manga #24B9`. The correct title is `みょーちゃん先生はかく語りき`.
3. Name of chapter:
    - Chapter that requires rental: `第７８限　シンメトリーフェティシズム`
    - Chapter that requires login: `第４限　バズ・レッスン！`
5. Copy any relevant error message that appears in a popup or (advanced) within the console (press F12)
    ```
   Uncaught (in promise) Error: You need to login to see 第７８限　シンメトリーフェティシズム
       at Yanmaga.mjs:63
   (anonymous) @ Yanmaga.mjs:63
   Promise.then (async)
   _getPageList @ Yanmaga.mjs:57
   (anonymous) @ Chapter.mjs:111
   Promise.then (async)
   getPages @ Chapter.mjs:109
   downloadPages @ DownloadJob.mjs:72
   processQueue @ DownloadManager.mjs:49
   
   Uncaught (in promise) Error: You need to login to see 第４限　バズ・レッスン！
       at Yanmaga.mjs:63
   (anonymous) @ Yanmaga.mjs:63
   Promise.then (async)
   _getPageList @ Yanmaga.mjs:57
   (anonymous) @ Chapter.mjs:111
   Promise.then (async)
   getPages @ Chapter.mjs:109
   downloadPages @ DownloadJob.mjs:72
   processQueue @ DownloadManager.mjs:49
   ```

### DIFF
--- a/src/web/mjs/connectors/Yanmaga.mjs
+++ b/src/web/mjs/connectors/Yanmaga.mjs
@@ -58,9 +58,6 @@ export default class Yanmaga extends SpeedBinb {
     }
 
     _getPageList( manga, chapter, callback ) {
-        if (chapter.id.includes('/sign-up')) {
-            throw new Error(`You need to login to see ${chapter.title}`);
-        }
         const uri = new URL(chapter.id, this.url);
         fetch(uri)
             .then(response => {
--- a/src/web/mjs/connectors/Yanmaga.mjs
+++ b/src/web/mjs/connectors/Yanmaga.mjs
@@ -48,21 +48,24 @@ export default class Yanmaga extends SpeedBinb {
         return Engine.Request.fetchUI(request, chapterScript, 10000);
     }
 
-    _getPageList( manga, chapter, callback ) {
-        if (chapter.id.includes('/sign-up')) {
-            throw new Error(`You need to login to see ${chapter.title}`);
+    async _getPageList(manga, chapter, callback) {
+        try {
+            if (chapter.id.includes('/sign-up')) {
+                throw new Error(`You need to login to see ${chapter.title}`);
+            }
+            const uri = new URL(chapter.id, this.url);
+            const response = await fetch(uri);
+            if (response.redirected) {
+                const newurl = new URL(response.url);
+                return super._getPageList(manga, { id: newurl.pathname + newurl.search }, callback);
+            }
+            if (!uri.searchParams.get('cid')) {
+                throw new Error(`You need to login to see ${chapter.title}`);
+            }
+            return super._getPageList(manga, chapter, callback);
+        } catch (error) {
+            console.error(error, chapter);
+            callback(error, undefined);
         }
-        const uri = new URL(chapter.id, this.url);
-        fetch(uri)
-            .then(response => {
-                if (response.redirected) {
-                    const newurl = new URL(response.url);
-                    return super._getPageList(manga, { id: newurl.pathname+newurl.search }, callback);
-                }
-                if (!uri.searchParams.get('cid')) {
-                    throw new Error(`You need to login to see ${chapter.title}`);
-                }
-                return super._getPageList(manga, chapter, callback);
-            });
     }
 }
--- a/src/web/mjs/connectors/Yanmaga.mjs
+++ b/src/web/mjs/connectors/Yanmaga.mjs
@@ -57,24 +57,25 @@ export default class Yanmaga extends SpeedBinb {
         return Engine.Request.fetchUI(request, chapterScript, 10000);
     }
 
-    async _getPageList(manga, chapter, callback) {
-        try {
-            if (chapter.id.includes('/sign-up')) {
-                throw new Error(`You need to login to see ${chapter.title}`);
-            }
-            const uri = new URL(chapter.id, this.url);
-            const response = await fetch(uri);
-            if (response.redirected) {
-                const newurl = new URL(response.url);
-                return super._getPageList(manga, { id: newurl.pathname + newurl.search }, callback);
-            }
-            if (!uri.searchParams.get('cid')) {
-                throw new Error(`You need to login to see ${chapter.title}`);
-            }
-            return super._getPageList(manga, chapter, callback);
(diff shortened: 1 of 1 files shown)


src/web/mjs/connectors/Yanmaga.mjs
```diff
@@ -48,21 +57,24 @@ export default class Yanmaga extends SpeedBinb {
         return Engine.Request.fetchUI(request, chapterScript, 10000);
     }
 
-    _getPageList( manga, chapter, callback ) {
-        if (chapter.id.includes('/sign-up')) {
-            throw new Error(`You need to login to see ${chapter.title}`);
+    async _getPageList(manga, chapter, callback) {
+        try {
```
MikeZeDev: I think there is a reason why we have to use callback and .then on this legacy code, instead of async and await .

I dont wanna approve that in case its breaking something. And Ronny have other things to do now.

MikeZeDev (changes requested):

src/web/mjs/connectors/Yanmaga.mjs
```diff
@@ -57,24 +57,25 @@ export default class Yanmaga extends SpeedBinb {
...
-                throw new Error(`You need to login to see ${chapter.title}`);
-            }
-            return super._getPageList(manga, chapter, callback);
-        } catch (error) {
-            console.error(error, chapter);
-            callback(error, undefined);
+    _getPageList( manga, chapter, callback ) {
+        if (chapter.id.includes('/sign-up')) {
+            throw new Error(`You need to login to see ${chapter.title}`);
         }
```
clfws9: The exception in this part does not seem to need to be handled within the connector.
It is handled in the `Chapter` class.
https://github.com/manga-download/hakuneko/blob/2f606386b912b0aeb53a9e3a2197ce52523a2ebd/src/web/mjs/engine/Chapter.mjs#L124-L126

MikeZeDev: will check that

clfws9: I still haven't found any chapters with `/sign-up` in their ID.
I used the patch below to always throw the exception in this part to see where the exception is handled.
```diff
 src/web/mjs/connectors/Yanmaga.mjs | 2 +-
 src/web/mjs/engine/Chapter.mjs     | 1 +
 2 files changed, 2 insertions(+), 1 deletion(-)

diff --git a/src/web/mjs/connectors/Yanmaga.mjs b/src/web/mjs/connectors/Yanmaga.mjs
index 901cfb60..d79f8233 100644
--- a/src/web/mjs/connectors/Yanmaga.mjs
+++ b/src/web/mjs/connectors/Yanmaga.mjs
@@ -58,7 +58,7 @@ export default class Yanmaga extends SpeedBinb {
     }
 
     _getPageList( manga, chapter, callback ) {
-        if (chapter.id.includes('/sign-up')) {
+        if (true) {
             throw new Error(`You need to login to see ${chapter.title}`);
         }
         const uri = new URL(chapter.id, this.url);
diff --git a/src/web/mjs/engine/Chapter.mjs b/src/web/mjs/engine/Chapter.mjs
index 43164ea8..f94a62c2 100644
--- a/src/web/mjs/engine/Chapter.mjs
+++ b/src/web/mjs/engine/Chapter.mjs
@@ -122,6 +122,7 @@ export default class Chapter extends EventTarget {
                     } );
                 } )
                 .catch( error => {
+                    console.error(error);
                     callback( error, undefined );
                 } );
         }

```

The following error message appeared in the console:
```
Chapter.mjs:125 Error: You need to login to see 第４限　バズ・レッスン！
    at Yanmaga._getPageList (Yanmaga.mjs:62)
    at Chapter.mjs:111
(anonymous) @ Chapter.mjs:125
Promise.catch (async)
getPages @ Chapter.mjs:124
downloadPages @ DownloadJob.mjs:72
processQueue @ DownloadManager.mjs:49
```

Therefore, I confirmed that the exception is handled in the `Chapter` class.
Is this behaviour intended?
If not, I will change the exception to be handled in the `Yanmaga` connector.

MikeZeDev: I worked on thousand of connectors but since SpeedBinb is still using old coding pattern i am a bit puzzled. 
I dont know the intended behavior but i think the correct way is to show a messagebox.

Indeed, your last changes are quite correct in that regard.

OK. If you didnt find any chapter with /sign-up the code , lets keep your changes and remove this particular check.

MikeZeDev (changes requested):
Remove first signup check and we are fine :)

clfws9:
I have confirmed that there are no chapters with `/sign-up` in their ID among the total of 9,781 chapters available through the `Yanmaga` connector.
Therefore, I have removed this check.

MikeZeDev (approved):

PR #7816, opened 2025-03-12
EDIT: Add the modern theme as a new default theme. ~~Replace the Classic Light theme with a More modern looking one.~~

AIR200:
image
image
image
image


Sheepux:
Thanks for the time spent on this, however contributors won't have time to review this PR as we are contributing all our small free time on the full rework https://github.com/manga-download/haruneko
contribution / comments on the new UI are welcome

voting to close as not to get this rotting

MikeZeDev:
Thanks for that contribution.  I agree with sheep.

I will say 3 things : 

1) Validating theme changes are not a big deal. Its not really code, there is no need for an complicated review process . However :
 
2) What if people prefer the ancient look? It would be better adding another choice in program settings and make this an alternate classic theme.

3) And by doing that we are doing some modifications that require more testing and i doubt people will have the time to validate and i dont know immediately what "adding a new theme" implies in term of building and releasing process.

AIR200:
Thanks for taking your time replying. 

Yes, I totally agree with the fact that people might prefer the old look. However I can say two things in defense of just replacing the theme.

1. Very similar to Ken's Dark theme and the design choices in haruneko. Given how this one integrates with modern OSes, I believe most won't be opposed to it.
2. Its just easy, no new coding and testing required.

I'm not saying that you should not try to add a new theme option, but that would require new code and testing, as you said. In the end, its the hakuneko team's choice as for whether to merge this PR, reject it or create support for a new theme.

And finally, thanks for creating this amazing project❤️

AIR200:
I found out how to add this as a new theme option. And done with the new commits (using github web). Reverted the changes to classic theme.

Since its just a copy of the classic-light folder, with only the theme file replaced and adding a new entry to the settings file, I hope you will find enough time to review this. 
image
Sorry for the mess.

27136773:
Suggestion to close PR after **Wednesday, 4 June 2025** (12 weeks) if not approved by a reviewer.

PR #7823, opened 2025-03-18
add connector for toonilyme website

YurDavid:


src/web/mjs/connectors/ToonilyMe.mjs
```diff
@@ -0,0 +1,78 @@
+import Connector from '../engine/Connector.mjs';
+import Manga from '../engine/Manga.mjs';
+
+export default class ToonilyMe extends Connector {
```
MikeZeDev: This website is using Madtheme Template. Please use Madtheme template.
No need to reinvent the wheel. Check Mangabuddy for example

MikeZeDev (changes requested):

src/web/mjs/connectors/ToonilyMe.mjs
YurDavid: Okay, cool. Just asking—how do you check what template a website is using ?

src/web/mjs/connectors/ToonilyMe.mjs
```diff
@@ -0,0 +1,78 @@
...
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: chapterTitle,
+                language: ''
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(this.url + chapter.id, this.requestOptions);
+        const data = await this.fetchRegex(request, /<img[^>]+(?:src|data-src)="((?!.*\/thumb\/)[^"]+)"/g);
```
MikeZeDev: you can just use a proper CSS selector, and get the data-src attribute.

img.dataset.src

YurDavid: I decided to try again with fetchDOM and even looked into promises, but had no luck either. I realized the issue is lazy loading, but not the usual kind. The easiest method that works for me is fetchRegex. Should I close it, or should I just go ahead and push the other changes?

MikeZeDev: Image links seems to be in `data.src` attribute of images so there is no problem fetching the pages nodes with `fetchDOM`

https://github.com/manga-download/haruneko/blob/7fba8234cc4ca97b3d550e8164b6408003299650/web/src/engine/websites/Toonilyme.ts#L7

Providing, ofc, that you replace `img `tag in CSS by `source`, as stated in https://github.com/manga-download/hakuneko/blob/master/src/web/mjs/connectors/templates/%40Samples.mjs

YurDavid: oh i see learnt something new today

YurDavid: done

src/web/mjs/connectors/ToonilyMe.mjs
```diff
@@ -0,0 +1,32 @@
...
+    constructor() {
+        super();
+        super.id = 'toonilyme';
+        super.label = 'ToonilyMe Manga';
+        this.tags = ['webtoon', 'english'];
+        this.url = 'https://toonily.me';
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(this.url + manga.id, this.requestOptions);
```
MikeZeDev: Dont concat url and chapter id.

Create an url using appropriate method
```

const uri = new URL(manga.id, this.url);
```

src/web/mjs/connectors/ToonilyMe.mjs
```diff
@@ -0,0 +1,32 @@
...
+            return {
+                id: this.getRootRelativeOrAbsoluteLink(element, this.url),
+                title: chapterTitle,
+                language: ''
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        let request = new Request(this.url + chapter.id, this.requestOptions);
```
MikeZeDev: Same remark about url & id

MikeZeDev: request is a constant, uses const and not let (template is a bit old )

src/web/mjs/connectors/ToonilyMe.mjs
```diff
@@ -0,0 +1,32 @@
...
+        super.label = 'ToonilyMe Manga';
+        this.tags = ['webtoon', 'english'];
+        this.url = 'https://toonily.me';
+    }
+
+    async _getChapters(manga) {
+        let request = new Request(this.url + manga.id, this.requestOptions);
+        let data = await this.fetchDOM(request, 'ul#chapter-list > li > a');
+        return data.map(element => {
+            let titleElement = element.querySelector('strong.chapter-title');
```
MikeZeDev: use `const `instead of `let `when variable is no reassigned.

MikeZeDev (changes requested):

src/web/mjs/connectors/ToonilyMe.mjs
YurDavid: oh yeah, code quality

src/web/mjs/connectors/ToonilyMe.mjs
YurDavid: sure

src/web/mjs/connectors/ToonilyMe.mjs
YurDavid: okay

MikeZeDev:
When website comes back up, ill test this :)

MikeZeDev:
image

use image.dataset.src.

YurDavid:
woah i will update soon

src/web/mjs/connectors/ToonilyMe.mjs
```diff
@@ -0,0 +1,31 @@
...
+                title:element.querySelector(this.queryChapterTitle).textContent.trim(),
+            };
+        });
+    }
+
+    async _getPages(chapter) {
+        const uri = new URL(chapter.id, this.url);
+        const request = new Request(uri, this.requestOptions);
+        const data = await this.fetchDOM(request, 'div.chapter-image source');
+        return data.map(element => this.createConnectorURI(element.dataset.src || element.getAttribute('data-src')));
```
MikeZeDev: `element.dataset.src` and `element.getAttribute('data-src')` are the same . using both is useless. Just use `element.dataset.src` .

MikeZeDev (changes requested):

src/web/mjs/connectors/ToonilyMe.mjs
YurDavid: oh sorry didn't see this message earlier

MikeZeDev: Its fine. Can i be blunt? I have the feeling that you submit some code without testing it.

Because there is no way it would have worked without taking dataset.src instead of src. 

We are grateful and contributions are always welcome. No one expect people to submit perfect code in the first place (there is a review process after all) but testing basic functions like "getting the good pictures" should be done before submitting.

Oh and, if you consider a conversation closed, feel free to hit "Resolve conservation" :)

src/web/mjs/connectors/ToonilyMe.mjs
```diff
@@ -0,0 +1,31 @@
+import MadTheme from './templates/MadTheme.mjs';
+
+export default class ToonilyMe extends MadTheme {
+    constructor() {
+        super();
+        super.id = 'toonilyme';
+        super.label = 'ToonilyMe Manga';
```
MikeZeDev: i believe i have removed the "Manga" myself because i didnt want to bother you with that.
Please accept reviewer modifications :D

src/web/mjs/connectors/ToonilyMe.mjs
YurDavid: would keep that in mind, I tend to push things at the heat of moment,

MikeZeDev (approved):

PR #7840, opened 2025-03-22
Manhwaclub: change domain & code

YurDavid:


src/web/mjs/connectors/ManhwaClub.mjs
```diff
@@ -4,45 +4,9 @@ export default class ManhwaClub extends WordPressMadara {
 
     constructor() {
         super();
-        super.id = 'manhwaclub';
-        super.label = 'ManhwaHentai';
+        super.id = 'manhwaclubnet';
```
MikeZeDev: Dont change the id. NEVER change the id.

Or we lose Haruneko portability

https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/ManhwaClub.ts#L15

MikeZeDev (changes requested):
Please rename the icon to match mjs filename / unchanged id

src/web/mjs/connectors/ManhwaClub.mjs
YurDavid: oh okay  but we have a icon called manhwaclub already what do we do

YurDavid:
> Please rename the icon to match mjs filename / unchanged id

Okay, I have an idea. Without touching manhwaclub.mjs, I would create manhwaclubnet.mjs. This way, the icon will match the MJS filename, and manhwaclub.mjs will be restored to its original state to maintain Hakuneko's portability.

src/web/mjs/connectors/ManhwaClub.mjs
MikeZeDev: delete  old icon, replace with the new one?  I dont see any problem. 

Just never change id for good reasons. Make sure id/classes/filenames matches if possible thats all :)

MikeZeDev:
> > Please rename the icon to match mjs filename / unchanged id
> 
> Okay, I have an idea. Without touching manhwaclub.mjs, I would create manhwaclubnet.mjs. This way, the icon will match the MJS filename, and manhwaclub.mjs will be restored to its original state to maintain Hakuneko's portability.

Why so complicated? 
just replace `manhwaclub` file with new icon, keep old id in `manhwaclub.mjs` and voila.
😢

src/web/mjs/connectors/ManhwaClub.mjs
YurDavid: okay we can do that

MikeZeDev (approved):

PR #7882, opened 2025-04-19
feat(MangaPlaza): create connector

### LINKED ISSUE #4824: [Site Request] MangaPlaza
**Name of the website**
MangaPlaza

- Site: https://mangaplaza.com/
- Manga List: there is no manga list
- Manga example: https://mangaplaza.com/title/0303002306/
- Chapter online viewer example: https://reader.mangaplaza.com/speedreader/?cid=103030023060001&u0=0&u1=https%3A%2F%2Fmangaplaza.com%2Ftitle%2F0303002306%2F%3Forder%3Ddown%26content_id%3D103030023060001

[It needs a free account to see the free chapter]

**Languages**
English

**Website relationship*
It is an official English store, with a lot of originals form smaller publishers. (The parent company owns cmoa.jp) It has a subscription which unlocks many chapters, but you can buy individual chapters without a subscription. Some chapters are free, but require login. There is a 7 day free trial, it only takes paypal.

**Additional details**
SpeedBinb.VERSION
'1.6460.0001'

assortedmints:
https://mangaplaza.com/

fixes [#4824](https://github.com/manga-download/hakuneko/issues/4824)

Currently, paid chapters aren't shown to the user if they do not own it. Should this behavior be changed?


MikeZeDev:
Jesus. SpeedBinb PR. Gotta take my time reviewing it.

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,121 @@
...
+    async _getMangaFromURI(uri) {
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchDOM(req, '#_title_nm');
+        const id = uri.pathname.split('/')[2];
+        const title = this._sanitizeMangaTitle(res[0].innerText.trim());
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let mangaList = [];
```
MikeZeDev: mangalist is not reaffected, you can use const. In many places in this file.

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,121 @@
...
+        };
+        this.apiPath = '/api';
+        this.requestOptions.headers.set('x-cookie', 'mp_over18_agreement=ON');
+        this.pageNoSelector = '.pages>:nth-last-child(2)>a';
+    }
+
+    async _getMangaFromURI(uri) {
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchDOM(req, '#_title_nm');
+        const id = uri.pathname.split('/')[2];
```
MikeZeDev, line 23: if its last par of the pathname you may use at(-1). In any case, better using .at() everywhere.

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,121 @@
...
+        const title = this._sanitizeMangaTitle(res[0].innerText.trim());
```
MikeZeDev: Better using at(0)

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,121 @@
...
+        return new Manga(this, id, title);
+    }
+
+    async _getMangas() {
+        let mangaList = [];
+        const uri = new URL('/new', this.url);
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchDOM(req, this.pageNoSelector);
+        let pageCount = parseInt(res[0].text);
```
MikeZeDev: pagecount is a const

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,121 @@
...
+        return res.map(e => {
+            return {
+                id: e.pathname.split('/')[2],
+                title: this._sanitizeMangaTitleFromNewReleases(e.innerText.trim())
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        let chapterList = [];
```
MikeZeDev: Prefer using DOMparser to transform text to HTML document and perform queries on it

MikeZeDev: Perhaps you could be inspired to look at `FetchChapters` from https://github.com/manga-download/haruneko/pull/179/files#diff-5b662c18f8d292ddca0441b22d9ef939b92a9acd8bb5e3bc9b053fb54e96df95. 

I think chapter logic is a bit complicated here.  Did i miss something?

src/web/mjs/connectors/MangaPlaza.mjs
MikeZeDev: Perpahs you could exploit "genre" list, like i did here : https://github.com/manga-download/haruneko/pull/179/files#diff-5b662c18f8d292ddca0441b22d9ef939b92a9acd8bb5e3bc9b053fb54e96df95.

Ofc, make sure to dedupe manga list at the end. It would avoid all those sanitize shenanigans.

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    _sanitizeMangaTitleFromNewReleases(title) {
+        const splitTitle = title.split(' ');
+        splitTitle.splice(0, 1);
+        return this._sanitizeMangaTitle(splitTitle.join(' '));
+    }
+
+    _querySelectorFromHtmlString(htmlStr, selector) {
+        const dom = document.createElement('html');
```
MikeZeDev: Dont use that, replace by DOMparser uses.

MikeZeDev (changes requested):

src/web/mjs/connectors/MangaPlaza.mjs
assortedmints: I believe the electron version used (8.5.5) does not support `.at()` unfortunately.

MikeZeDev: Ho, right. My bad

src/web/mjs/connectors/MangaPlaza.mjs
assortedmints: Could you elaborate on which section could be simplified?

The current implementation only retrieves chapters that are accessible to the user by using the `a[href*="/reader/"]` selector; chapter links accessible to premium users have the class of `.freeBtn` but would lead elsewhere if the user isn't logged in or a premium member.

I had to construct the response url for the chapter id as the method in SpeedBinb retrieves params from chapter id. 
https://github.com/manga-download/hakuneko/blob/c37f5290140b7049e031a31ec09703744338a6f3/src/web/mjs/connectors/templates/SpeedBinb.mjs#L210

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+        this.apiPath = '/api';
+        this.requestOptions.headers.set('x-cookie', 'mp_over18_agreement=ON');
+        this.pageNoSelector = '.pages>:nth-last-child(2)>a';
+    }
+
+    async _getMangaFromURI(uri) {
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchDOM(req, '#_title_nm');
+        const id = uri.pathname.split('/')[2];
+        const title = this._sanitizeMangaTitle(res[0].innerText.trim());
```
MikeZeDev: Is there really a need for cleaning the title when using clipboard?

MikeZeDev: Nevermind. https://mangaplaza.com/title/0303010376/ ...Whats going on in their head, thats ugly.

MikeZeDev: https://mangaplaza.com/title/0404000207/. May replace all <> stuff at the beginning..
```js
title.replace(/^<.*>/, '').trim()
```

kefi4ik:
Hello, from Russia. Because of our government a lot of sites were banned. So can you change address of Desu,please? It is only site that have rear adress https://desu.work/

MikeZeDev:
1) Dont hijack unrelated PR
2) Use a VPN / change DNS, its not haku job to bypass government censorship.

src/web/mjs/connectors/MangaPlaza.mjs
assortedmints: @MikeZeDev Sorry, could I get an update on this?

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+
+    async _getMangasFromGenre(path) {
+        const uri = new URL(path, this.url);
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchDOM(req, this.pageNoSelector);
+        const pageCount = res.length != 0 ? parseInt(res[0].text) : 1;
+
+        const mangaList = [];
+        for (let i = 1; i <= pageCount; i++) {
+            mangaList.push(...await this._getMangasFromPage(uri.href, i));
```
MikeZeDev: if you reconstruct an url  in _getMangasFromPage, why passing uri.href and not the uri directly? 

I dont think there is really a need for _getMangasFromPage. Include _getMangasFromPage content in _getMangasFromGenre loop

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        return mangaList;
+    }
+
+    async _getMangasFromPage(uri, page) {
+        page = page || 1;
```
MikeZeDev: why would page be undefined? you made the proper control before :)

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+        super();
+        super.id = 'mangaplaza';
+        super.label = 'MangaPlaza';
+        this.tags = ['manga', 'english'];
+        this.url = 'https://mangaplaza.com';
+        this.baseURL = 'https://reader.mangaplaza.com/speedreader';
+        this.links = {
+            login: 'https://mangaplaza.com/login'
+        };
+        this.apiPath = '/api';
```
MikeZeDev: woudnt it be easier to define 

```ts
this.apiURl = new URL('./api', this.url);
```

and then call the endpoints like this 

```ts
        const uri = new URL('./title/content_list/`, this.apiURl);
```
rather than 

```ts
        const uri = new URL(`${this.apiPath}/title/content_list/`, this.url);
```

src/web/mjs/connectors/MangaPlaza.mjs
MikeZeDev: This still needs to be fixed

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+
+        return mangaList;
+    }
+
+    async _getMangasFromPage(uri, page) {
+        page = page || 1;
+        uri = new URL(uri);
+        uri.searchParams.set('page', page);
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchDOM(req, 'ul.listBox li div.titleName a');
```
MikeZeDev: Iff you dont plan to reuse the values, you can get rid of variables and inline stuff.
You are also free to use interpolation.
Also, uses  'request' rather than "req".
Also in many places of that file

```ts 
        const request = new Request(new URL(`${path}?page=${page}`, this.url), this.requestOptions);
        const results= await this.fetchDOM(request, 'ul.listBox li div.titleName a');
```

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+
+        return chapterList;
+    }
+
+    async _getChaptersFromPage(uri, page, mangaId) {
+        page = page || 1;
+        uri = new URL(uri);
+        uri.searchParams.set('page', page);
+
+        const req = new Request(uri, this.requestOptions);
```
MikeZeDev: combine request construction and request call in fetchjson

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+        return res.map(e => {
+            return {
+                id: e.pathname.split('/')[2],
+                title: this._sanitizeMangaTitle(e.innerText.trim())
+            };
+        });
+    }
+
+    async _getChapters(manga) {
+        const chapterList = Array.from(new Set());
```
MikeZeDev:
```ts
        const chapterList = [];
```

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+        const uri = new URL(`${this.apiPath}/title/content_list/`, this.url);
+        uri.searchParams.set('title_id', manga.id);
+
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchJSON(req);
+        const pageBtn = this._querySelectorFromHtmlStr(res.data.html_page, this.pageNoSelector);
```
MikeZeDev:
```ts
const { data: { html_page}} = await this.fetchJSON(request);
const doc = new DOMParser().parsefromString(html_page, 'text/html');
const pageBtn = doc.querySelector(this.pageNoSelector);
const pageCount = pageBtn ?  parseInt(pageBtn.text) : 1;
```
 perhaps? 

I dont see the need for a fonction with just a domparser.

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+
+    async _getChaptersFromPage(uri, page, mangaId) {
+        page = page || 1;
+        uri = new URL(uri);
+        uri.searchParams.set('page', page);
+
+        const req = new Request(uri, this.requestOptions);
+        const res = await this.fetchJSON(req);
+        
+        return this._querySelectorFromHtmlStr(res.data.html_content, '.inner_table')
```
MikeZeDev:
```ts
const { data: { html_content}} = await this.fetchJSON(request);
const doc = new DOMParser().parsefromString(html_page, 'text/html');
return [...doc.querySelectorAll('.inner_table')]
    .filter
    .map
```

src/web/mjs/connectors/MangaPlaza.mjs
```diff
@@ -0,0 +1,124 @@
...
+                };
+            });
+    }
+
+    _sanitizeMangaTitle(title) {
+        return title.replace('<Chapter release>', '').trim();
+    }
+
+    _querySelectorFromHtmlStr(htmlStr, selector) {
+        return Array.from(new DOMParser().parseFromString(htmlStr, 'text/html').querySelectorAll(selector));
```
MikeZeDev: remove the function

MikeZeDev (changes requested):
I tried to be more clear about what i meant by saying "cleaner code".
We usually combine url construction / request construction, sometimes even with the fetch call . One level or two of inlining is better for readability when there is nothing fancy. Also it removes useless variables. time to fixes those tiny things 👍

MikeZeDev (approved):
LGTM. I took the time to test others SpeedBinb websites, since the template was modified, everything runs fine.

Thanks for contribution 👍

PR #7935, opened 2025-05-17
HentaiNexus: fix getting pages

DraconicSalad:


MikeZeDev (approved):
thanks for contributing

PR #7977, opened 2025-06-07
fix 1-connector-website-not-working.yml

Mark-Cal:
body[4]: options[0]: default is not a permitted attribute
body[4]: default is not a permitted attribute

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-connector-website-not-working.yml
+++ b/.github/ISSUE_TEMPLATE/1-connector-website-not-working.yml
@@ -36,7 +36,6 @@ body:
       options:
         - label: Yes, the problem cannot be reproduced with Nightly Build
           required: false
-          default: true
   - type: dropdown
     attributes:
       label: What kind of issue are you encountering


MikeZeDev (approved):

27136773:
The purpose was that users does not blindly check / uncheck the boxes.
This PR nullifies the effect since the default is now `false`.

Mark-Cal:
> The purpose was that users does not blindly check / uncheck the boxes. This PR nullifies the effect since the default is now `false`.

I tried but you can't make checked checkboxes unless you use a different coding language

MikeZeDev:
Well it was just not working at all. People kept opening "website suggestion" instead of "website changes" because of that bugged template and it was bothering me. I kept asking myself why, until i saw that PR.

I thought that a fixed template even with that behavior would be better than a non working one, and i didnt want to disturb you for that , ronny. So that pr was convenient . And i validated it, because its harmless.

27136773:
If users are choosing the obviously mismatching `Suggest a new Website` template to report a broken website instead of the general purpose `Blank issue` template (which is just four entries below), than just close the ticket.

MikeZeDev:
> If users are choosing the obviously mismatching `Suggest a new Website` template to report a broken website instead of the general purpose `Blank issue` template (which is just four entries below), than just close the ticket.

Thats what i ma doing but still i cant tell them "use the proper template" if it doesnt exists.

PR #8019, opened 2025-06-30
Remove some dead websites

thiraphat98:
Removed dead connectors: bakaman.net and Biamam.
Both websites are no longer accessible, and their connectors were no longer functional.
This PR removes the related `.mjs` files and icon from the project.


MikeZeDev (approved):

MikeZeDev (changes requested):
* you deleted bakaman icon and not the MJS
* you deleted Biamam mjs but not the icon

thiraphat98:
Oh sorry, I must have deleted the wrong file.

MikeZeDev (approved):

PR #8050, opened 2025-07-12
Add Connector MangaDistrict

DraconicSalad:


MikeZeDev:
Sorry but its a waste of time as Haruneko will success Hakuneko and its already done. 😔 

https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/MangaDistrict.ts

PR #8053, opened 2025-07-12
Updated AGS.mjs and changed name

Mark-Cal:
not sure if it works but this is all I can do

### DIFF
--- a/src/web/mjs/connectors/AGS.mjs
+++ b/src/web/mjs/connectors/AGS.mjs
@@ -5,10 +5,10 @@ export default class AGS extends WordPressMangastream {
     constructor() {
         super();
         super.id = 'ags';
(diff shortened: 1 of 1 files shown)


MikeZeDev:
> not sure if it works but this is all I can do

You are supposed to test if it works before submitting

Mark-Cal:
I don't know how to test it

MikeZeDev:
Locate the mjs file on your drive and modify it , restart hakuneko. It will stay like this until next haku update so you can test.

MikeZeDev:
Its another theme, it wont work at all https://github.com/manga-download/haruneko/blob/28506e43a495dddaaf70fe791867b779068a668a/web/src/engine/websites/AGS.ts#L4

PR #8096, opened 2025-08-05
Update AsuraScans.mjs

Champ3939:
Url Change


MikeZeDev:
1) Its a SETTING. The purpose of this is for user to change manually the URL.  Its written in the code. So this PR is useless.

image

image

2) Not only that but, it doesnt work, as website layout changed for a long time now . Never submit code without testing if it works, you'll waste your time and reviewer time. And they may be harsher than me.

PR #8124, opened 2025-08-17
Feat/automate downloads

rzanattaa-gif:


MikeZeDev:
First, hello and thanks for contributing. Second. Let me quote the repo page.

>The philosophy is ad-hoc consumption, get it when you going to read/watch it. It is not meant to be a mass downloader to stock up thousands of chapters that are just collected and will probably never be read.

We refused PR with command line support, and similar stuff that could lead to automation. We especially dont want any kind of automation (at least in Hakuneko). I am afraid we can accept this PR in this state. That is a straight no.

What is really a shame is that you improved download management. Under other circumstances we would gladly have reviewed it as is really a very useful feature. 

But the current state of the project is that we only accept simple PR like domain change or simple bugfixes. The reason is we are focusing on next gen Hakuneko, codename [HARUneko](https://github.com/manga-download/haruneko/).  While its far from ready for release, it already comes with a better download manager.

It breaks my heart to refuse contributions like that but we have no plans to add further features in Hakuneko. Haruneko is the future and it will finally merged to Hakuneko repo one day.. Sorry about that 😔 .

Sheepux:
There was however a debate about automation for bookmarks (this is what is done in this PR). But yep we will only review haku's PR.
Deleting a download task is already present in haru

PR #8133, opened 2025-08-23
Update MangaSusuku.mjs

amzz888:


src/web/mjs/connectors/MangaSusuku.mjs
```diff
@@ -7,7 +7,7 @@ export default class MangaSusuku extends WordPressMangastream {
         super.id = 'mangasusuku';
         super.label = 'MangaSusuku';
         this.tags = ['manga', 'indonesian'];
-        this.url = 'https://mangasusuku.xyz ';
+        this.url = 'https://mangasusuku.com ';
```
27136773, line 10: Spaces are not allowed in domain name

27136773 (changes requested):

PR #8134, opened 2025-08-23
Update MangaSusuku.mjs

amzz888:


27136773 (changes requested):
Spaces are not allowed in domain name

PR #8170, opened 2025-09-09
Update AsuraScans.mjs

Torablue:
Updating the website.


MikeZeDev (changes requested):
Same answer than here https://github.com/manga-download/hakuneko/pull/8096

* Website doesnt use Mangastream anymore

* The domain you changed is a SETTING that user can change itself in program settings. Its even written inside the code 🤦 . Its a useless change.

* This code will never work i can tell you without even testing. Which is something you should have done : testing.

Please never submit untested code. Its wasting your time and reviewers time.

Suggestion : recode website entirely, test it, or close this PR.

Issue #8369, opened 2026-01-16
[Site Request] mangabz


sinhaugoh:
### Is there an existing request for this website?

- [x] I have searched the existing issues

### Website Name

mangabz

### Site URL

https://mangabz.com/

### Series list

https://mangabz.com/manga-list/

### Manga example

https://mangabz.com/279bz/

### Manga chapter viewer

https://mangabz.com/m441367/

### Languages

Chinese

MikeZeDev:
Dev notes : Modified DM5

MikeZeDev:
✅ Is added in [HaruNeko](https://docs.hakuneko.download/user-manual/)
🚫 Will not be added to HakuNeko

PR #8374, opened 2026-01-18
Add mangabz connector

sinhaugoh:
Added connector for Mangabz

Fix issue: #8369


src/web/mjs/connectors/Mangabz.mjs
```diff
@@ -0,0 +1,110 @@
...
+        const data = await this.fetchDOM(request, "#chapterlistload a");
+
+        return data.map((element) => ({
+            id: element.getAttribute("href"),
+            title: element.innerText,
+        }));
+    }
+
+    async _getPages(chapter) {
+        const script = `
```
MikeZeDev, line 63: I dont wanna be rude  but considering the website only load one image at once, there is now way this code would work. Dont submit code that doesnt work / AI slop / without testing.

Image

src/web/mjs/connectors/Mangabz.mjs
```diff
@@ -0,0 +1,110 @@
...
+        super.id = "mangabz";
+        super.label = "MangaBZ";
+        this.tags = ["webtoon", "chinese", "anime"];
+        this.url = "https://www.mangabz.com";
+    }
+
+    async _getMangas() {
+        const BATCH_COUNT = 8;
+        let pageNumber = 1;
+        let hasNextBatch = true;
```
MikeZeDev, line 16: this is overcomplicated for nothing.

just loop, incrementing page until there is no manga on page

```js
    async _getMangas() {
        const mangaList = [];
        for(let page = 1, run = true; run; page++) {
            mangas.length ? mangaList.push(...mangas) : run = false;
        }
        return mangaList;
    }
        const request = new Request(new URL(this.path + '?page=' + page, this.url), this.requestOptions);
        const data = await this.fetchDOM(request, '');
        return data.map(element => {
            return {
                id: element.pathname,
                title: element.title.trim()
            };
        });
    }
```

src/web/mjs/connectors/Mangabz.mjs
```diff
@@ -0,0 +1,110 @@
...
+        const data = await this.fetchDOM(request, ".mh-list .mh-item-detali a");
+
+        return data.map((element) => ({
+            id: element.getAttribute("href"),
+            title: element.getAttribute("title"),
+        }));
+    }
+
+    async _getChapters(manga) {
+        const uri = new URL(manga.id, this.url);
```
MikeZeDev, line 51: you can inline stuff

```js
        const data = await this.fetchDOM(new Request(new URL(manga.id, this.url), this.requestOptions), "#chapterlistload a");
```

src/web/mjs/connectors/Mangabz.mjs
```diff
@@ -0,0 +1,110 @@
...
+                        } catch(error) {
+                            reject(error);
+                        }
+                    }, 500);
+                });
+            `;
+
+        const uri = new URL(chapter.id, this.url);
+        this.requestOptions.headers.set(
+            "x-user-agent",
```
MikeZeDev, line 79: Why Custom UA?

MikeZeDev (changes requested):
This PR is not able to fetch chapter images.
Please dont submit untested code.

src/web/mjs/connectors/Mangabz.mjs
MikeZeDev: This website  is based on DM5.  Check samples in our codebase to know how to handle it.

MikeZeDev:
Closing . No addition to haku anymore.

MikeZeDev:
✅ Is added in [HaruNeko](https://docs.hakuneko.download/user-manual/)
🚫 Will not be added to HakuNeko

sinhaugoh:
I have tested and I am able to fetch the images. I am setting the UA so that the website is rendered in mobile view which will show all images instead of just one (which you saw from the desktop view).

As for the batch fetching of mangas, I am doing it so to reduce the waiting time.

Also, FYI, DM5 connector doesn't work anymore.

Thanks for reviewing anyways.

MikeZeDev:
Perhaps. Anyway, we have more reliable way to fetch images on DM5 like website: https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/decorators/DM5.ts 

that async script call the website api. Ofc you have to tweak it a bit for this website, and since hakuneko electron engine is old i wouldnt be ok with using "eval" inside a  browser window.

Sorry that you lost you time on this. Haruneko is the future.

see https://github.com/manga-download/haruneko/commit/7197fbecbc9a5888bfdf9c62029a86bbe09cf014

sinhaugoh:
Understood, thanks for the clarification. I wasn’t aware about the new Haruneko project until today. Appreciate the review and thanks for creating MangaBz connector in the new project!

MikeZeDev:
All contributions are welcomed in Haruneko btw <3

PR #8459, opened 2026-03-13
Add Comic Y-Ours connector

viliml:


MikeZeDev (changes requested):
Add an icon. 64x64 px in png format, not extension and its done

MikeZeDev:
Fix selectors. Use partial classs matching because random garbage at the end is prone to changes.

PR #8460, opened 2026-03-15
Fix MangaKakalotFun: offset-based manga list pagination and pages format

xantasiadev:
Fixes two issues with the MangaKakalotFun connector.

The parent class uses `limit: 99999`, but the site's API was updated, deprecating it. it uses offset-based pagination (count: true, offset: N) with 30 items per page. only the first batch of results was being returned. connector now overrides _getMangas to paginate through all results using the count field from the response.

MangaHub._getPages only handled the `{p, i}` pages format. site's client also falls back to Object.values(data) when the i key is not there. connector now overrides inherited method to handle both formats

both fixes were tested, and linted, and derived from the website's javascript.

### DIFF
--- a/src/web/mjs/connectors/MangaKakalotFun.mjs
+++ b/src/web/mjs/connectors/MangaKakalotFun.mjs
@@ -10,40 +10,5 @@ export default class MangaKakalotFun extends MangaHub {
         this.url = 'https://mangakakalot.fun';
 
         this.path = 'mn01';
-        this.mangaPerPage = 30;
-    }
-
-    async _getMangas() {
-        const mangaList = [];
-        for (let offset = 0; ; offset += this.mangaPerPage) {
-            const gql = `{
-                search(x: ${this.path}, q: "", genre: "all", mod: ALPHABET, count: true, offset: ${offset}) {
-                    rows {
-                        id, slug, title
-                    }
-                    count
-                }
-            }`;
-            const data = await this.fetchGraphQL(this.apiURL, undefined, gql, undefined);
-            const rows = data.search.rows;
-            mangaList.push(...rows.map(manga => ({ id: manga.slug, title: manga.title })));
-            if (mangaList.length >= data.search.count || rows.length < this.mangaPerPage) {
-                break;
-            }
-        }
-        return mangaList;
-    }
-
-    async _getPages(chapter) {
-        const gql = `{
-            chapter(x: ${this.path}, slug: "${chapter.manga.id}", number: ${chapter.id}) {
-                pages
-            }
-        }`;
-        let data = await this.fetchGraphQL(this.apiURL, undefined, gql, undefined);
-        data = JSON.parse(data.chapter.pages);
-        const prefix = data.p || '';
-        const images = data.i || Object.values(data);
-        return images.map(page => this.createConnectorURI(new URL(prefix + page, this.cdnURL).href));
     }
 }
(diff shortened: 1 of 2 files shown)


src/web/mjs/connectors/MangaKakalotFun.mjs
```diff
@@ -1,19 +1,49 @@
...
+        for (let offset = 0; ; offset += this.mangaPerPage) {
+            const gql = `{
+                search(x: ${this.path}, q: "", genre: "all", mod: ALPHABET, count: true, offset: ${offset}) {
+                    rows {
+                        id, slug, title
+                    }
+                    count
+                }
+            }`;
+            const data = await this.fetchGraphQL(this.apiURL, undefined, gql, undefined);
```
MikeZeDev: destructure if you can  :D
            const { rows , search }= await this.fetchGraphQL(this.apiURL, undefined, gql, undefined);

src/web/mjs/connectors/MangaKakalotFun.mjs
```diff
@@ -1,19 +1,49 @@
...
+            const rows = data.search.rows;
+            mangaList.push(...rows.map(manga => ({ id: manga.slug, title: manga.title })));
+            if (mangaList.length >= data.search.count || rows.length < this.mangaPerPage) {
```
MikeZeDev: Better rewriting it according to https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/templates/MangaHubBase.ts#L70

src/web/mjs/connectors/MangaKakalotFun.mjs
```diff
@@ -1,19 +1,49 @@
...
+        return mangaList;
+    }
+
+    async _getPages(chapter) {
+        const gql = `{
+            chapter(x: ${this.path}, slug: "${chapter.manga.id}", number: ${chapter.id}) {
+                pages
+            }
+        }`;
+        let data = await this.fetchGraphQL(this.apiURL, undefined, gql, undefined);
```
MikeZeDev: Does it apply only on KalalotFun or other Mangahub websites? 

Perhaps it would be better to patch Mangahub?

src/web/mjs/connectors/MangaKakalotFun.mjs
```diff
@@ -1,19 +1,49 @@
...
         super.label = 'MangaKakalotFun';
+        this.tags = [ 'manga', 'webtoon', 'english' ];
         this.url = 'https://mangakakalot.fun';
 
         this.path = 'mn01';
+        this.mangaPerPage = 30;
+    }
+
+    async _getMangas() {
+        const mangaList = [];
```
MikeZeDev: Does it apply only on KalalotFun or other Mangahub websites?

Perhaps it would be better to patch Mangahub?

MikeZeDev (changes requested):
Make sure those changes arent needed for other Mangahub based website in that case put your changes in base class

Needs some minor code refactoring

MikeZeDev:
Can you please give an example of an affected chapter, if possible?

xantasiadev:
- noved both fixes to `MangaHub.mjs`, confirmed it will work for all subclasses extending it
- updated to match haruneko's MangaHubBase.ts approach
- destructured response and handles both the `{p, i}` format and the older `Object.values(data)` fallback

MikeZeDev (approved):
seems good

MikeZeDev:
Thanks !

PR #8478, opened 2026-03-28
Update MangaLivre.mjs

ho1903554-gif:


MikeZeDev:
Dont submit PR if you dont know what you are doing. Websites are totally different, changing an url wont make it magically work.

PR #8533, opened 2026-04-30
Yanmaga : Fix chapter list always returns empty

ghost:
The fetchUI script clicks the "more" button until it disappears, but the site's bound XHR handler does not fire under fetchUI's load conditions, so the button never disappears, the Promise never resolves, and the 10s timeout returns an empty array.

Replaces the click-loop with direct AJAX calls to the pagination endpoint.

### DIFF
--- a/src/web/mjs/connectors/Yanmaga.mjs
+++ b/src/web/mjs/connectors/Yanmaga.mjs
@@ -35,26 +35,39 @@ export default class Yanmaga extends SpeedBinb {
     async _getChapters(manga) {
         const baseUri = new URL(manga.id, this.url);
 
-        const initialLinks = await this.fetchDOM(new Request(baseUri, this.requestOptions), '.mod-episode > ul.mod-episode-list:first-of-type a.mod-episode-link');
-        const chapters = initialLinks.map(a => ({
-            id: new URL(a.getAttribute('href'), this.url).pathname,
-            title: a.querySelector('.mod-episode-title').textContent.trim(),
-            language: ''
-        }));
-
         const buttons = await this.fetchDOM(new Request(baseUri, this.requestOptions), 'button.mod-episode-more-button');
-        if (buttons.length === 0) {
-            return chapters;
+
+        if (buttons.length > 0) {
+            const btn = buttons[0];
+            const limit = parseInt(btn.getAttribute('data-limit')) || 100;
+            const sort = btn.getAttribute('data-sort') || 'older';
+            const path = btn.getAttribute('data-path');
+
+            const chapters = await this._fetchChaptersFromAjax(path, limit, sort);
+            if (chapters.length > 0) return chapters;
         }
 
-        const btn = buttons[0];
-        const limit = parseInt(btn.getAttribute('data-limit')) || 54;
-        const sort = btn.getAttribute('data-sort') || 'older';
-        const path = btn.getAttribute('data-path');
-        let offset = parseInt(btn.getAttribute('data-offset')) || chapters.length;
+        const links = await this.fetchDOM(new Request(baseUri, this.requestOptions), '.mod-episode ul.mod-episode-list a.mod-episode-link');
+        const seen = new Set();
+        const chapters = [];
+        for (const a of links) {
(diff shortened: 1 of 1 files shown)


ghost:
**Cause**
  The current `_getChapters` implementation injects a script via `Engine.Request.fetchUI` that polls for the "more" button (`.mod-episode-more-button`) every 1 second, clicks it until it disappears, then collects `a.mod-episode-link`:

```
async _getChapters(manga) {
    const chapterScript = `
        new Promise(resolve => {
            const interval = setInterval(() => {
                let morebtn = document.querySelector('.mod-episode-more-button');
                if (morebtn) morebtn.click()
                    else {
                        clearInterval(interval);
                        const chapters = [...document.querySelectorAll('a.mod-episode-link')];
                        resolve(chapters.map(...));
                    }
            }, 1000);
        });
    `;
    const uri = new URL(manga.id, this.url);
    return Engine.Request.fetchUI(request, chapterScript, 10000);
}
```
This breaks on the current Yanmaga site because:

  The DOM selectors (`.mod-episode-more-button`, `a.mod-episode-link`, `.mod-episode-title`) still exist and are correct. However, when the script clicks the "more" button inside the fetchUI BrowserWindow, the site's click handler issues an XHR to `/comics/<slug>/episodes?offset=N&limit=54&sort=older` that never completes successfully in this context (likely missing CSRF/cookie state or the site's bound handler does not fire under fetchUI's load conditions). Because the AJAX never returns, the more button is never removed from the DOM. The `setInterval` keeps clicking forever, the `else` branch never runs, the Promise never resolves, and the 10-second `fetchUI` timeout fires — returning an empty array.

**Proposed Fix**
  Replace the script-injection / button-click loop with a direct HTTP-driven enumeration. This has no timeout risk, and does not depend on the site's client-side click handler firing in a headless context.

```
async _getChapters(manga) {
    const baseUri = new URL(manga.id, this.url);

    // 1. Collect the initial server-rendered episodes from the manga page.
    const initialLinks = await this.fetchDOM(
        new Request(baseUri, this.requestOptions),
        '.mod-episode > ul.mod-episode-list:first-of-type a.mod-episode-link'
    );
    const chapters = initialLinks.map(a => ({
        id: new URL(a.getAttribute('href'), this.url).pathname,
        title: a.querySelector('.mod-episode-title').textContent.trim(),
        language: ''
    }));

    // 2. Locate the "more" button to obtain pagination parameters.
    const buttons = await this.fetchDOM(
        new Request(baseUri, this.requestOptions),
        'button.mod-episode-more-button'
    );
    if (buttons.length === 0) {
        return chapters;
    }

    const btn = buttons[0];
    const limit = parseInt(btn.getAttribute('data-limit')) || 54;
    const sort = btn.getAttribute('data-sort') || 'older';
    const path = btn.getAttribute('data-path');
    let offset = parseInt(btn.getAttribute('data-offset')) || chapters.length;

    // 3. Page through the AJAX endpoint directly.
    const itemRe = /data-episode-title=\\"([^"]+?)\\"[\s\S]*?data-original-url=\\"([^"]+?)\\"/g;
    const seen = new Set(chapters.map(c => c.id));

    for (let safety = 0; safety < 50; safety++) {
        const moreUri = new URL(path, this.url);
        moreUri.searchParams.set('offset', offset);
        moreUri.searchParams.set('limit', limit);
        moreUri.searchParams.set('sort', sort);

        const moreReq = new Request(moreUri, this.requestOptions);
        moreReq.headers.set('X-Requested-With', 'XMLHttpRequest');
        moreReq.headers.set('Accept', '*/*');

        let text;
        try {
            const response = await fetch(moreReq);
            text = await response.text();
        } catch (e) {
            console.error('[Yanmaga] AJAX fetch failed', e);
            break;
        }

        const found = [];
        itemRe.lastIndex = 0;
        let m;
        while ((m = itemRe.exec(text)) !== null) {
            const id = m[2];
            if (seen.has(id)) continue;
            seen.add(id);
            found.push({ id, title: m[1], language: '' });
        }

        if (found.length === 0) break;
        chapters.push(...found);
        offset += found.length;
        if (found.length < limit) break;  // last page
    }

    return chapters;
}
```

PR #8584, opened 2026-06-09
Update PixivComics buildId

skg321:
Update the hard-coded Next.js buildId used by the PixivComics connector.

The previous buildId caused _next/data/.../viewer/stories/{id}.json requests to return 404. The updated buildId restores page retrieval.
Tested locally with HakuNeko 6.1.7.


MikeZeDev:
The good way would be to adapt the script that is supposed to properly update the build id instead of making an update each time the build id change. Idk why it doesnt work , i guess its a timing problem.

Because same script works in [HaRUneko](https://github.com/manga-download/haruneko/blob/master/web/src/engine/websites/PixivComics.ts). 

Unfortunately, Hakuneko doesnt have a built in delay support for injected scripts, you would have to tweak the script a bit. 

I'll close this. Since its working without problems in Haruneko there is no reason to stick with Hakuneko.
Thanks for your contribution anyway.

skg321:
Thank you for the explanation. That makes sense.

PR #8637, opened 2026-07-13
Update URLs for Olympus Scanlation connector

W4ts0n:


MikeZeDev:
Hakuneko dev is stopped in favor of Haruneko

PR #8641, opened 2026-07-15
3asq (MangaLover): Change Domain

FiberAhmed:
3asq.org ->3asq.online
Idk if "BookmarkImporter.mjs" should be changed, but yeah whatever


MikeZeDev:
Use Haruneko

